PR #12, opened 2019-03-26
Release pipeline

lolski:

Release pipeline for `client-java`, which performs:
1. deployment of a Github draft release
2. Deployment of artifacts to Maven


.circleci/config.yml
```diff
@@ -83,11 +83,46 @@ jobs:
...
+  deploy-github:
+    machine: true
+    working_directory: ~/client-java
+    steps:
+      - install-bazel-linux-rbe
+      - checkout
+      - run: DEPLOY_GITHUB_TOKEN="$GRABL_CREDENTIAL" bazel run //:deploy-github
+
+  deploy-maven:
+    machine: true
```
haikalpribadi: where's the test equivalent of this job, @lolski ? I.e. `deploy-maven-test`? I see that it's still bundled up in `test-deployment-maven`. I think we should split `deploy-maven-test` and it should be as clean as this `deploy-maven`, except with argument `-- snapshot`

RELEASE_TEMPLATE.md
```diff
@@ -0,0 +1,3 @@
+# Client-Java
+
+ { Please add the release notes here }
```
haikalpribadi: Are you planning to add the installation instruction here too? I think we should show how to declare the dependency (`<groupId></>`, `<artifactId></>`, `<version></>`) in their pom.xml (like how we show in docs)

WORKSPACE
```diff
@@ -115,6 +115,30 @@ graknlabs_graql_maven_dependencies = "maven_dependencies")
...
+bazel_rules_docker_repositories = "repositories")
+bazel_rules_docker_repositories()
+
+load("@io_bazel_rules_docker//container:container.bzl", "container_pull")
+container_pull(
+  name = "openjdk_image",
+  registry = "index.docker.io",
+  repository = "library/openjdk",
+  tag = "8"
+)
```
haikalpribadi: same question as in https://github.com/graknlabs/graql/pull/28/files#r269271565

Why do we need these docker rules?

.circleci/config.yml
lolski: good point

WORKSPACE
lolski: fixed

RELEASE_TEMPLATE.md
lolski: fixed

.circleci/config.yml
lolski: fixed

PR #56, opened 2019-10-24
Integrate checking for unused Java dependencies

vmax:

Integrate checking for unused Java dependencies

- Load `@graknlabs_build_tools//unused_deps`'s dependencies
- Check for unused dependencies in `build` CI job
- Move libraries that are only required in runtime to `runtime_deps`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,6 +50,8 @@ jobs:
       - run: bazel run @graknlabs_build_tools//checkstyle:test-coverage
       - run-bazel-rbe:
           command: bazel build //...
+      - run: bazel run @graknlabs_build_tools//unused_deps -- list
+
 
   build-checkstyle:
     machine: true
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -61,6 +61,16 @@ jobs:
...
 
+  build-unused-deps:
+    machine: true
+    working_directory: ~/grakn
+    steps:
+      - install-bazel-linux-rbe
+      - checkout
+      - run:
+          command: bazel run @graknlabs_build_tools//unused_deps -- list
+          no_output_timeout: 30m
```
haikalpribadi: Can we make this run statement as part of the `build` job, @vmax ?

haikalpribadi: why do we need `no_output_timeout` ?

vmax: @haikalpribadi fixed in https://github.com/graknlabs/client-java/pull/56/commits/300d686ceade3d837b82bcab63e08a001c3a3e18

vmax: @haikalpribadi essentially, `@graknlabs_build_tools//unused_deps` invokes `bazel build //...` internally which might've taken more than 10 minutes (which is the default timeout)

haikalpribadi: why does it take more than 10 minutes, @vmax ? Is that expected or a problem?

lolski: it should be less than 1 hour now, given that we no longer need to unpack and re-pack the zip

vmax: @haikalpribadi i think that is expected for a cold (no cache/RBE involved) build

PR #68, opened 2020-02-21
Graql Behavior Steps and Define

flyingsilverfin:
Create the framework for BDD-based definitions of the graql behavior tests. This includes creating the step definitions for testing graql queries.

* Create the `test/behavior/graql/DefineTest.java` and corresponding `Steps`


test/behaviour/graql/BUILD
```diff
@@ -0,0 +1,82 @@
...
+    srcs = [
+        "TheoremSteps.java"
+    ],
+    deps = [
+        # Package dependencies
+        "//test/behaviour/connection:steps",
+        "//:client-java",
+
+        # Grakn Labs dependencies
+        "@graknlabs_common//:common",
```
haikalpribadi: I don't think this passes `@graknlabs_build_tools//unused_deps`

test/behaviour/graql/TheoremSteps.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Given("the data")
+    public void graql_insert_data(List<String> insertQueryStatements) {
+        GraqlQuery graqlQuery = Graql.parse(String.join("\n", insertQueryStatements));
+        tx().execute(graqlQuery);
+        tx().commit();
+    }
+
+    @When("executing")
+    public void graql_query(List<String> graqlQueryStatements) {
```
haikalpribadi: The best practice in defining Cucumber steps is to have the method name to be as identical as possible to the step syntax - i.e. by simply turning the step syntax into the snake_case version of them. I actually agree to this practice.
1) It makes it easier to debug when errors are thrown (when method and step names are the same)
2) Snake case is better than camelCase when you have very long names (in which tests do)

So let's stick to the above practice.

However, in addition to the above, notice how in the test name you came up with 3 consistent names that makes up the test scenarios? `graql_define`, `graql_insert` and `graql_query`. Let's also use this style for the step definitions in the feature files!

test/behaviour/graql/BUILD
```diff
@@ -0,0 +1,82 @@
...
+        "@graknlabs_grakn_core//:assemble-linux-targz", # Make sure to pass the path in args below
+    ],
+    args = [ # The order of the arguments matter
+        "grakn-core", # Keep at index 0, will be accessible at args[1]
+        "$(location @graknlabs_grakn_core//:assemble-linux-targz)", # Keep at index 1, will be accessible at args[2]
+    ],
+    classpath_resources = ["//test/setup:logback"],
+    size = "medium",
+    visibility = ["//visibility:public"]
+)
```
haikalpribadi: Let's move this target, along with the class `DefineTest`, to its own package `//define` ? It would make it easier to organise the target rule definitions when then get long, as we also want to add `test-kgms` for each `test-core`.

haikalpribadi: Also, let's follow the `//behaviour` package structure as outlined in the `@graknlabs_verification` repo?  The `//behaviour/graql` package should have 3 sub-packages:
- `//language` (where `//define` in this PR should go)
- `//reasoner` (where generalisation test should go)
 - `//analytics` (once we have it)

test/behaviour/graql/TheoremSteps.java
```diff
@@ -0,0 +1,138 @@
...
+import io.cucumber.java.en.Then;
+import io.cucumber.java.en.When;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+public class TheoremSteps {
```
haikalpribadi: Is `TheoremSteps` the best name for these set of steps? I think `GraqlSteps` would be better.

rpc/Transceiver.java
```diff
@@ -88,6 +88,7 @@ public Response receive() throws InterruptedException {
     public void close() {
         try {
             requestSender.onCompleted();
+            responseListener.onCompleted();
```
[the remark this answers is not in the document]
flyingsilverfin: Will look into this, this was meant to fix a bug that when transations are committed, they aren't also closed

PR #70, opened 2020-02-21
Refactor GraknClient public methods and add grabl tracing

adammitchelldev:

To add grabl tracing to GraknClient.

- Refactored the public methods on `GraknClient.Transaction` so that they don't call into each other in a way that makes tracing too difficult to achieve (some still call back in but in a way that only uses overloading to provide default parameters).
- Added grabl tracing using `GrablTracingThreadStatic`.


GraknClient.java
```diff
@@ -361,16 +381,19 @@ public Keyspace keyspace() {
...
         }
 
         public Stream<AnswerGroup<ConceptMap>> stream(GraqlGet.Group query) {
             return stream(query, true);
         }
 
         public Stream<AnswerGroup<ConceptMap>> stream(GraqlGet.Group query, boolean infer) {
-            Iterable<AnswerGroup<ConceptMap>> iterable = () -> rpcIterator(query, infer);
-            return StreamSupport.stream(iterable.spliterator(), false);
+            try (ThreadTrace trace = traceOnThread("tx.stream.get_group")) {
```
haikalpribadi: Can we not use underscores or hyphens, and just use dots (`.`) to scope trace names? In this case, `tx.stream.get.group`, but this should be applied to all other trace names.

PR #71, opened 2020-02-24
Close transaction on fully on commit

flyingsilverfin:
A transaction can currently be committed but still return `true` for `isOpen()`. We should have the following behavior:

1. comitting the tx should close the tx; re-committing the tx should throw an exception
2. closing a committed/closed tx should be idempotent and do nothing.

* Fully close the receivers and senders


haikalpribadi:
What was breaking before you found this bug, @flyingsilverfin ?

flyingsilverfin:
More of a behavioral change; need to figure out how to fit this into Transaction BDD steps though

PR #79, opened 2020-03-09
Add tracing metadata to requests

adammitchelldev:

To add grabl tracing information to the metadata in the requests so that we can continue the tracing on the Grakn server side.

- Added static tracing info to the metadata of the GRPC requests.
- Drive-by: removed the data logging as it may be contributing to poor tracing performance.

### DIFF
--- a/rpc/RequestBuilder.java
+++ b/rpc/RequestBuilder.java
@@ -34,7 +34,10 @@ import graql.lang.query.GraqlQuery;
 
 import java.time.LocalDateTime;
 import java.time.ZoneId;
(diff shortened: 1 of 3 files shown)


rpc/RequestBuilder.java
```diff
@@ -34,8 +34,11 @@
 
 import java.time.LocalDateTime;
 import java.time.ZoneId;
-import java.util.Collection;
+import java.util.*;
```
haikalpribadi: Are we sure we need everything from `java.util`?

PR #81, opened 2020-03-16
Read rules from the explanation response proto

jmsfltchr:

The rule that an explanation is explaining is now returned inside the explanation object. We receive this message via the gRPC and build a `Rule` instance for the user to access from `getRule()` on an explanation.

This requires [https://github.com/graknlabs/protocol/pull/24](https://github.com/graknlabs/protocol/pull/24)
and is required by [https://github.com/graknlabs/grakn/pull/5657](https://github.com/graknlabs/grakn/pull/5657)

- Updates dependencies
- Maps gRPC response to `Rule` object


dependencies/graknlabs/dependencies.bzl
```diff
@@ -44,26 +44,26 @@ def graknlabs_grakn_core():
...
-        tag = "1.0.4", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
+        remote = "https://github.com/jmsfltchr/protocol",
+        commit = "87c5b205ab7b8747e05e9310cc8e64bdc34c41bc", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
     )
 
 def graknlabs_verification():
     git_repository(
         name = "graknlabs_verification",
-        remote = "https://github.com/graknlabs/verification",
+        remote = "git@github.com:graknlabs/verification.git",
```
flyingsilverfin: Are we keeping this using ssh just for private repos now?

answer/Explanation.java
```diff
@@ -32,14 +34,21 @@
 public class Explanation {
 
     private final List<ConceptMap> answers;
+    private final Rule rule;
 
-    public Explanation(List<ConceptMap> ans) {
+    public Explanation(List<ConceptMap> ans, Rule rule) {
```
flyingsilverfin: can put `@nullable` infront of `Rule rule` here, just to indicate the intent is that it _may_ be null

dependencies/graknlabs/dependencies.bzl
jmsfltchr: Good question, a question for the team

haikalpribadi: I didn't know you can provide the SSH address here in `dependencies.bzl`. @vmax is this right? it seems to work, but is this an expected pattern in Bazel?

vmax: @haikalpribadi you can, and you actually should — otherwise cloning fails with a cryptic error

haikalpribadi: you mean that we should always use `git@` for all `git_repository()` ? We should make this sweeping change in all repositories then! I didn't know. @vmax

haikalpribadi: by "always", I mean for both `public` and `private` repos? or is it just meant for `private` repos?

vmax: @haikalpribadi nope, I only meant those that are not public, such as `@graknlabs_verification`

haikalpribadi: what happens if you use `git@` for all `git_repository()` @vmax ?

answer/Explanation.java
```diff
@@ -32,14 +35,21 @@
 public class Explanation {
 
     private final List<ConceptMap> answers;
+    private final Rule rule;
 
-    public Explanation(List<ConceptMap> ans) {
+    public Explanation(List<ConceptMap> ans, @Nullable Rule rule) {
```
haikalpribadi, line 40: Why can the `rule` be `@Nullable`?

rpc/ResponseReader.java
```diff
@@ -79,7 +80,9 @@ public static Answer answer(AnswerProto.Answer res, GraknClient.Transaction tx)
     public static Explanation explanation(AnswerProto.Explanation.Res res, GraknClient.Transaction tx) {
         List<ConceptMap> answers = new ArrayList<>();
         res.getExplanationList().forEach(explanationMap -> answers.add(conceptMap(explanationMap, tx)));
-        return new Explanation(answers);
+        ConceptProto.Concept ruleProto = res.getRule();
+        Rule rule = res.hasRule() ? ConceptImpl.of(ruleProto, tx).asRule() : null;
```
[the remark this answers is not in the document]
flyingsilverfin: shouldn't be

answer/Explanation.java
jmsfltchr: Because we have (on the server side) 3 kinds of explanation - JoinExplanation, LookupExplanation and RuleExplanation. Only the latter actually shows the derivation of facts from a rule. 
I'm inclined to think that these could be named `Resolution`s rather than `Explanation`s.

rpc/ResponseReader.java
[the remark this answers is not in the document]
jmsfltchr: Yes, in the case of a JoinExplanation or LookupExplanation (see above)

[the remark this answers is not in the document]
haikalpribadi: 👍

answer/Explanation.java
haikalpribadi: 👍

dependencies/graknlabs/dependencies.bzl
vmax: @haikalpribadi then we probably won't be able to clone them in CI unless we specifically add the SSH key in CircleCI settings

Issue #84, opened 2020-03-23
Performance refactor of Answer and Concept API


adammitchelldev:
## Problem to Solve

Currently, the Concept API is the main point of data transfer to the Grakn user. A Graql query is executed (or streamed), which results in a number of Concepts being returned (usually a list of ConceptMaps). These concepts are very simply Grakn IDs which are then used by synchronous Concept API RPCs to retrieve data and other associated concepts.

This means that all data returned from a query is actually fetched as separate concept API RPC calls, which results in a large amount of bidirectional communication and waiting. Any amount of latency introduced by the network is multiplied by the number of these RPCs, which can often be a multiple of the number of concepts retrieved by a single query.

As an example, if network latency is 10ms and 100 concepts are fetched, with 4 RPC calls per concept, 4 seconds of total latency is introduced.

It is clear that this can pose a performance issue to users who intend to build low-latency applications on Grakn.

## Current Workaround

There is currently no good workaround, other than to minimise the number of Concept API calls made (it is impossible to do asynchronous calls on a transaction currently).

A short-term fix was proposed which would simply pre-fetch common Concept API answers, but then the boundaries between RPC-driven communication and results parsing would be obfuscated. This solution would alleviate the issues but we've decided on a wider-spread refactor to deal with the issue.

## Proposed Solution

Instead of the full `Concept` API being the single point of contact for `Answer`s, the `Answer` become "Value-only" `Concepts`, as in they only expose the methods to retrieve the pre-fetched results and don't expose anything further. This makes the separation between what exists in results (and therefore doesn't require any network overhead to use) and what must be fetched or performed via RPC explicit to the user.

The way I am planning to achieve this is to simplify the current Concept interfaces down to a simple set of operations that should superset the 99% use case for queries. This will include attribute values and schema type information that will also be embedded in the answers protocol. When a user wants to use the full Concept API instead, it must construct the "live" concept from the retrieved concepts using a method and passing a transaction. For example:

```
concept.withTransaction(tx).sups()
```

This requires an API change that will break existing users of Grakn client, so we will also take the opportunity to rework the client API into an interface with a static factory method.

## Summary of required work

- [ ] Update protocol [x], server [ ] and client [ ] to allow retrieval of multiple concepts from a single RPC.
- [ ] Update protocol to include basic concepts to be used in "Answer" API.
- [ ] Update server to transmit all basic concepts for the "Answer" API.
[...]

PR #85, opened 2020-03-23
Client Java: Ferrari Edition

adammitchelldev:

Linked to #84

* To allow the client to use the "all" results feature of the protocol in order to reduce overall number of roundtrips in communication to 1 per execute stream iteration.
* To pre-fill results in answers in order to avoid common fetch operations, particularly in console.
* To refactor the Concept API to distinguish between remote and local concepts. This is important because remote concepts are transaction locked.

- Added separate `execute` logic that doesn't iterate the results individually and instead requests all of them.
- Using the new protocol for pre-filled results.
- Refactor into remote/local concept API.
- Returning the pre-filled local concepts instead of remote ones.


adammitchelldev:
This won't pass tests until the protocol changes are merged and dependency updated here.

adammitchelldev:
Cool, those package changes look better I agree.

adammitchelldev:
> Based on the call we had where you walked us through the changes, they're all clear and very good! I'd just like us to organise the new Concept classes and interfaces in a more structured way:
> 
> Split the concept packages the following package:
> 
> ```
> concept/
> ├── thing/
> │   ├── Attribute.java
> │   ├── Entity.java
> │   ├── Relation.java
> │   ├── Role.java
> │   └── impl/
> │       ├── AttributeImpl.java
> │       ├── EntityImpl.java
> │       ├── RelationImpl.java
> │       └── RoleImpl.java
> └── type/
>     ├── AttributeType.java
>     ├── EntityType.java
>     ├── RelationType.java
>     ├── RoleType.java
>     └── impl/
>         ├── AttributeTypeImpl.java
>         ├── EntityTypeImpl.java
>         ├── RelationTypeImpl.java
>         └── RoleTypeImpl.java
> ```
> 
> For each interface/class, nest the 2 classes `Local` and `Remote`. So you'll have something like `AttributeType.Local` and `AttributType.Remote` for the interfaces, and `AttributeTypeImpl.Local` and `AttributeType.Remote` for the implementations.
> 
> You may need to fit `SchemaConcept` and `Rule` in `type/` for now.

Have implemented this

PR #86, opened 2020-04-01
Allow duplicate role players

flyingsilverfin:
Allowing duplicate roles to be played in a relation as of https://github.com/graknlabs/grakn/pull/5667.

This means now that `relation.getRolePlayerMap()` should return a map from `Role` to `List<Thing>` rather than `Set<Thing>`

* Write new test for retrieving duplicate role players
* Update types for `rolePlayerMap` in the concept API

### DIFF
--- a/concept/Relation.java
+++ b/concept/Relation.java
@@ -20,6 +20,7 @@
 package grakn.client.concept;
 
 import javax.annotation.CheckReturnValue;
+import java.util.List;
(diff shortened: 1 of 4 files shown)


flyingsilverfin:
TODO revert dependency to graknlabs dependencies because of dependency cycle right now: grakn core -- console -- client-java -- grakn core

PR #92, opened 2020-04-20
Async iteration

adammitchelldev:

To allow the client to begin iteration on their local stream of results before all of the results for a batch have completed being fetched.

Decoupled the results fetching to a separate thread.

### DIFF
--- a/GraknClient.java
+++ b/GraknClient.java
@@ -33,6 +33,7 @@ import grakn.client.answer.Numeric;
 import grakn.client.answer.Void;
 import grakn.client.concept.Concept;
 import grakn.client.concept.DataType;
+import grakn.client.concept.GraknConceptException;
 import grakn.client.concept.type.Role;
 import grakn.client.concept.Rule;
 import grakn.client.concept.SchemaConcept;
@@ -72,8 +73,12 @@ import javax.annotation.Nullable;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
@@ -82,6 +87,7 @@ import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import static grabl.tracing.client.GrablTracingThreadStatic.traceOnThread;
+import static grakn.client.rpc.Transceiver.Response.Type.OK;
 
 /**
  * Entry-point which communicates with a running Grakn server using gRPC.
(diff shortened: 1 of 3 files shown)


exception/GraknClientException.java
```diff
@@ -45,7 +45,10 @@ public static GraknClientException create(String error) {
         return new GraknClientException(error);
     }
 
-    public static GraknClientException create(String error, StatusRuntimeException e) {
+    public static GraknClientException create(String error, Exception e) {
+        if (e instanceof StatusRuntimeException) {
+            return new GraknClientException(error, (StatusRuntimeException) e);
+        }
```
haikalpribadi, line 51: What if we just overload the method by like the following, rather than manually downcasting?
```
public static GraknClientException create(String error, StatusRuntimeException e) {
        return new GraknClientException(error, e);
    }
```

GraknClient.java
```diff
@@ -617,9 +625,55 @@ public boolean isClosed() {
...
+                    }
+                }
+                return false;
+            }
+
+            @Override
+            protected SessionProto.Transaction.Iter.Res computeNext() {
+                Transceiver.Response response;
+                try {
+                    response = queue.take();
```
haikalpribadi, line 651: What happens when the queue is empty (but more responses is yet to come) and `queue.take()` is called?

GraknClient.java
```diff
@@ -617,9 +625,55 @@ public boolean isClosed() {
...
+                    throw GraknConceptException.create("Iteration Interrupted");
+                }
+
+                switch (response.type()) {
+                    case OK:
+                        return response.ok().getIterRes();
+                    case ERROR:
+                        throw GraknClientException.create(response.error().getMessage(), response.error());
+                    case COMPLETED:
+                        throw GraknClientException.create("Transaction interrupted, all running queries have been stopped.");
```
haikalpribadi, line 662: Are we sure we're supposed to throw an exception here? Does `COMPLETED` not mean that the stream of responses has "correctly completed"?

exception/GraknClientException.java
adammitchelldev: Can do

GraknClient.java
adammitchelldev: No, this is when the GRPC stream is completed, which is not expected (should happen after the transaction is closed).

GraknClient.java
adammitchelldev: `queue.take()` is the blocking form (as opposed to `.poll()` which does not block), so this will wait for the next result.

exception/GraknClientException.java
adammitchelldev: Actually, no, I remember why I didn't want to do that, it's because the `Exception` might be a `StatusRuntimeException` but the compile time type is an `Exception` (caught as `Exception` for a more general case).

adammitchelldev: Since the logic to extract error code is different, it's important it gets treated as a special case even if it was upcast to an `Exception` before the method call.

haikalpribadi: Right. I suspected that. And I wasn't sure which method would be called by java. 🤔

In any case, your way is safer.

GraknClient.java
haikalpribadi: 👍

GraknClient.java
haikalpribadi: right. 👍

PR #100, opened 2020-05-15
Use release version of tracing

adammitchelldev:

To fix a dependency issue where a snapshot version of grabl tracing was being used rather than a release version, requiring the user to add our snapshot repo. This release version was not available at the time but the snapshot repo was already in the maven test pom, which meant the issue was not noticed when the snapshot version was used.

- Removed snapshot repo from maven test application
- Changed version of grabl tracing to 0.1.0


haikalpribadi:
This PR has an empty `diff`, @adammitchelldev @lolski. Is this expected? Should we close it?

PR #102, opened 2020-05-19
Use release version of deps

adammitchelldev:

Update to the latest release of all graknlabs dependencies.

- Updated all graknlabs dependencies to their latest release.


haikalpribadi:
What's the status of this? Does it need any help?

PR #103, opened 2020-05-19
Revert to 1.7.x compatible deps and syntax for release

flyingsilverfin:
In preparation of release: client-java was ahead of grakn-core and other clients in that it implemented Delete functionality for 1.8 and used `ValueType` instead of `DataType` (also 1.8.x functionality). This also required it used the next version of the Graql and protocol (though it is possible to use it as gRPC affords flexibility here, we don't want guesswork of which dependency is "ahead" and which is "in sync" in general). We aim to keep 1.N in line between Clients and Core, using the in-sync and compatible Graql and Protocol dependencies.

* Revert dependencies back to 1.0.6 and 1.0.5 for Graql and Protocol respectively
* Relatedly, undo `ValueType` and use `DataType`, and remove one instance of new `delete` syntax

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,8 +98,8 @@ jobs:
     steps:
       - install-bazel-linux-rbe
       - checkout
-      - run-bazel-rbe:
-          command: bazel test //test/behaviour/graql/language/define:test-core --test_output=errors
+#      - run-bazel-rbe:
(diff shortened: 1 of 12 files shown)


lolski:
Approved.

It looks like the release requires a lot of effort, caused by the fact that we had to temporarily revert features meant for 1.8.

In general this is the limitation we've got to live with, given that we work with one mainline branch.

PR #104, opened 2020-05-20
Fix static tracing not working for async requests

adammitchelldev:

Grabl tracing requires us to propagate the trace across threads ourselves, so the introduction of an async decoupling executor in our RPC transceiver (for the results streaming) breaks users of Grabl tracing. We have fixed this.

Propagate the grabl tracing trace to the transceiver executor task.

### DIFF
--- a/rpc/Transceiver.java
+++ b/rpc/Transceiver.java
@@ -19,6 +19,7 @@
 
 package grakn.client.rpc;
 
+import grabl.tracing.client.GrablTracingThreadStatic;
 import grabl.tracing.client.GrablTracingThreadStatic.ThreadTrace;
 import grakn.client.exception.GraknClientException;
 import grakn.protocol.session.SessionProto;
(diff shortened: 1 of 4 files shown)


haikalpribadi:
Can we add a test for this, @adammitchelldev ?

rpc/Transceiver.java
```diff
@@ -116,22 +117,24 @@ public Response sendAndReceive(Transaction.Req request) throws InterruptedExcept
...
-            try {
-                send(request);
-                while (true) {
-                    Response res = receive();
-                    if (!receiver.onResponse(res)) {
-                        break;
+        try (ThreadTrace parentTrace = GrablTracingThreadStatic.traceOnThread("asyncQueue")) {
+            executorService.execute(() -> {
+                lock.lock();
+                try (ThreadTrace ignored = parentTrace.traceOnThread("asyncExecute")) {
```
flyingsilverfin, line 123: why is this called ignored?

flyingsilverfin: but unintuitive

adammitchelldev: `ignored` is a "reserved" variable name that disables warnings automatically. Anonymous resources are not available until a later version of Java.

flyingsilverfin: lol never knew this, okay

PR #106, opened 2020-05-22
Add 'grakn define throws' to BDD framework, update verification

alexjpwalker:

To update verification and support the new version.

I implemented grakn define throws in the BDD framework and updated the version of verification to a local fork. It will later be updated to a new version on the main branch of verification.


.circleci/config.yml
```diff
@@ -261,10 +261,10 @@ workflows:
...
             branches:
               ignore: client-java-release-branch
-      - test-behaviour-graql-reasoner:
-          filters:
-            branches:
-              ignore: client-java-release-branch
+#      - test-behaviour-graql-reasoner:
+#          filters:
+#            branches:
+#              ignore: client-java-release-branch
```
haikalpribadi, line 267: We don't need to comment these lines out if we're already commenting them out of the workflow definition.

PR #108, opened 2020-05-27
Refactor client to sit behind interfaces

adammitchelldev:

GraknClient has been implemented as plain classes, which makes it difficult to compatibility with future versions. This has not been important for now but will become a lot more important when releasese become stable, so the sooner we can make this change, the better.

This will cause all users of GraknClient to break, as they will need to use the new factory method syntax:

```
GraknClient.open("hostname:port");
```

This is also more in-line with typical Java clients and APIs, since they have followed this reasoning already.

- Refactor out all of the main classes from GraknClient, leaving them as implementation classes and new Interfaces.


adammitchelldev:
I would like some comments on how we want to structure the client in code and also to the users. We could leave them as separate interface classes to avoid requiring nested naming for the users, which should make them slightly less terse and better organised for javadocs (the implementation classes are still nested).

I think we should also move all of the interfaces and the implementation to a `client` folder, but leave them in the base package.

haikalpribadi:
This has been done in https://github.com/graknlabs/client-java/pull/157

Issue #110, opened 2020-05-28
Meta Types break


flyingsilverfin:
## Description
Asking `tx.getMetaEntityType()` breaks

## Environment
Grakn Core 1.8 (master) and client-java 1.8 (master, 6074059cbaf6eb5cb4e564ef950be68a1653b9c2).

## Reproducible Steps
```

    @Test
    public void testMeta() {
        GraknClient.Session testing = client.session("testing");
        GraknClient.Transaction tx = testing.transaction().write();
        tx.execute(Graql.parse("define person sub entity;").asDefine());

        MetaType.Remote<?, ?> metaEntityType = tx.getMetaEntityType();
    }
```

## Expected Output
Get the meta "entity" type.

## Actual Output
Exception: 
```
The concept [grakn.client.concept.type.impl.EntityTypeImpl.Remote{tx=grakn.client.GraknClient$Transaction@5ce8d869, id=V4152}] is not of type [interface grakn.client.concept.type.MetaType$Remote]
grakn.client.concept.GraknConceptException: 
	at grakn.client.concept.GraknConceptException.create(GraknConceptException.java:35)
	at grakn.client.concept.GraknConceptException.invalidCasting(GraknConceptException.java:42)
	at grakn.client.concept.Concept$Remote.asMetaType(Concept.java:559)
	at grakn.client.GraknClient$Transaction.getMetaEntityType(GraknClient.java:770)
	at grakn.client.test.integration.answer.AnswerIT.testMeta(AnswerIT.java:74)
```
 
## Additional information
I thought we had tests that should hit this case?

adammitchelldev:
I was sure we did

adammitchelldev:
Fixed by #111

PR #111, opened 2020-05-28
Fix metatypes

adammitchelldev:

To fix the basic metatype getters on transaction and add tests to ensure they don't break again in future. This was broken because the API was confused between using metatypes and using the actual types. Here, we can switch properly to using the actual types, as they contain all the methods and are type-compatible with `subs()`.

- Metatype getters now get the actual type class, rather than the special metatype class. This is how the implementation worked and the fact that they didn't would just cause a casting exception, so this is a bug fix.

### DIFF
--- a/GraknClient.java
+++ b/GraknClient.java
@@ -750,28 +750,28 @@ public class GraknClient implements AutoCloseable {
             }
         }
 
-        public SchemaConcept<?> getMetaConcept() {
+        public SchemaConcept.Remote<?> getMetaConcept() {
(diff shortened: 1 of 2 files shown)


flyingsilverfin:
sooo can we delete the `Meta` type classes?

adammitchelldev:
You would hope so but the protocol can still give us `META_TYPE` or whatever, so we need something to represent that until we can remove it altogether.

flyingsilverfin:
hmm okay

adammitchelldev:
fix #110

PR #112, opened 2020-05-28
Add full async for queries

adammitchelldev:

Added fully async queries, which allows clients to queue up multiple queries that will run in sequence without requiring additional round-trips to receive results. This is especially useful for streaming large amounts of writes.

- Refactored the Transceiver to use the same pattern as other clients, but with more complicated threading.
- [x] Refactor the query iterator to use the new pattern.
- [x] Change the API to introduce Future for queries.


rpc/Transceiver.java
```diff
@@ -160,49 +122,137 @@ public boolean isOpen() {
...
+         */
+        boolean onResponse(Response response);
+    }
+
+    /**
+     * Simple response collector for when a single result is expected.
+     */
+    private static class SingleResponseCollector implements ResponseCollector {
+        private volatile Response response;
+        private final CountDownLatch latch = new CountDownLatch(1);
```
flyingsilverfin, line 142: why do we need this? Is it for corectness?

adammitchelldev: It's because the collector receives messages from a pushing thread and returns them to pulling thread, `CountDownLatch` is used to block the pulling thread until the pushing thread has delivered the value.

adammitchelldev:
> looks good! So query -> sends initial request; `.get()` blocks until it has a first response, right?

Yes, that's correct.

adammitchelldev:
To help with reviewing, here's an overview of what's going on with the threading:

There are (or should only be) two threads, User and GRPC.

The User thread is any thread that the client makes API calls with. Technically, it could be called concurrently, which we actually support. A key thing to bear in mind is that API only ever returns sequential streams, so iteration methods like `computeNext()` are never called in parallel, even if they are being called by a User thread (we can safely assume it's always the same User thread).

The GRPC thread delivers our responses from the server. There is only ever one GRPC thread.

The User thread needs to fetch responses corresponding to requests that it sent, so there needs to be a way to hand off the correct incoming responses to the correct User thread that is waiting for it.

We achieve this by using "collectors", which provide one method to accept and buffer responses from the GRPC thread, and another method to fetch the corresponding buffered response in a blocking manner from the User thread. The result is that the User thread can make a call to fetch a result and it will be blocked until the GRPC thread has received that result.

The general flow for a unary call is:

1. User initiates a query
2. User thread simultaneously sends a request to the server and queues a Collector.
3. The User thread blocks on the `get()` method of the Collector.
4. The GRPC thread receives a response.
5. The GRPC thread dequeues the Collector and delivers the response.
6. The Collector saves the response from the GRPC thread call and then unblocks the User thread call, returning the saved response.

In the streaming flow, the responses from the GRPC thread are placed into a blocking queue, which achieves this effect.

flyingsilverfin:
Ok - you want to put a shortened version of that in the code itself or not?

adammitchelldev:
> Ok - you want to put a shortened version of that in the code itself or not?

What I actually want to do is refactor it in an upcoming PR so that the threading split is clear by class and interface. I also would like to introduce the idea of building in an "event-loop", which will require that refactor.

test/integration/answer/AnswerIT.java
```diff
@@ -118,4 +125,34 @@ public void testExplanation() {
...
+            try (GraknClient.Transaction tx = session.transaction().write()) {
+                List<ConceptMap> answers = tx.execute(match(
+                        var("x").isa("person").has("name", var("y"))
+                ).get()).get();
+
+                assertEquals(answers.size(), 2);
+            }
+        }
+    }
 }
```
haikalpribadi, line 158: Missing new line ;)

rpc/Transceiver.java
```diff
@@ -82,69 +78,36 @@ public static Transceiver create(SessionServiceGrpc.SessionServiceStub stub) {
...
-            Response response = responseListener.poll();
-            LOG.trace("receive:{}", response);
-            if (response.type() != Response.Type.OK) {
-                close();
-            }
-            return response;
+        // We must add the response collectors in exact the same order we send the requests
+        synchronized (this) {
+            responseListener.addCollector(collector); // Must add collector first to be watertight
+            requestSender.onNext(request);
```
haikalpribadi, line 90: how does synchronising these two lines guarantee that the order, @adammitchelldev ? Don't you need to make the method to be `synchronised` in order to make the order guaranteed?

haikalpribadi: and do we need to add a test for this?

adammitchelldev: The `synchronized (this)` acquires an implicit monitor lock for this block, so only one thread can ever call this block of code on this `Transceiver` at a time. This is the point at which we send the actual request, so by synchronizing this with recording the collector (corresponding to this request), we guarantee that we'll be processing the results in the same order we send the requests.

Nothing else is conflicting, so nothing else needs to be synchronized.

adammitchelldev: Worth bearing in mind that
```
public synchronized void doSync() {
    // code
}
```
Is the same as:
```
public void doSync() {
    synchronized (this) {
        // code
    }
}
```
`synchronized (this)` obtains an object-wide lock.

haikalpribadi: Yes I know they're the same, but judging from your comment, you said that you wanted to maintain order or based on who called `send()` first? If so, then adding the `synchronized` inside the function won't guarantee it being executed in the order of who called the `send()` method, right?

adammitchelldev: Ah, the order of who calls the outer send method isn't important, only that whichever one gets given to GRPC first also has its collector added first. If two different threads were to make requests at the same time, there would be no guarantee which order they'd actually get sent in either way.

haikalpribadi: Ah, right. Okay merging this.

adammitchelldev: By send I mean send the GRPC request (`requestSender.onNext(request);`)

PR #116, opened 2020-06-09
Fix error message missing from async queries on commit

adammitchelldev:
Fix an issue where error messages were unclear when committing a transaction with async queries that fail.

- Forward the actual async errors rather than an unexpected completion response.
- add a test to confirm this happens on commit

### DIFF
--- a/rpc/Transceiver.java
+++ b/rpc/Transceiver.java
@@ -241,7 +241,7 @@ public class Transceiver implements AutoCloseable {
 
             // Exhaust the queue
             while (currentCollector != null || collectorQueue.peek() != null) {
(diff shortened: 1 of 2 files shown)


flyingsilverfin:
can we add tests for this now?

flyingsilverfin:
What's the behaviour when there are a bunch of async queries that all have errors - do the errors all get stacked up?

adammitchelldev:
Only the first error is reported, since beyond the first error the Grakn
server behaviour is not well-defined.

flyingsilverfin:
ahh right okay!

PR #121, opened 2020-06-15
Explainable query flag

adammitchelldev:

We've introduced an explanation cache that is used on the server side to cache and return explanations, but since we didn't previously know if an explanation is required ahead of time, we now give clients the option to specify whether or not the explanation cache is enabled for a given query.

Introducing this new option highlights an issue with client-java gaining too many optional parameters and we expect these parameters to change over time, so we're taking the step to use an Options object instead, which will allow clients to use a simple builder syntax to set various options.

- Added explanation flag option
- Updated to new protocol including explanation option
- Updated other dependencies to fix various issues
- Added an options object to replace the `infer` boolean that now contains multiple options.


GraknClient.java
```diff
@@ -1073,4 +1081,72 @@ public int hashCode() {
...
+
+        @FunctionalInterface
+        interface FlagConsumer {
+            void accept(boolean flag);
+        }
+    }
+
+    public interface Options {
+        OptionsBuilder DEFAULT = new OptionsImpl();
+        OptionsBuilder NO_INFER = DEFAULT.infer(false);
```
flyingsilverfin, line 1141: easier to read without these two extra enums - just use `DEFAULT.explain(...)` and its inverse

test/integration/answer/AnswerIT.java
```diff
@@ -169,4 +167,36 @@ public void writingInAReadTransactionThrows() {
             }
         }
     }
+
+    @Test
```
flyingsilverfin, line 171: need tests for the other option too - `explain`

adammitchelldev: Have added

GraknClient.java
adammitchelldev: Removed

PR #131, opened 2020-06-26
Update to new structure for explanation tests

jmsfltchr:

We bring the structure of tests for explanations in line with that of `graknlabs/grakn` and `graknlabs/verification`.

- Moves explanation tests to their own module
- Within the module, we have one Feature with reasoning and one without
- Updates CI to run these two test Features


haikalpribadi:
@jmsfltchr is this PR still valid?

jmsfltchr:
No we don't need this. We do however need to reintroduce these explanation tests into client-java once we have an implementation for them.

PR #134, opened 2020-07-08
Depend on grakn artifact

adammitchelldev:

To depend on the artifact of grakn for tests instead of depending on it through source. This should cut down issues where the version of grakn core being tested against is incorrect due to dependency leakage from this repository.

- Removed dependency on grakn core code
- Added dependency on grakn core artifact

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -175,17 +175,10 @@ graknlabs_grabl_tracing()
 load("@graknlabs_grabl_tracing//dependencies/maven:artifacts.bzl", graknlabs_grabl_tracing_artifacts = "artifacts")
 
 ##############################
-# Load @graknlabs_grakn_core #
+# Load @graknlabs_grakn_core_artifact #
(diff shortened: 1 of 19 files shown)


WORKSPACE
```diff
@@ -175,17 +175,10 @@ graknlabs_grabl_tracing()
...
 ##############################
-load("//dependencies/graknlabs:dependencies.bzl", "graknlabs_grakn_core")
-graknlabs_grakn_core()
-
-load("@graknlabs_grakn_core//dependencies/graknlabs:dependencies.bzl", "graknlabs_console")
-graknlabs_console()
-
-load("@graknlabs_grakn_core//dependencies/maven:dependencies.bzl",
-graknlabs_grakn_core_maven_dependencies = "maven_dependencies")
-graknlabs_grakn_core_maven_dependencies()
```
haikalpribadi, line 188: ❤️

PR #136, opened 2020-07-13
Test building with updated @graknlabs_dependencies

vmax:

Update `@graknlabs_dependencies` to latest version


adammitchelldev:
We have already updated to a more recent version of dependencies in another PR: #137 , so this is obselete.

PR #145, opened 2020-08-03
Add concept BDD tests and remove old integration tests

adammitchelldev:

We are now going to use verification's BDD tests to test client-java concept API, rather than rolling our own concept integration tests. This gives us more complete test coverage.

- Merged some things from master.
- Added BDDs (which are currently failing) and refactor to client-java so that they compile.
- Updated to use new 2.0 test runner from common.
- Updated to use private snapshot repo for grakn 2.0 artifact.


test/behaviour/connection/session/SessionSteps.java
```diff
@@ -40,27 +43,61 @@
...
+    @When("connection open (data )session(s) for database(s):")
+    public void connection_open_data_sessions_for_databases(List<String> names) {
         for (String name : names) {
             sessions.add(client.session(name));
         }
     }
 
-    @When("connection open sessions in parallel for keyspaces:")
-    public void connection_open_sessions_in_parallel_for_keyspaces(List<String> names) {
+    @When("connection open (data )sessions in parallel for databases:")
```
flyingsilverfin, line 70: will this work if there's a mismatch in the spacing in the BDD scenario definition?

test/integration/tracing/TracingIT.java
```diff
@@ -1,81 +0,0 @@
-/*
```
flyingsilverfin, line 1: so how come this test is gone as well? Now we have not tracing tests...

test/integration/answer/AnswerIT.java
```diff
@@ -1,332 +0,0 @@
-/*
```
flyingsilverfin, line 1: why is this test gone now? We don't have any tests for Answer now right? If we don't want it get we should @ignore it and then add it to the issue you're making with all the tests to re-enable

GraknClient.java
```diff
@@ -634,6 +634,15 @@ public void commit() {
             close();
         }
 
+        public ThingType.Remote<?, ?> getRootType() {
+            grakn.client.concept.type.Type.Remote<?> concept = getType(Label.of("thing")); // TODO do this properly
```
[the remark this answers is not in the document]
adammitchelldev: Not yet, MetaTypes work very differently on hypergraph and client-java and this is not in the scope of this PR, the idea of thing as the root type for things is actually different to how it was before.

Also, the existing client-java code did just this, I added the TODO only I think.

test/behaviour/connection/session/SessionSteps.java
adammitchelldev: no

test/integration/answer/AnswerIT.java
adammitchelldev: All of the integration tests were out of date and would be unbuildable, not just untestable. They can be migrated in at a later time right? They still exist in master.

adammitchelldev:
I have created issues for the 2.0 TODOs, which Alex will probably move on to implementing.

GraknClient.java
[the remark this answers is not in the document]
adammitchelldev: Okay, more specifically, this method is new, but this is the only way client-java is able to get types and is the same method as the other meta/root type getters.

test/integration/answer/AnswerIT.java
flyingsilverfin: ok good thanks for making issue for these

test/behaviour/connection/session/SessionSteps.java
flyingsilverfin: So where is the typo - here `(data )` ? or In verification do we need to change it to be correct `(data)`?

PR #149, opened 2020-08-05
Migrate Client Tests From Core

flyingsilverfin:
We migrate one test and implement BDD steps required for https://github.com/graknlabs/verification/pull/88 which migrated client-java tests from core to BDD as well.

* Implement new `TransactionSteps` methods
* implement new `KeyspaceSteps` method
* Migrate a concurrecny test to `AnswerIT` -- should be in ACID tests in the future


test/behaviour/graql/GraqlSteps.java
```diff
@@ -91,6 +89,13 @@ public void transaction_is_initialised() {
         assertTrue(tx.isOpen());
     }
 
+    @Given("transaction is closed and opened without commit")
+    public void transaction_is_reopened_without_commit() {
```
haikalpribadi, line 93: We should probably align the step syntax with the method name at all times, right? Here "reopen" feels better.

PR #150, opened 2020-08-07
Split Relation.players into 2 methods: one filtered by roles, and one that isn't; and replace Role with RoleType in Thing classes

alexjpwalker:

To fix Relation.players() always returning an empty collection, and to enforce consistent naming.

In https://github.com/graknlabs/grakn-2.0/pull/13, we updated the Grakn RPC to take in two different kinds of messages for retrieving roleplayers from a relation; one filtered by roles, and one that isn't.

Then we aligned client-java's `Relation` and `RelationImpl` with the new message format.

In doing this we realised that methods named `roles()` returning RoleTypes was weird and unclear, particularly when called from Things. So we decided to refactor these methods.


alexjpwalker:
Depends on protocol: see graknlabs/protocol#44

alexjpwalker:
Note: `test-behaviour` is currently failing - this is intentional and will be fixed with further, upcoming changes to client-java, protocol and grakn-2.0.

dependencies/graknlabs/repositories.bzl
```diff
@@ -37,14 +37,14 @@ def graknlabs_dependencies():
...
-        commit = "504561c1343b12cfb08cd65c25155045da0b709e", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_dependencies
+        commit = "bc2d9db7d2b5530ae90e76e6623b63f515cd326f", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_dependencies
     )
 
 def graknlabs_protocol():
     git_repository(
         name = "graknlabs_protocol",
-        remote = "https://github.com/graknlabs/protocol",
-        commit = "9f98160d1b3cbb2eac5d6cef5ea9a986cdb4d4a2", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
+        remote = "https://github.com/alexjpwalker/protocol",
```
alexjpwalker, line 46: Update to graknlabs once protocol is merged

rpc/RequestBuilder.java
```diff
@@ -182,7 +182,7 @@
             }
         }
 
-        public static Collection<ConceptProto.Concept> concepts(Collection<Concept<?>> concepts) {
```
alexjpwalker, line 185: The reason this shaky code was "working" before was because we never passed it a List - only an Array, and arrays in Java are not type checked as strictly as generic collections.

test/integration/concept/.gitkeep
```diff
@@ -0,0 +1 @@
+Placeholder for upcoming integration tests - delete this file once the directory has other files in it
```
alexjpwalker, line 1: Previously the directory was "kept alive" by an empty Java file. We felt that this was awkward and the right thing to do would be to use a .gitkeep instead.

haikalpribadi:
We'll need to address the change request in [here](https://github.com/graknlabs/protocol/pull/44#pullrequestreview-463819194) first before we proceed with this PR.

concept/thing/Relation.java
```diff
@@ -114,15 +114,23 @@ default boolean isRelation() {
...
          *
          * @param roles used to filter the returned instances only to ones that play any of the role types.
-         *              If blank, returns all role players.
-         * @return a list of every Thing involved in the Relation.
+         *
+         * @return a list of every Thing involved in the Relation, filtered by Role played.
          */
         @CheckReturnValue
-        Stream<Thing.Remote<?, ?>> players(RoleType... roles);
+        Stream<Thing.Remote<?, ?>> players(List<RoleType> roles);
```
haikalpribadi, line 133: Let's refer to `RoleType` as "Role Type" in docs, and `roleType` in method arguments. Let's do this consistently everywhere.

PR #156, opened 2020-08-11
Fix typo: "subs" vs. "sups", causing sups() to always fail

alexjpwalker:

To fix sups() always failing

There was a pretty unfortunate typo. We fixed it.


alexjpwalker:
Note: test-behaviour still fails - this is intentional and will be fixed in later PRs

PR #157, opened 2020-08-13
Align package structure, type hierarchy and Concept API with Core

alexjpwalker:

To make client-java beautiful and functional.

Substantially improving the code quality and usability of client-java by aligning the package structure, type hierarchy and Concept API with Core.

We broke down the 1400-line beast that was GraknClient, separating it into 6 files:

- an interface, named Grakn, containing sub-interfaces named Transaction, Session, Client, Database and DatabaseManager;
- an implementation file for each of the 5 sub-interfaces.

Then, we introduced a method Grakn.client() that instantiates a Grakn.Client. And we separated out some of the inner classes of Transaction, including GraknTransceiver and RpcIterator, while refactoring out QueryOptions completely in favour of the Options class that Grakn Core uses.

We aligned Concept API and its type hierarchy with Core, deleting, renaming and adding methods as appropriate; deleting `MetaType` in favour of `ThingType` (for 'thing'), `AttributeType` (for 'attribute'), and so on. We also deleted almost all generics from Concept API, which made the internal code more awkward, but the user-facing code cleaner. We also refactored out `Label` in favour of `String`, and `ConceptIID` in favour of another `String`.

Then, we refactored out the `rpc` package by moving Transceiver to `connection`, RequestBuilder to `connection.ConnectionMessageWriter` and ResponseReader to `answer.AnswerMessageReader`. We also refactored methods out of `ResponseReader` and into the `Answer` subtypes, to align the structure of `answer` with `concept`.

After that we introduced the `Concepts` interface into `concept`, made an implementation called `GraknConcepts` that takes in a `GraknTransceiver`, and refactored the entire `concept` package to only ever reference the `Concepts`, not a `Transaction`. By doing so, `concept` package no longer has a dependency on the root package.

We refactored out `GraknConceptException` and unified all Exceptions into `GraknClientException`, constructed using an `ErrorMessage` which extends the `ErrorMessage` in the **common** repository - similar to the way Core does it. This means that each exception thrown on the Client has a unique error code.


concept/connection/GraknConcepts.java
```diff
@@ -0,0 +1,249 @@
+/*
```
adammitchelldev, line 1: We decided we should move this file to the main `connection` package, which breaks a cyclic dependency between `connection` and `concepts`

connection/GraknTransaction.java
```diff
@@ -0,0 +1,635 @@
...
+    private final GraknTransceiver transceiver;
+
+    public static class Builder implements Transaction.Builder {
+
+        private final ManagedChannel channel;
+        private final Session session;
+        private final ByteString sessionId;
+        private final QueryOptions options;
+
+        Builder(final ManagedChannel channel, final Session session, final ByteString sessionId) {
```
adammitchelldev, line 84: Session can be changed to `GraknSession`, and then the `ManagedChannel` can be retrieved directly from session.

test/behaviour/concept/ConceptSteps.java
```diff
@@ -0,0 +1,12 @@
+package grakn.client.test.behaviour.concept;
```
adammitchelldev, line 1: Need to add a license header

adammitchelldev (changes requested):

concept/connection/GraknConcepts.java
alexjpwalker: Done

test/behaviour/concept/ConceptSteps.java
alexjpwalker: done

alexjpwalker:
Failing in CircleCI because ARTIFACT_USERNAME is not set, I intend to fix this in Grabl (but Grabl is not working right now)

adammitchelldev (approved):

flyingsilverfin (changes requested):
The `Options` should not set defaults, which are left to the backend

flyingsilverfin (approved):

haikalpribadi (approved):

PR #159, opened 2020-08-18
Add Automation YML

flyingsilverfin:
Implement Grabl Automation in client-java


.grabl/automation.yml
```diff
@@ -0,0 +1,98 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
```
lriuui0x0, line 2: We don't need this I think.

flyingsilverfin: you sure?

PR #164, opened 2020-09-01
More cleanup and bugfixes

alexjpwalker:

To continue cleaning up and fixing bugs in the client.

- Use grakn.common.util.Objects.className() instead of getCanonicalName() and getSimpleName()
- Change Thing.getRelations to return Relation.Locals
- Rename RPCClient to GraknClient and delete Grakn.client
- Store Remote attributes' values locally - avoids a server roundtrip when calling `getValue`
- Change `putEntityType` and `putRelationType` to return Local concepts
- Overhaul of our Error Messages - now cleanly divided into Client, Concept and Query errors, and each error code uniquely identifies each problem

### DIFF
--- a/concept/type/Type.java
+++ b/concept/type/Type.java
@@ -20,7 +20,7 @@
 package grakn.client.concept.type;
 
 import grakn.client.Grakn;
-import grakn.client.common.exception.GraknException;
+import grakn.client.common.exception.GraknClientException;
 import grakn.client.concept.Concept;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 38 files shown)


test/deployment/pom.xml
```diff
@@ -26,10 +26,6 @@ under the License.
     <version>1.0-SNAPSHOT</version>
     <name>test-deployment</name>
     <url>http://maven.apache.org</url>
-    <properties>
```
lolski, line 29: Does test-deployment work without it?

I think it is needed, otherwise `maven` will try default to compiling it with Java 7.

concept/type/Type.java
```diff
@@ -56,27 +57,27 @@
 
         @Override
         default ThingType.Local asThingType() {
-            throw new GraknException(INVALID_CONCEPT_CASTING.message(this, ThingType.class.getCanonicalName()));
+            throw new GraknClientException(INVALID_CONCEPT_CASTING.message(this, className(ThingType.class)));
```
haikalpribadi, line 60: I think this should be written as the following?
```
INVALID_CONCEPT_CASTING.message(className(this.getClass()), className(ThingType.class))
```

haikalpribadi (approved):
All looks good except for the INVALID_CONCEPT_CASTING message argument. Please fix in subsequent PR.

test/deployment/pom.xml
haikalpribadi: yes that's the point, @lolski . We do want Client Java to work with *any* Java version, and NOT just just 8+. I previously had to add this `properties` section to enforce Java+ because of the `Grakn` interface having a static method `Grakn.client(..)`, which is a Java 8 feature. We've removed that now, so users open client with `Grakn.Client client = new GraknClient(...);`. So we now we're backwards compatible and should remove the `properties` section`.

lolski: Alright makes sense

PR #165, opened 2020-09-04
Implement missing race condition fix from master

adammitchelldev:

We patched a race condition that caused transactions to hang during closing a while ago but the patch was never released and only exists in master. During debugging of a new issue, we rediscovered the same problem, but are now more sure about the nature of the bug and a more correct fix. However, we are also certain that the original fix has proven itself to be stable, so in the interest of keeping master the same as the latest release, we are going to release the original fix as 1.8.3.

- Add `synchronized` to transceiver methods to ensure that no calls result in a race condition.


flyingsilverfin:
Ok this works for now, but please make sure we link the proper fix in master to this PR so we know the real solution. I think it would be preferable that you submit and and write up the description in the PR on master @adammitchelldev 👍

flyingsilverfin (approved):

haikalpribadi (approved):

PR #168, opened 2020-09-07
Upgrade dependencies and credentials for remote builds

vmax:

Credential paths needs to be updated according to graknlabs/dependencies#196. In addition, `@graknlabs_graql` and `@graknlabs_common`are upgraded.

* Use `/opt/credentials/` as RBE credential path
* Upgrade Graql and Common
* Fix `ClientQueryTest` to use updated Graql syntax


test/integration/ClientQueryTest.java
```diff
@@ -217,7 +217,7 @@ public void applicationTest() {
             GraqlDelete deleteQuery = Graql.match(var("m").isa("mating")).delete(var("m").isa("mating"));
             LOG.info("clientJavaE2E() - '" + deleteQuery + "'");
             tx.execute(deleteQuery);
-            List<ConceptMap> matings = tx.execute(Graql.match(var("m").isa("mating")).get()).get();
+            List<ConceptMap> matings = tx.execute(Graql.match(var("m").isa("mating")).get("m")).get();
```
haikalpribadi, line 220: You can also just omit the `.get(...)` method call altogether in the new Graql, @vmax. `Graql.match(...)` will already return a `GraqlMatch` query. But I'll approve this for now and up to you if you wanna make the change. This is also fine.

haikalpribadi (approved):

PR #174, opened 2020-10-08
Query API

alexjpwalker:

To support Graql queries against Grakn 2.0 in client-java.

- Strip all the `execute` and `stream` methods from Transaction (there were like 30+ of them!!!) and replace with simply [define, undefine, match, insert, delete] in a new class, Query, which is accessed from Transaction by calling `tx.query()`. Implement the 5 Graql queries.
- Rules are no longer Concepts.
- RPCTransceiver is gone - its logic has been moved to RPCTransaction. All Concept **implementation** classes now store a reference to the RPCTransaction, obtained from Transaction by a simple cast. This allows us to get rid of the unwanted methods `iterateConceptMethod`, `runConceptMethod` that were previously exposed to the user as public methods of `Concepts`.
- The Answer types [ConceptList, ConceptSet, ConceptSetMeasure, Value and Void] are gone - leaving only AnswerGroup, ConceptMap, and Number. Value has been temporarily replaced with Number while we continue to flesh out the format of our aggregate Answers. Explanations are also gone.


concept/answer/Answer.java
```diff
@@ -29,26 +29,22 @@
...
-                return ConceptSet.of(res.getConceptSet());
-            case CONCEPTSETMEASURE:
-                return ConceptSetMeasure.of(res.getConceptSetMeasure());
-            case VALUE:
-                return Numeric.of(res.getValue());
-            case VOID:
-                return Void.of(res.getVoid());
             case ANSWER_NOT_SET:
                 throw new GraknClientException(MISSING_ANSWER.message(className(AnswerProto.Answer.AnswerCase.class)));
+            case NUMBER:
```
alexjpwalker, line 47: This is a temporary stopgap - the actual format of aggregate answers has not been decided yet.

rpc/QueryFuture.java
```diff
@@ -0,0 +1,12 @@
+package grakn.client.rpc;
+
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+public interface QueryFuture<T> extends Future<T> {
```
alexjpwalker, line 6: This class (and others) used to be inner classes of Grakn.Transaction.

test/behaviour/graql/GraqlSteps.java
```diff
@@ -105,189 +70,96 @@ public void integrity_is_validated() {
...
-            tx.commit();
-        } catch (RuntimeException e) {
-            threw = true;
-        } finally {
-            tx.close();
-            tx = session.transaction(WRITE);
-        }
-        assertTrue(threw);
+    public void graql_query(final String graqlQueryStatements) {
+        // TODO: re-enable when match is implemented
```
alexjpwalker, line 130: This, and many other test methods, depend on `graql match` in order to work.
To allow our tests to pass and to get meaningful results, we have temporarily commented them out. But we must remember to re-enable them once match is implemented.

lolski (approved):
Looks good to me. The simplification makes sense.

rpc/response/Response.java
```diff
@@ -0,0 +1,121 @@
...
+            Response that = (Response) o;
+            return ((this.nullableOk == null) ? (that.nullableOk() == null) : this.nullableOk.equals(that.nullableOk()))
+                    && ((this.nullableError == null) ? (that.nullableError() == null) : this.nullableError.equals(that.nullableError()));
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int h = 1;
```
flyingsilverfin, line 110: just use Objects.hash(this.nullableOk, this.nullableError)

rpc/response/Response.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == this) {
+            return true;
+        }
+        if (o instanceof Response) {
+            Response that = (Response) o;
+            return ((this.nullableOk == null) ? (that.nullableOk() == null) : this.nullableOk.equals(that.nullableOk()))
```
flyingsilverfin, line 102: Trick here that might be useful to avoid doing null checks is `Objects.equals(a, b)`, which performs null checks, etc.

flyingsilverfin (changes requested):

concept/Concepts.java
```diff
@@ -127,23 +118,30 @@ public AttributeType getAttributeType(final String label) {
...
+    public Thing getThing(final String iid) {
+        final TransactionProto.Transaction.Req req = TransactionProto.Transaction.Req.newBuilder()
+                .putAllMetadata(tracingData())
+                .setGetThingReq(TransactionProto.Transaction.GetThing.Req.newBuilder().setIid(iid(iid))).build();
+
+        final TransactionProto.Transaction.Res response = rpcTransaction.execute(req);
+        switch (response.getGetThingRes().getResCase()) {
+            case THING:
+                return ThingImpl.of(response.getGetThingRes().getThing());
+            default:
```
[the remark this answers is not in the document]
alexjpwalker: It's functionally the same whichever way round they are

concept/type/Rule.java
```diff
@@ -25,23 +25,26 @@
 import javax.annotation.Nullable;
 import java.util.stream.Stream;
 
-public interface Rule extends Type {
+public interface Rule {
+
+    String getLabel();
+
+    Pattern getWhen();
```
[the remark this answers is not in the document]
alexjpwalker: Hmm okay. Currently we use `Graql.parsePattern(ruleProto.getWhen())` to create a Pattern on the client side.

How would we parse a `String` into a `Conjunction<? extends Pattern>`? `Graql.and(Graql.parsePattern(ruleProto.getWhen()))`?

rpc/response/Response.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private static Response create(@Nullable TransactionProto.Transaction.Res response, @Nullable StatusRuntimeException error) {
+        if (!(response == null || error == null)) {
+            throw new GraknClientException(new IllegalArgumentException("One of Transaction.Res or StatusRuntimeException must be null"));
+        }
+        return new Response(response, error);
+    }
+
+    static Response completed() {
+        return create(null, null);
```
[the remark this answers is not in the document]
alexjpwalker: I'm not sure that's avoidable - very common in Java for `null` to be a "default" value of a constructor or pseudo-constructor parameter.

I can see another clear issue in this file though - the constructor is almost identical to `create`, so we can refactor out `create`.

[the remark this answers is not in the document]
alexjpwalker: So there is an fact an enum, Type, which is `OK`, `ERROR` or `COMPLETED`, and it's determined by the states of nullableOk and nullableError.

To be honest I dislike the names nullableOk and nullableError - maybe now is the time to fix that

rpc/response/Response.java
alexjpwalker: done!

rpc/response/Response.java
```diff
@@ -0,0 +1,121 @@
...
+            return Type.COMPLETED;
+        }
+    }
+
+    public final TransactionProto.Transaction.Res ok() {
+        if (nullableOk != null) {
+            return nullableOk;
+        } else if (nullableError != null) {
+            // TODO: parse different GRPC errors into specific GraknClientException
+            throw new GraknClientException(nullableError);
```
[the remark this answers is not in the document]
alexjpwalker: I don't fully understand what "completed" means yet. This might be important when rearchitecting this package.

rpc/response/Response.java
alexjpwalker: Haha IntelliJ just suggested the same thing

flyingsilverfin (approved):

haikalpribadi (approved):

PR #178, opened 2020-10-28
2nd RPC Rework: Client

alexjpwalker:

To add support for concurrent operations within a single transaction, and optimise Grakn in scenarios with high network latency.

Also, reorganise and simplify the RPC package in order to make Grakn clients easier to implement.

- New structure for RPC classes:
```
| GraknClient
| RPCDatabaseManager
| RPCSession
| RPCTransaction
|____| Response
|____|____| Ok
|____|____| Error
|____| ResponseCollector
|____|____| Single
|____|____| Multiple
| QueryFuture
| QueryIterator
```
- New data structure for managing requests and responses within a Transaction; `RPCTransaction` now holds a `ConcurrentMap<UUID, ResponseCollector>` that maps requests (now identified by a unique UUID) onto response collectors (which can be either `Single` or `Multiple`), allowing for concurrent operations to occur within a single transaction
- The network latency is now sent to the server when opening iteration requests to allow the server to use an adaptive batch size
- The protocol is flattened and simplified
- QueryFutures issue requests when constructed and return responses on calling `.get`
- QueryIterator (formerly RPCIterator) no longer blocks the current thread on construction
- RPC state objects (Session, Transaction) are now closed atomically (using `AtomicBoolean`)
- Cleaned out any remaining interface defaults (eg. in Grakn)
- The default server address is now localhost:1729


alexjpwalker:
This PR depends on https://github.com/graknlabs/grakn-2.0/pull/59 in order to build in CI, and should not be merged until this is done.

rpc/RPCTransaction.java
```diff
@@ -135,45 +129,159 @@ public void close() {
...
-            collectorQueue.add(collector);
-            return new QueryFuture.Stream<>(iterator);
+    public <T> QueryFuture<Stream<T>> iterateAsync(TransactionProto.Transaction.Req.Builder request, Function<TransactionProto.Transaction.Res, T> transformResponse) {
+        try (ThreadTrace ignored = traceOnThread("iterateAsync")) {
+            final ResponseCollector.Multiple responseCollector = new ResponseCollector.Multiple();
+            final UUID requestId = UUID.randomUUID();
+            request.setId(requestId.toString());
+            request.setLatencyMillis(networkLatencyMillis);
+            collectors.put(requestId, responseCollector);
+            return new QueryFuture.Stream<>(request.build(), requestObserver, responseCollector, transformResponse);
```
alexjpwalker, line 157: There is an amount of asymmetry between these methods which is far from ideal.

There are a couple of roadblocks here:
- `Future` doesn't contain a `map` method for returning a new `Future` with the result transformed. We could make one, but given that `Future` is a Java-specific thing, I don't think we should overcomplicate our `QueryFuture`. This means we cannot simply take `transformResponse` out of `executeAsync`.
- Adding `transformResponse` to `execute` would make calls to `execute` sharply less readable. Take one example:
```
return rpcTransaction.execute(request).getThingRes();
```
becomes
```
return rpcTransaction.execute(request, TransactionProto.Transaction.Res::getThingRes);
```
Adding `transformResponse` to `execute` is surely less hideous than taking it out of `executeAsync` and creating `QueryFuture.map`, but given that there are many calls to `execute` and only one definition of `execute`, I would rather that the calls looked nice.

rpc/RPCTransaction.java
```diff
@@ -135,45 +129,159 @@ public void close() {
...
+            }
+
+            @Override
+            public String toString() {
+                return className(getClass()) + "{" + error + "}";
+            }
+        }
+    }
+
+    abstract static class ResponseCollector {
```
alexjpwalker, line 240: I'm struggling with naming this class. Is it a `ResponseCollector`? A `ResponseBuffer`? A `ResponseBucket`? A `ResponseQueue`?

Conceptually, we maintain a map of Requests to Response buckets. Whenever a Response comes in it gets added to the correct bucket.

Internally, the bucket is implemented as a response _buffer_. Theoretically, we could limit the buffer size to perform flow control.

But then comes the issue of differentiating a Single response collector (that will only ever collect one response) and a Multiple response collector (that may receive multiple responses). What is a `ResponseBuffer.Single` supposed to mean? In this case `ResponseCollector.Single` seems the clearest. How many responses do we collect? Just a single one.

concept/ConceptManager.java
```diff
@@ -73,12 +74,10 @@ public AttributeType getRootAttributeType() {
     }
```
alexjpwalker, line 74: The `concept` changes are pretty much entirely just renaming due to updated `protocol`.

rpc/RPCTransaction.java
flyingsilverfin: I'm starting to feel that the `ResponseCollector` is correct - "collecting" has a notion of collecting something into something else already. I think the API would be a bit clearer if you rename the `put` method to `collect`. Then you get:
```
collector.collect(response
```
vs
```
collector.put(response)
```
i think the former has more story to it.

alternatives we might consider are also `collector.record(response)` or `collector.buffer(response)`, which convey more of the semantic of saving responses somewhere to be used later!

flyingsilverfin: I think if its called `ResponseCollector` it should actually be `.collect()`. if its `ResponseBuffer` it shuold be `.buffer()` etc. 

overall collector seems about right

rpc/QueryIterator.java
```diff
@@ -0,0 +1,73 @@
...
+            case DONE:
+                return endOfData();
+            case CONTINUE:
+                final TransactionProto.Transaction.Req continueReq = TransactionProto.Transaction.Req.newBuilder()
+                        .setId(requestId.toString()).setContinue(true).build();
+                requestObserver.onNext(continueReq);
+                return computeNext();
+            case RES_NOT_SET:
+                throw new GraknClientException(MISSING_RESPONSE.message(className(TransactionProto.Transaction.Res.class)));
+            default:
```
flyingsilverfin, line 69: does default correspond to an actual case? or is it NOT_SET? because that's the above case, no?

flyingsilverfin: basically hard for me to tell if this can possibly recurse infinitely in the CONTINUE case, and when its not continuing from this code :) however, i may just be missing an acceptable level of context for this one

concept/type/impl/TypeImpl.java
```diff
@@ -272,28 +281,28 @@ public final boolean isDeleted() {
...
+            final TransactionProto.Transaction.Req.Builder request = TransactionProto.Transaction.Req.newBuilder()
+                    .setTypeReq(method.setLabel(label));
+            return rpcTransaction.execute(request).getTypeRes();
         }
 
-        /*TypeMethod.Res iterate(final TypeMethod.Iter.Req.Builder method) {
-            final TransactionProto.Transaction.Iter.Req request = TransactionProto.Transaction.Iter.Req.newBuilder()
-                    .setTypeMethodIterReq(method.setLabel(label)).build();
-            return rpcTransaction.iterate(request).
+        /*ConceptProto.Type.Res iterate(final ConceptProto.Type.Req.Builder method) {
```
flyingsilverfin, line 302: meant to be commented out?

flyingsilverfin (changes requested):

concept/type/impl/TypeImpl.java
alexjpwalker: I originally left it in there because it seemed likely to be needed in the future - but in fact, it actually doesn't and I've now deleted it.

rpc/QueryIterator.java
alexjpwalker: In theory, if we connected to a broken Grakn server that just continuously sent back `CONTINUE` responses, then yes this would recurse infinitely. On each iteration we `.take` from the `ResponseCollector` which is implemented as a `BlockingQueue`.

`default:` corresponds to any "standard" response - e.g. a `Query.Res` containing a query answer, a `Type.Res` containing a single instance of a `ThingType`... and so on. Essentially it is the meat of the iterator, with all other cases being metadata used to handle the internal iteration implementation.

rpc/RPCTransaction.java
```diff
@@ -135,45 +129,157 @@ public void close() {
...
+                if (collector == null) throw new GraknClientException(UNKNOWN_REQUEST_ID.message(requestId));
+                collector.put(new Response.Ok(res));
+                if (collector.isDone()) collectors.remove(requestId);
+            }
+
+            @Override
+            public void onError(Throwable error) {
+                assert error instanceof StatusRuntimeException : "The server only yields these exceptions";
+                // TODO: is it desirable behaviour to kill all the pending requests?
+                // TODO: this isn't nice in any case - the error for other pending requests is misleading
```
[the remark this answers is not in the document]
alexjpwalker: No I think we should talk about them

rpc/RPCTransaction.java
alexjpwalker: Hmmm. I don't really like `collect`. `ResponseCollector` roughly describes the big picture of what the thing does over time, over the course of multiple `put`s. I want a name that makes it clear that this is adding a single response to the collector.

The old name was `onResponse`, but `add` seems good, `put` too, `receive` would be OK.

alexjpwalker: We settled on `add` in the end.

rpc/RPCTransaction.java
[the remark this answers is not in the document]
alexjpwalker: Raised https://github.com/graknlabs/client-java/issues/180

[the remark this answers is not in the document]
alexjpwalker: This is now resolved.

The correct behaviour is that both `onError` and `onCompleted` should send an error signal to all consumers which will cause them to throw an exception if they attempt to keep consuming. `onCompleted` can occur as the result of a successful `commit` on the server. This will cause any response collectors open on the client side to cease to be able to receive further responses. So we send them just one more response, which when the user attempts to read it, throws an exception saying the transaction has been closed. Also we set `isOpen` to false, which will prevent the RPCTransaction attempting to close the RPC stream that has already been closed.

Nonetheless the error in `onError` may not be appropriate for these failing consumers - see #180 .

rpc/QueryIterator.java
flyingsilverfin: what kind of res case is set in the default case?

alexjpwalker: Followed up on Discord

flyingsilverfin (approved):

common/exception/GraknClientException.java
```diff
@@ -37,8 +35,7 @@ public GraknClientException(final ErrorMessage error) {
     }
 
     public GraknClientException(final StatusRuntimeException statusRuntimeException) {
-        super(statusRuntimeException.getMessage(), statusRuntimeException);
-        this.statusCode = statusRuntimeException.getStatus().getCode().name();
+        super(statusRuntimeException.getStatus().getDescription());
```
alexjpwalker, line 38: This change causes "INTERNAL:" to no longer be part of the error message.

concept/Concept.java
```diff
@@ -44,20 +44,10 @@
     Remote asRemote(Grakn.Transaction transaction);
 
     @CheckReturnValue
-    default boolean isRemote() {
-        return false;
-    }
+    boolean isRemote();
```
alexjpwalker, line 47: Cleaning out interface defaults to keep interfaces clean

dependencies/graknlabs/artifacts.bzl
```diff
@@ -17,14 +17,15 @@
...
 load("@graknlabs_dependencies//distribution:deployment.bzl", "deployment_private")
-def graknlabs_grakn_core_artifact():
-    artifact_file(
+
+def graknlabs_grakn_core_artifacts():
+    native_artifact_files(
         name = "graknlabs_grakn_core_artifact",
         group_name = "graknlabs_grakn_core",
-        artifact_name = "grakn-core-server-linux-{version}.tar.gz",
+        artifact_name = "grakn-core-server-{platform}-{version}.tar.gz",
```
alexjpwalker, line 27: This macro expands to two `artifact_file` definitions that pull in `grakn-core-server-linux` and `grakn-core-server-mac`. When you run test targets, the correct artifact for your machine's OS is selected.

query/QueryManager.java
```diff
@@ -48,78 +44,60 @@ public QueryManager(final RPCTransaction rpcTransaction) {
         this.rpcTransaction = rpcTransaction;
     }
 
-    public QueryFuture<Stream<ConceptMap>> match(final GraqlMatch query) {
+    public Stream<ConceptMap> match(final GraqlMatch query) {
```
alexjpwalker, line 47: We change `QueryFuture<Stream>` to just `Stream`, because `Stream` provides us the behaviour of "send a request and return answers only when you specifically ask for them (by iterating)" that we want.

rpc/QueryFuture.java
```diff
@@ -21,98 +21,59 @@
...
-
-    @Override
-    public abstract T get(long timeout, TimeUnit unit);
-
-    static class Execute<T> extends QueryFuture<T> {
-        private final RPCResponseAccumulator collector;
-        private final Function<TransactionProto.Transaction.Res, T> responseReader;
+public class QueryFuture<T> implements Future<T> {
 
-        public Execute(final RPCResponseAccumulator collector, final Function<TransactionProto.Transaction.Res, T> responseReader) {
```
alexjpwalker, line 41: Now that `QueryFuture.Stream` is deleted, `Execute` becomes the only concrete subclass of `QueryFuture`, so we simply dissolve `Execute` and just use `QueryFuture` itself.

rpc/RPCTransaction.java
```diff
@@ -23,82 +23,86 @@
...
-
     @Override
     public Type type() {
         return type;
     }
 
     @Override
     public boolean isOpen() {
-        return isOpen.get();
+        return transactionWasOpened.get() && !transactionWasClosed.get();
```
alexjpwalker, line 105: Technically nothing prevents open and close being triggered out of order, so we record both events.

test/behaviour/connection/database/BUILD
```diff
@@ -59,7 +62,8 @@ grakn_test(
     data = [
         "@graknlabs_behaviour//connection:database.feature",
     ],
-    grakn_artifact = "@graknlabs_grakn_core_artifact//file",
+    grakn_artifact_linux = "@graknlabs_grakn_core_artifact_linux//file",
+    grakn_artifact_mac = "@graknlabs_grakn_core_artifact_mac//file",
```
alexjpwalker, line 66: Now, we selectively load the artifact appropriate for the host OS.

test/behaviour/concept/thing/attribute/BUILD
```diff
@@ -62,7 +65,8 @@ grakn_test(
     data = [
         "@graknlabs_behaviour//concept/thing:attribute.feature",
     ],
-    grakn_artifact = "@graknlabs_grakn_core_artifact//file",
+    grakn_artifact_linux = "@graknlabs_grakn_core_artifact_linux//file",
+    grakn_artifact_mac = "@graknlabs_grakn_core_artifact_mac//file",
```
alexjpwalker, line 69: We load the artifact appropriate for the host OS.

test/behaviour/concept/thing/entity/BUILD
```diff
@@ -19,7 +19,7 @@
```
alexjpwalker, line 19: The rest of these test files are just identical changes to the above - they are not interesting

alexjpwalker (commented):
Annotated

haikalpribadi (approved):

Issue #181, opened 2020-11-17
TransactionTest intermittently stalls


alexjpwalker:
## Description

TransactionTest intermittently stalls.

## Environment

1. OS (where Grakn server runs): Mac OS 10
2. Grakn version (and platform): Grakn Core 2.0
 
## Additional information

It seems to occur about 1 in 5 runs when run as part of `bazel test //test/behaviour/connection/...`. I haven't managed to reproduce the issue when running it by itself, but that might just be coincidence.

PR #184, opened 2020-11-20
Fix thread safety, data & memory usage issues

alexjpwalker:

To fix a number of issues causing intermittent test failures.

- Atomically modify the `ResponseCollector` map in `RPCTransaction` (enforced using the `AtomicBoolean` `transactionWasClosed`). This fixes an issue where `TransactionTest` would intermittently stall.
- Don't call `close()` after commit - it's better for the server to take responsibility for closing.
- Declare `native_grakn_artifact` explicitly and in one place. This fixes an issue where the build would fail in CI because it was trying to upload Grakn 32 times, once for each test.
- Add `--jobs=1` to `automation.yml` when running `bazel test //test/behaviour/connection/...`. This fixes an issue where this test would intermittently fail in CI due to Grabl not having enough memory to handle 24 concurrent RocksDB databases.

### DIFF
--- a/.grabl/automation.yml
+++ b/.grabl/automation.yml
@@ -61,9 +61,10 @@ build:
         export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
         bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
         bazel build //...
-        bazel test //test/behaviour/connection/... --test_output=errors
-        bazel test //test/behaviour/concept/... --test_output=errors
-        bazel test //test/behaviour/graql/language/define/... --test_output=errors
+        sleep 1800
+#        bazel test //test/behaviour/connection/... --test_output=errors
+#        bazel test //test/behaviour/concept/... --test_output=errors
+#        bazel test //test/behaviour/graql/language/define/... --test_output=errors
       # TODO: use --config=rbe once Grakn Core runs in RBE
       # TODO: revert above tests to //test/behaviour/...
     deploy-maven-snapshot:
--- a/.grabl/automation.yml
+++ b/.grabl/automation.yml
@@ -31,40 +31,38 @@ build:
           --project-key=graknlabs_client_java \
           --branch=$GRABL_BRANCH --commit-id=$GRABL_COMMIT
   correctness:
-#    build:
-#      machine: graknlabs-ubuntu-20.04
-#      script: |
-#        export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
-#        export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
-#        bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
-#        bazel build --config=rbe //...
-#        bazel run @graknlabs_dependencies//tool/checkstyle:test-coverage
-#        bazel test --config=rbe $(bazel query 'kind(checkstyle_test, //...)')
-#    build-dependency:
-#      machine: graknlabs-ubuntu-20.04
-#      script: |
-#        dependencies/maven/update.sh
-#        git diff --exit-code dependencies/maven/artifacts.snapshot
-#        bazel run @graknlabs_dependencies//tool/unuseddeps:unused-deps -- list
-#    test-integration:
-#      machine: graknlabs-ubuntu-20.04
-#      script: |
-#        export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
-#        export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
-#        bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
-#        bazel test //test/integration/... --test_output=errors
-#      # TODO: use --config=rbe once Grakn Core runs in RBE
+    build:
+      machine: graknlabs-ubuntu-20.04
+      script: |
+        export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
+        export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
+        bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
+        bazel build --config=rbe //...
+        bazel run @graknlabs_dependencies//tool/checkstyle:test-coverage
+        bazel test --config=rbe $(bazel query 'kind(checkstyle_test, //...)')
+    build-dependency:
+      machine: graknlabs-ubuntu-20.04
+      script: |
+        dependencies/maven/update.sh
+        git diff --exit-code dependencies/maven/artifacts.snapshot
+        bazel run @graknlabs_dependencies//tool/unuseddeps:unused-deps -- list
+    test-integration:
+      machine: graknlabs-ubuntu-20.04
+      script: |
+        export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
+        export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
+        bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
+        bazel test //test/integration/... --test_output=errors
+      # TODO: use --config=rbe once Grakn Core runs in RBE
     test-behaviour:
       machine: graknlabs-ubuntu-20.04
       script: |
         export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
         export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
         bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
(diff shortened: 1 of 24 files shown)


rpc/RPCTransaction.java
```diff
@@ -121,12 +118,6 @@ public void commit() {
                 .putAllMetadata(tracingData())
                 .setCommitReq(TransactionProto.Transaction.Commit.Req.getDefaultInstance());
         execute(commitReq);
-        // TODO: Currently this close request is parsed as an "unknown frame" on the server because the server
-        //  terminates the stream when commit is called. Three possible solutions are:
-        //  1) don't close the transaction, or the stream, on commit
-        //  2) let the server take full responsibility for closing the stream on commit
-        //  3) let the client take full responsibility for closing the stream on commit
-        close();
```
alexjpwalker, line 129: This call to `close()` is unneeded as the server closes the transaction itself on a successful commit. Essentially this implements solution (2) from the comments: `let the server take full responsibility for closing the stream on commit`.

The client should be as dumb as possible; this change helps to achieve that.

rpc/RPCTransaction.java
```diff
@@ -165,12 +157,19 @@ public void close() {
...
             request.setId(requestId.toString());
             request.setLatencyMillis(networkLatencyMillis);
-            collectors.put(requestId, responseCollector);
+            addCollectorIfTransactionIsOpen(requestId, responseCollector);
             final QueryIterator<T> queryIterator = new QueryIterator<>(request.build(), requestObserver, responseCollector, transformResponse);
             return StreamSupport.stream(((Iterable<T>) () -> queryIterator).spliterator(), false);
         }
     }
 
+    private void addCollectorIfTransactionIsOpen(UUID requestId, ResponseCollector collector) {
```
alexjpwalker, line 166: We use `compute` to atomically read our `AtomicBoolean` and add the new `ResponseCollector`.

rpc/RPCTransaction.java
```diff
@@ -139,10 +130,11 @@ public void rollback() {
 
     @Override
     public void close() {
-        if (streamIsOpen.compareAndSet(true, false)) {
+        if (transactionWasClosed.compareAndSet(false, true)) {
```
alexjpwalker, line 133: We use our `AtomicBoolean` to ensure `collectors` gets purged once and only once, and that after the purge has started, no more collectors can be added to it.

The purge flushes all the collectors' `BlockingQueue`s by sending them an error response. This, in turn, causes all `QueryFutures` and `Stream`s to immediately throw `TRANSACTION_CLOSED`.

Note that `close` can be triggered either by the client calling `close`, or the server terminating the transaction stream (e.g. after a commit)

alexjpwalker:
Currently debugging the Grabl failures. Builds fail intermittently because of an issue in RBE; tests fail intermittently which appears to be due to the CI server holding too many network connections open concurrently (including the gRPC connections to the Grakn Servers run in the behaviour tests)

.grabl/automation.yml
```diff
@@ -60,9 +60,10 @@ build:
         export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
         export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
         bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
-        bazel test //test/behaviour/connection/... --test_output=errors
+        bazel test //test/behaviour/connection/... --test_output=errors --jobs=1
```
alexjpwalker, line 63: Because `SessionTest` and `DatabaseTest` each create 12 databases, running the two in parallel can cause the two Grakn processes combined to use over 2.1GB of memory. This is more than Grabl has.

We work around this limitation by executing the tests serially using `--jobs=1`.

dependencies/graknlabs/repositories.bzl
```diff
@@ -29,8 +29,8 @@ def graknlabs_graql():
 def graknlabs_common():
     git_repository(
         name = "graknlabs_common",
-        remote = "https://github.com/graknlabs/common",
-        commit = "39664389f458124c14fa307ab78619d17280b73d" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_common
+        remote = "https://github.com/alexjpwalker/common",
+        commit = "011c48e5c1a6a3d3ecec2176e223ca75baeb3403" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_common
```
alexjpwalker, line 33: This is https://github.com/graknlabs/common/pull/60 - let's merge `common` before merging this `client-java` PR.

test/BUILD
```diff
@@ -17,9 +17,17 @@
...
+load("@graknlabs_common//test/server:rules.bzl", "native_grakn_artifact")
 load("@graknlabs_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
 load("@graknlabs_bazel_distribution//artifact:rules.bzl", "artifact_extractor")
 
+native_grakn_artifact(
+    name = "native-grakn-artifact",
+    mac_artifact = "@graknlabs_grakn_core_artifact_mac//file",
+    linux_artifact = "@graknlabs_grakn_core_artifact_linux//file",
+    visibility = ["//test:__subpackages__"],
+)
```
alexjpwalker, line 29: This change depends on the new `common` and ensures that the Grakn artifact Bazel target is only declared once. Previously, it was redeclared in each test.

rpc/RPCTransaction.java
```diff
@@ -165,12 +157,19 @@ public void close() {
...
             final QueryIterator<T> queryIterator = new QueryIterator<>(request.build(), requestObserver, responseCollector, transformResponse);
             return StreamSupport.stream(((Iterable<T>) () -> queryIterator).spliterator(), false);
         }
     }
 
+    private void assertOpenAndAddCollector(UUID requestId, ResponseCollector collector) {
+        collectors.compute(requestId, (id, col) -> {
+            if (transactionWasClosed.get()) throw new GraknClientException(TRANSACTION_CLOSED);
+            return collector;
+        });
```
alexjpwalker, line 170: Here we atomically verify that the transaction has not been closed before retrieving a `ResponseCollector` from our `ConcurrentMap`.

flyingsilverfin (approved):

dependencies/graknlabs/repositories.bzl
```diff
@@ -30,21 +30,21 @@ def graknlabs_common():
...
-        commit = "30e30cec9e3fe4821103cafbd240ee9862e262ea" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_bazel_distribution
+        commit = "ba4a7d6e6fa98fcb44e7c6eae492aeeb9d85001b" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_bazel_distribution
     )
 
 def graknlabs_dependencies():
     git_repository(
         name = "graknlabs_dependencies",
         remote = "https://github.com/graknlabs/dependencies",
-        commit = "e282f9248b65e43467903aef492c32797cd30942", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_dependencies
+        commit = "63ea26dabe7320dc7dd191a5f3bfa93b60dca38b", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_dependencies
```
alexjpwalker, line 47: `deploy_github` rule has a `draft` field in it now which is only present in the latest `dependencies`.

rpc/RPCTransaction.java
```diff
@@ -186,65 +182,42 @@ public void onNext(TransactionProto.Transaction.Res res) {
...
+            map.remove(requestId);
+        }
 
-            @Override
-            public String toString() {
-                return className(getClass()) + "{" + error + "}";
-            }
+        private synchronized void clearWithError(Response.Error errorResponse) {
+            map.values().parallelStream().forEach(x -> x.add(errorResponse));
+            map.clear();
```
alexjpwalker, line 220: The main reason for synchronization is the possibility of the transaction becoming closed immediately after calling `if (transactionWasClosed.get())`.

`TransactionTest` was stalling intermittently because the original implementation of response collectors (just a plain old `ConcurrentHashMap`) was not being modified in a thread-safe way. It was possible that we would add a collector and, concurrently, close the transaction which initiates a "tear-down" process, namely `clearWithError`.

Given that we therefore can't allow `put` and `clearWithError` (and presumably `remove`) to run concurrently, I believe it makes sense to extract `ResponseCollectors` as an inner class, with these three basic methods marked as `synchronized`. If we didn't use an inner class, we would have to synchronize on some arbitrary object - which would be ugly.

alexjpwalker:
This PR fixes #181 , #186 and #187 .

lolski (approved):

PR #185, opened 2020-11-24
Receiving batched answers from one GRPC message

lriuui0x0:

This PR is complementary to https://github.com/graknlabs/grakn-2.0/pull/91

Too many GRPC calls cause performance problem which we decided to solve by allowing multiple answers into one message. We need to update client accordingly so that we can receive multiple answers in one message.

- Update protocol and fix relevant code
- Change `QueryIterator` to save an `iterator` field so that we will exhaust the answers in one message before we process the next message.

### DIFF
--- a/.grabl/automation.yml
+++ b/.grabl/automation.yml
@@ -60,7 +60,7 @@ build:
         export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
         export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
         bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
-        bazel test //test/behaviour/connection/... --test_output=errors
+        bazel test //test/behaviour/connection/... --test_output=errors --jobs=1
         bazel test //test/behaviour/concept/... --test_output=errors
         bazel test //test/behaviour/graql/language/define/... --test_output=errors
       # TODO: use --config=rbe once Grakn Core runs in RBE
--- a/concept/thing/impl/ThingImpl.java
+++ b/concept/thing/impl/ThingImpl.java
@@ -40,6 +40,7 @@ import grakn.protocol.ConceptProto.Thing.UnsetHas;
 import grakn.protocol.TransactionProto;
 
 import java.util.Arrays;
+import java.util.List;
 import java.util.Objects;
 import java.util.function.Function;
 import java.util.stream.Stream;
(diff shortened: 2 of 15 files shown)


concept/thing/impl/ThingImpl.java
```diff
@@ -297,16 +298,16 @@ public final boolean isRemote() {
             return rpcTransaction;
         }
 
-        Stream<ThingImpl> stream(ConceptProto.Thing.Req.Builder method, Function<ConceptProto.Thing.Res, ConceptProto.Thing> thingGetter) {
+        Stream<ThingImpl> stream(ConceptProto.Thing.Req.Builder method, Function<ConceptProto.Thing.Res, List<ConceptProto.Thing>> thingGetter) {
```
lriuui0x0, line 301: Here each `res` may contain a list of answers, so we change the signature (and all similar functions) to accept to transform `res` to a list of items instead of just one.

rpc/QueryIterator.java
```diff
@@ -48,6 +51,10 @@
 
     @Override
     protected T computeNext() {
+        if (currIterator != null && currIterator.hasNext()) {
```
lriuui0x0, line 54: Here we need to exhaust all of the answers in the current message before proceed to the next message.

rpc/QueryIterator.java
```diff
@@ -67,7 +74,12 @@ protected T computeNext() {
             case RES_NOT_SET:
                 throw new GraknClientException(MISSING_RESPONSE.message(className(TransactionProto.Transaction.Res.class)));
             default:
-                return transformResponse.apply(res);
+                currIterator = transformResponse.apply(res).iterator();
+                if (currIterator.hasNext()) {
+                    return currIterator.next();
+                } else {
+                    return computeNext();
```
lriuui0x0, line 81: Although the server will not send response with no answer (it will just send `done` message in that case), I feel it's still nicer to have client handle this case by simply calling `computeNext`.

rpc/QueryIterator.java
```diff
@@ -67,7 +74,12 @@ protected T computeNext() {
             case RES_NOT_SET:
                 throw new GraknClientException(MISSING_RESPONSE.message(className(TransactionProto.Transaction.Res.class)));
             default:
-                return transformResponse.apply(res);
+                currIterator = transformResponse.apply(res).iterator();
```
alexjpwalker, line 77: I think we can simplify the `default` block to just:
```java
currIterator = transformResponse.apply(res).iterator();
return computeNext();
```
because the `if` check is already handled on line 54.

alexjpwalker (changes requested):
I'm delighted that we've managed to make this change (which looked conceptually complex) with such a simple, clean refactor on the client side. Great work!

I've suggested one improvement to `QueryIterator`.

alexjpwalker (approved):

.grabl/automation.yml
```diff
@@ -60,7 +60,7 @@ build:
         export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
         export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
         bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
-        bazel test //test/behaviour/connection/... --test_output=errors
+        bazel test //test/behaviour/connection/... --test_output=errors --jobs=1
```
lolski, line 63: Is this set in order to avoid creating too many databases concurrently, causing memory overflow? @lriuui0x0

lolski: Spoke verbally with @lriuui0x0 and the answer is yes

lolski (approved):

PR #189, opened 2020-11-30
Unused code cleanup and streaming fixes

alexjpwalker:

To clean up redundant code and fix errors when performing streaming requests.

- Clear unused methods from `concept` package
- Fix error in `RPCTransaction` when the server sends a `DONE` response and then receives a `CONTINUE` request from the client - this is, in fact, the expected flow and should not trigger an error.
- Fix `setLabel` not updating the `label` locally.

### DIFF
--- a/concept/type/impl/TypeImpl.java
+++ b/concept/type/impl/TypeImpl.java
@@ -143,7 +143,7 @@ public abstract class TypeImpl implements Type {
     public abstract static class Remote implements Type.Remote {
 
         final RPCTransaction rpcTransaction;
-        private final String label;
+        private String label;
         private final boolean isRoot;
         private final int hash;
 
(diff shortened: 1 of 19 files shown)


concept/type/impl/TypeImpl.java
```diff
@@ -193,6 +175,7 @@ public final boolean isRemote() {
         public final void setLabel(String label) {
             execute(ConceptProto.Type.Req.newBuilder()
                     .setTypeSetLabelReq(ConceptProto.Type.SetLabel.Req.newBuilder().setLabel(label)));
+            this.label = label;
```
alexjpwalker, line 178: `setLabel` was updating the concept on the server, but not updating the local field, `label`. This change fixes that.

rpc/RPCTransaction.java
```diff
@@ -175,7 +177,6 @@ public void onNext(TransactionProto.Transaction.Res res) {
                 final ResponseCollector collector = collectors.get(requestId);
                 if (collector == null) throw new GraknClientException(UNKNOWN_REQUEST_ID.message(requestId));
                 collector.add(new Response.Ok(res));
-                if (collector.isDone()) collectors.remove(requestId);
```
alexjpwalker, line 178: During the flow of streaming calls, it's highly likely that the client will send a `CONTINUE` request that will be picked up by the server after it has already sent a `DONE` message. The server will respond to this `CONTINUE` with a second `DONE`.

When the client received the 1st `DONE` message, it would destroy the response collector. Then, on receiving the 2nd `DONE`, it would throw an exception, saying that the `requestId` is not matched to any response collector.

This change fixes that.

lolski: Makes sense

lolski (approved):

PR #196, opened 2020-12-08
Update BDD resource clean up strategy

lriuui0x0:

This PR is the counterpart of https://github.com/graknlabs/behaviour/pull/116.

In order to make sure all resources are released after each scenario run, so that during debugging it's easy to reason about resource usage.

We now create new connection before each scenario, and we delete all databases at the end of each scenario.

### DIFF
--- a/test/behaviour/connection/ConnectionSteps.java
+++ b/test/behaviour/connection/ConnectionSteps.java
@@ -26,6 +26,7 @@ import grakn.client.rpc.GraknClient;
 import grakn.common.test.server.GraknSingleton;
 import io.cucumber.java.After;
 import io.cucumber.java.Before;
+import io.cucumber.java.en.Given;
 
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


test/behaviour/connection/ConnectionSteps.java
```diff
@@ -51,56 +53,44 @@
...
 
+    @Before
+    public synchronized void before() {
+        assertNull(client);
+        System.out.println("Connecting to Grakn ...");
+        String address = GraknSingleton.getGraknRunner().address();
+        assertNotNull(address);
+        System.out.println("Establishing Connection to Grakn Core at " + address);
+        client = new GraknClient(address);
+        System.out.println("Connection to Grakn Core established");
```
alexjpwalker, line 79: Opening a new client for each scenario is fine, perhaps even desirable- but in that case let's take out the `System.out.println` statements. They'll feel like unnecessary noise when they're run in every single scenario.

alexjpwalker (approved):

PR #199, opened 2020-12-09
Update protocol to extract out LogicManager + Session pulse

lriuui0x0:

We have separated out the `LogicManager` to manage rules in the newest version of protocol, now we update client-java with this newest version of the protocol.

As part of the upgrade process, we need to also implement session pulse in order to keep the session alive. Otherwise the session will be killed by the server.

- Extract `logicManager` class that sits in `logic` package and manages rules.
- Implement a `TimerTask` for sending pulse request to the server.

### DIFF
--- a/rpc/RPCSession.java
+++ b/rpc/RPCSession.java
@@ -27,6 +27,8 @@ import grakn.protocol.GraknGrpc;
 import grakn.protocol.SessionProto;
 import io.grpc.Channel;
 
+import java.util.Timer;
+import java.util.TimerTask;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static grakn.client.common.ProtoBuilder.options;
@@ -38,6 +40,7 @@ public class RPCSession implements Session {
     private final Type type;
     private final ByteString sessionId;
     private final AtomicBoolean isOpen;
+    private final Timer pulse;
     private final GraknGrpc.GraknBlockingStub blockingGrpcStub;
 
     RPCSession(GraknClient client, String database, Type type, GraknOptions options) {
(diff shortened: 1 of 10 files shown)


rpc/RPCTransaction.java
```diff
@@ -110,6 +113,11 @@ public ConceptManager concepts() {
         return conceptManager;
     }
 
+    @Override
+    public LogicManager logics() {
```
alexjpwalker, line 117: We've agreed to rename this to `logic`.
```suggestion
    public LogicManager logic() {
```

alexjpwalker (approved):

rpc/RPCSession.java
```diff
@@ -97,4 +103,12 @@ public String database() {
                 return SessionProto.Session.Type.UNRECOGNIZED;
         }
     }
+
+    private class PulseTask extends TimerTask {
+        @Override
+        public void run() {
+            if (!isOpen()) return;
+            blockingGrpcStub.sessionPulse(SessionProto.Session.Pulse.Req.newBuilder().setSessionId(sessionId).build());
```
alexjpwalker, line 111: This won't work by itself - we need to use the `blockingGrpcStub` to actually transmit the request.

The response from the server will then include a Boolean flag named `isAlive` that indicates if the session is still alive on the server. If it is not, we should close it in the client.

alexjpwalker (changes requested):
Requested some changes

rpc/RPCSession.java
alexjpwalker: Never mind I read it wrong

alexjpwalker (approved):

PR #206, opened 2020-12-16
Fix a race condition in the timer

lolski:

Fix a `NullPointerException` issue where `isOpen` is asynchronously accessed by the timer job before it is initialised:

```
Exception in thread "Timer-0" java.lang.NullPointerException
	at grakn.client.rpc.RPCSession.isOpen(RPCSession.java:78)
	at grakn.client.rpc.RPCSession$PulseTask.run(RPCSession.java:110)
	at java.util.TimerThread.mainLoop(Timer.java:555)
	at java.util.TimerThread.run(Timer.java:505)
```

### DIFF
--- a/rpc/RPCSession.java
+++ b/rpc/RPCSession.java
@@ -54,8 +54,8 @@ public class RPCSession implements Session {
 
         sessionId = blockingGrpcStub.sessionOpen(openReq).getSessionId();
         pulse = new Timer();
-        pulse.scheduleAtFixedRate(this.new PulseTask(), 0, 5000);
         isOpen = new AtomicBoolean(true);
+        pulse.scheduleAtFixedRate(this.new PulseTask(), 0, 5000);
     }
 
     @Override


rpc/RPCSession.java
```diff
@@ -54,8 +54,8 @@
 
         sessionId = blockingGrpcStub.sessionOpen(openReq).getSessionId();
         pulse = new Timer();
-        pulse.scheduleAtFixedRate(this.new PulseTask(), 0, 5000);
         isOpen = new AtomicBoolean(true);
+        pulse.scheduleAtFixedRate(this.new PulseTask(), 0, 5000);
```
alexjpwalker, line 58: ah yes, another reminder that running new threads in the constructor is evil...

alexjpwalker (approved):

rpc/RPCSession.java
alexjpwalker: I think this is fine. It's a little bit fragile but the solutions to that seem very arbitrary (e.g: initialising isOpen inline; using `scheduleAtFixedRate(this.new PulseTask(), 100, 5000)` etc....) so I don't think we can really do any better.

lolski:
Merging the PR. The CI jobs have all passed but due to Grabl bug the notification wasn't sent.

```
# build
Executed 29 out of 29 tests: 29 tests pass.

# test-integration
Executed 2 out of 2 tests: 2 tests pass.

# test-behaviour
Executed 2 out of 2 tests: 2 tests pass.
```

PR #210, opened 2020-12-22
Adapt to new grabl

vmax:

Grabl recently changed the `automation.yml` syntax to have `image` instead of `machine`

* Replace all `machine:` usages with `image:`
* Only depend on `build` versions of other repos


lolski (approved):

lolski:
Merging despite the fact `test-behaviour` failed - the failure might have been introduced in a prior PR.

PR #211, opened 2020-12-22
Make getSupertypes() return all supertypes, including 'thing'

lolski:

Comply with the latest change where `thing` must be returned when invoking `getSupertypes()` on a subtype of thing such as `entity`, `attribute`, and `relation`.

1. `ConceptManager::getType` has been replaced with `ConceptManager::getThingType`
2. `getSupertypes()` on `AttributeType`, `EntityType` and `RelationType` must now include `thing`, we can no longer offer specialised implementation for each type and therefore they have been replaced with a generic implementation that sits at `ThingType`.


concept/type/Type.java
```diff
@@ -34,6 +34,8 @@
     @CheckReturnValue
     boolean isRoot();
 
+    boolean isRelationType();
```
haikalpribadi, line 37: Add the others too

concept/type/impl/TypeImpl.java
```diff
@@ -185,6 +191,11 @@ public final boolean isAbstract() {
                     .getTypeIsAbstractRes().getAbstract();
         }
 
+        @Override
+        public boolean isRelationType() {
+            return false;
+        }
```
haikalpribadi, line 197: We don't need this one. The root one is the correct one, and sufficient.

dependencies/maven/artifacts.snapshot
```diff
@@ -18,12 +18,18 @@
...
-@maven//:com_google_j2objc_j2objc_annotations_1_1
+@maven//:com_google_j2objc_j2objc_annotations_1_3
 @maven//:com_google_protobuf_protobuf_java
 @maven//:com_google_protobuf_protobuf_java_3_5_1
+@maven//:commons_codec_commons_codec
+@maven//:commons_codec_commons_codec_1_11
 @maven//:commons_io_commons_io
 @maven//:commons_io_commons_io_2_3
+@maven//:commons_logging_commons_logging
+@maven//:commons_logging_commons_logging_1_2
```
haikalpribadi, line 32: Let's check why we are bringing in 6 additional dependencies.

test/behaviour/connection/transaction/TransactionSteps.java
```diff
@@ -103,7 +103,9 @@ public void transaction_commits_throws_exception() {
     public void for_each_session_transactions_commit() {
         for (Session session : sessions) {
             for (Transaction transaction : sessionsToTransactions.get(session)) {
+                System.out.println("before commit");
                 transaction.commit();
+                System.out.println("after commit");
```
haikalpribadi, line 108: Let's remove the print outs.

test/behaviour/connection/transaction/BUILD
```diff
@@ -67,7 +67,7 @@ grakn_java_test(
         "@graknlabs_behaviour//connection:transaction.feature",
     ],
     native_grakn_artifact = "//test:native-grakn-artifact",
-    size = "small",
+    size = "medium",
```
haikalpribadi, line 70: Don't forget to revert

haikalpribadi (changes requested):

concept/ConceptManager.java
```diff
@@ -52,22 +52,22 @@ public ConceptManager(RPCTransaction rpcTransaction) {
 
     @CheckReturnValue
     public ThingType getRootThingType() {
-        return getType(GraqlToken.Type.THING.toString()).asThingType();
+        return getThingType(GraqlToken.Type.THING.toString()).asThingType();
```
alexjpwalker, line 55: can be simplified to 
```suggestion
        return getThingType(GraqlToken.Type.THING.toString());
```
There are more examples further down the file.

concept/ConceptManager.java
```diff
@@ -81,7 +81,7 @@ public EntityType putEntityType(String label) {
     @Nullable
     @CheckReturnValue
     public EntityType getEntityType(String label) {
-        final Type type = getType(label);
+        final Type type = getThingType(label);
```
alexjpwalker, line 84:
```suggestion
        final ThingType type = getThingType(label);
```
Although the difference is academic right now, we are planning to reintroduce `asThingType`, `asEntityType` etc. to `Type` class which will make stricter typing more relevant.

concept/type/impl/ThingTypeImpl.java
```diff
@@ -79,13 +80,14 @@ void setSupertype(ThingType thingType) {
 
         @Override
         public ThingTypeImpl getSupertype() {
-            final TypeImpl supertype = super.getSupertype();
+            TypeImpl supertype = super.getSupertype();
```
alexjpwalker, line 83: Why did we lose `final` here?
```suggestion
            final TypeImpl supertype = super.getSupertype();
```

dependencies/maven/artifacts.snapshot
alexjpwalker: Agreed. This doesn't look right at all- as far as I can tell, this PR doesn't add any code that uses any of them.

alexjpwalker (changes requested):

concept/type/impl/TypeImpl.java
alexjpwalker: This one **is** the root one, right?

lolski: @haikalpribadi  is referring to the fact that we have duplication between `TypeImpl::isRelationType` and `TypeImpl.Remote::isRelationType`.

We can remove the duplication by moving the implementation up to the `Type` interface as default methods. What do you think of that @alexjpwalker ?

alexjpwalker: `Type.Remote` does not extend `Type`. Therefore this one is the root one.

I am of the opinion that interface defaults are a hack in Java and should be generally avoided unless there is a good reason. Yes, we end up with some code duplication between `Type` and `Type.Remote` but that seems more like a limitation of Java when applied to our type hierarchy than anything else.

The reason is that interfaces are designed to be as dumb as possible. They should give a high-level overview of what the implementing classes are capable of, not actually tell them how to implement themselves.

dependencies/maven/artifacts.snapshot
lolski: They were brought in by the new implementation of `release_validate_deps`. Therefore, the dependency will not be included in the JAR. It's only used in build. See [this discussion](https://discord.com/channels/665254494820368395/672728623706275845/793452248284790814) for reference.

The fact that it's not clear which dependencies are only used for build / test and which one are actually included in the distribution will be addressed soon: https://github.com/graknlabs/dependencies/issues/244.

test/behaviour/connection/transaction/TransactionSteps.java
lolski: Fixed

test/behaviour/connection/transaction/BUILD
lolski: Fixed

concept/type/impl/TypeImpl.java
haikalpribadi: > I am of the opinion that interface defaults are a hack in Java and should be generally avoided unless there is a good reason

Yes, I agree, that they're not ideal. I try to avoid them too. But in this case, we have a good reason to use them: to avoid duplicating code multiple times, @alexjpwalker.

So as @lolski said:

> We can remove the duplication by moving the implementation up to the Type interface as default methods.

lolski: Fixed in 3a1f1aa182cf3506c89d3d2e43e21fa689834aec

concept/type/Type.java
lolski: Fixed in https://github.com/graknlabs/client-java/pull/211/commits/36f8dd8033f9cc19da804e18f78a5d347f40aa76

concept/type/impl/ThingTypeImpl.java
lolski: Good catch. I missed them because we never use `final` on local variables in other repositories. Fixed in https://github.com/graknlabs/client-java/pull/211/commits/63040ad062025b93a3389394149933c6e99bf1c4

concept/ConceptManager.java
lolski: Fixed in https://github.com/graknlabs/client-java/pull/211/commits/ace3adaff902dab2bca5d1d1961d29e164c4ddf8.

concept/ConceptManager.java
lolski: Also fixed in https://github.com/graknlabs/client-java/commit/ace3adaff902dab2bca5d1d1961d29e164c4ddf8

concept/ConceptManager.java
```diff
@@ -81,8 +81,8 @@ public EntityType putEntityType(String label) {
     @Nullable
     @CheckReturnValue
     public EntityType getEntityType(String label) {
-        final Type type = getType(label);
-        if (type instanceof EntityType) return type.asEntityType();
+        final ThingType thingType = getThingType(label);
+        if (thingType instanceof EntityType) return thingType.asEntityType();
```
alexjpwalker, line 85:
```suggestion
        if (thingType.isEntityType()) return thingType.asEntityType();
```
There are other places in this file - let's change them all!

concept/type/impl/AttributeTypeImpl.java
```diff
@@ -79,6 +81,11 @@ public final boolean isKeyable() {
         return new AttributeTypeImpl.Remote(transaction, getLabel(), isRoot());
     }
 
+    @Override
+    public boolean isAttributeType() {
```
alexjpwalker, line 85: Let's push this up into `AttributeType`.

There are other examples - apply the change to all of them!

alexjpwalker (changes requested):

concept/ConceptManager.java
lolski: Fixed in https://github.com/graknlabs/client-java/pull/211/commits/aa64cc93a7e0c3155a2b314f1dc9f400e88d41b8

concept/type/impl/AttributeTypeImpl.java
lolski: Fixed: https://github.com/graknlabs/client-java/pull/211/commits/128016426240102c1db53eff8c33755138a8b0ad

alexjpwalker (approved):

lolski:
Merging the PR based on @alexjpwalker's approval and @haikalpribadi 's verbal approval (despite the fact that he hasn't clicked on the "approve" button on the PR)

PR #212, opened 2020-12-29
Close all requests when the transaction stream is closed

lriuui0x0:

https://github.com/graknlabs/client-java/pull/208 and https://github.com/graknlabs/client-java/pull/209 introduced a bug such that when transaction is closed (either by client or server), the pending requests might be hanging forever. This is due to we didn't post an error message to the pending request's response buffer to make sure they will all be unblocked and errored.

- Post `Done` response when the transaction is closed, either on completed or on error.
- Refactor the response model so that we split the scenario of stream shutdown due to normal condition, or due to error condition.

### DIFF
--- a/rpc/RPCTransaction.java
+++ b/rpc/RPCTransaction.java
@@ -224,7 +224,7 @@ public class RPCTransaction implements Transaction {
 
         void add(Response response) {
             responseBuffer.add(response);
-            if (!(response instanceof Response.Result) || isLastResponse(read(response))) isDone.set(true);
+            if (!(response instanceof Response.Result) || isLastResponse(response.read())) isDone.set(true);
         }
 
         boolean isDone() {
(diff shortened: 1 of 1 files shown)


rpc/RPCTransaction.java
```diff
@@ -170,38 +174,23 @@ public void onNext(TransactionProto.Transaction.Res res) {
...
-                closeSuccess();
-            }
-
-            private void closeError(StatusRuntimeException ex) {
-                if (isOpen.compareAndSet(true, false)) {
-                    collectors.error(new Response.Error(ex));
-                    collectors.clear();
-                    requestObserver.onCompleted();
-                }
+                close(new Response.Done.Completed());
```
lriuui0x0, line 189: Here is the main bug fix. Previously we use `closeSuccess()` which doesn't post an error response to all pending requests, which means they may still be blocked waiting for server response forever. Now we make sure we always post the error response (implemented in `close` method) so that all pending requests will be unblocked.

rpc/RPCTransaction.java
```diff
@@ -238,23 +224,36 @@ private synchronized void clear() {
...
         }
 
         TransactionProto.Transaction.Res take(long timeout, TimeUnit unit) throws InterruptedException, TimeoutException {
             final Response response = responseBuffer.poll(timeout, unit);
-            if (response != null) return response.read();
+            if (response != null) return read(response);
             else throw new TimeoutException();
         }
 
+        TransactionProto.Transaction.Res read(Response response) {
```
alexjpwalker, line 244: Why did we take `read` out of `Response`? This method shows the very reason why I put it _in_ `Response` in the first place- effectively a big switch statement with several `instanceof`s. It would be cleaner to take advantage of inheritance, and make it a method inside `Response` that is overloaded in `Result`, `Done.Completed` and `Done.Error`.

lriuui0x0: @lolski suggested that change. I think the reasoning to make the code reads better in the top level `close` method. It would be counter-intuitive if we do `close(new Response.Done.Error(...))` because it's not clear why a normal `close` operation would cause error. Instead we do `close(new Response.Done.Completed(...))` which is more expected.

lriuui0x0: That said, it seems that we should still be able to use inheritance. Let me check.

alexjpwalker (approved):

haikalpribadi (approved):

PR #213, opened 2020-12-29
Remove "Received DATA frame for an unknown stream" warning

lriuui0x0:

We currently see a lot of warning from the log that says

```
Received DATA frame for an unknown stream <id>
...
```

This seems to be caused client trying to send requests when server has already shutdown the grpc connection. We try to fix this issue by making sure client doesn't send an extra `requestObserver.onCompleted();` when server shutsdown the connection.

- Only call `requestObserver.onCompleted();` when client actively closes the connection.

### DIFF
--- a/rpc/RPCTransaction.java
+++ b/rpc/RPCTransaction.java
@@ -127,10 +127,16 @@ public class RPCTransaction implements Transaction {
 
     @Override
     public void close() {
-        close(new Response.Done.Completed());
+        closeFromClient(new Response.Done.Completed());
     }
 
-    private void close(Response.Done doneResponse) {
+    private void closeFromServer(Response.Done doneResponse) {
+        if (isOpen.compareAndSet(true, false)) {
+            collectors.clear(doneResponse);
+        }
+    }
+
+    private void closeFromClient(Response.Done doneResponse) {
         if (isOpen.compareAndSet(true, false)) {
             collectors.clear(doneResponse);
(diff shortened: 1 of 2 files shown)


rpc/RPCTransaction.java
```diff
@@ -127,10 +127,16 @@ public void rollback() {
 
     @Override
     public void close() {
-        close(new Response.Done.Completed());
+        closeFromClient(new Response.Done.Completed());
     }
 
-    private void close(Response.Done doneResponse) {
+    private void closeFromServer(Response.Done doneResponse) {
```
alexjpwalker, line 133: Hmm. I think a better name for this method would be `onServerClose` - it's something that gets triggered remotely, not by the client. In which case we can take out `closeFromClient` and simply move its code into `close`.

alexjpwalker:
I now see why this change was necessary. `isOpen` used to be split into 3 AtomicBooleans - `streamIsOpen`, `transactionWasOpened`, `transactionWasClosed`. It would check `streamIsOpen` before sending a Completed message, and check `transactionWasOpened` before sending errors to all response collectors.

If it can be made to work with just one `AtomicBoolean` then that's definitely cleaner - but we should test extensively.

lriuui0x0:
Close this PR as we have made contribution to the grpc-java repository to change the logging level https://github.com/grpc/grpc-java/pull/7771

PR #214, opened 2020-12-31
Make flaky tests use debug mode grakn runner

lriuui0x0:

In order to have more information when flaky test fails, we enable the debug mode grakn server and print out output for bazel test.

- Use debug mode grakn for flaky BDD tests.
- Set test_output=streamed for CI job.


alexjpwalker (approved):
Looks good to me

PR #216, opened 2021-01-04
API cleanup: close on commit, isX/asX methods

alexjpwalker:

As we close transactions on commit at the server end, it makes sense to also close transactions when they are committed at client side. This change does that. Additionally, for user experience reasons, "isX" functions have been added to the Concept interface to determine what variety of Concept an object represents.

We added a try/finally to `commit()` that closes the transaction when it completes or fails.
We added is[Concept] functions to all Concepts indicating if the current Concept is an instance of that Concept variety.

### DIFF
--- a/concept/Concept.java
+++ b/concept/Concept.java
@@ -21,8 +21,16 @@ package grakn.client.concept;
 
 import grakn.client.Grakn;
 import grakn.client.common.exception.GraknClientException;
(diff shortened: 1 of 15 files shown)


alexjpwalker (commented):
These changes look good

Phillammon (approved):
Confirmed against client-node, all good.

PR #218, opened 2021-01-04
RPC implementation for sending match aggregate, match group, and match group aggregate queries

lolski:

We have implemented RPC support for sending match aggregate, match group, and match group aggregate queries to server

1. Update `@graknlabs_protocol` to the version that supports the new queries.
2. Update `RPCTransaction` and `QueryManager` to support processing the new queries


concept/answer/AnswerGroup.java
```diff
@@ -39,13 +33,6 @@ public AnswerGroup(Concept owner, List<T> answers) {
         this.answers = answers;
     }
 
-    public static AnswerGroup<? extends Answer> of(Transaction tx, AnswerProto.AnswerGroup res) {
```
alexjpwalker, line 42: Why is this deleted? We need it to deserialize `ConceptMapGroup` received from the server.

concept/answer/AnswerGroup.java
```diff
@@ -19,17 +19,11 @@
...
-import grakn.client.concept.thing.impl.ThingImpl;
-import grakn.client.concept.type.impl.TypeImpl;
-import grakn.protocol.AnswerProto;
 
 import java.util.List;
 
-import static java.util.stream.Collectors.toList;
-
-public class AnswerGroup<T> implements Answer {
+public class AnswerGroup<T> {
```
alexjpwalker, line 26: Should this not, now, be renamed to `ConceptMapGroup`, and a new class created named `NumberGroup`?

It's possible that they could inherit or implement `AnswerGroup<ConceptMap>` and `AnswerGroup<Number>`.

query/QueryManager.java
```diff
@@ -100,4 +160,11 @@ public QueryManager(RPCTransaction rpcTransaction) {
                 .setQueryReq(request.setOptions(options(options)));
         return rpcTransaction.stream(req, responseReader);
     }
+
+    private Concept concept(ConceptProto.Concept owner) {
```
alexjpwalker, line 164: For consistency with other `Concept` deserializers, this method should be `static`, named `of`, and live in `Concept.java`.

query/QueryManager.java
```diff
@@ -50,17 +58,63 @@ public QueryManager(RPCTransaction rpcTransaction) {
 
     public Stream<ConceptMap> match(GraqlMatch query, GraknOptions options) {
         final QueryProto.Query.Req.Builder request = QueryProto.Query.Req.newBuilder().setMatchReq(
-                QueryProto.Graql.Match.Req.newBuilder().setQuery(query.toString()));
+                QueryProto.Query.Match.Req.newBuilder().setQuery(query.toString()));
         return iterateQuery(request, options, res -> res.getQueryRes().getMatchRes().getAnswersList().stream().map(ConceptMap::of));
     }
 
+    public QueryFuture<String> match(GraqlMatch.Aggregate query) {
```
alexjpwalker, line 65: I know we serialize the number as a `String`, but that feels like an internal implementation detail. I don't think we should be returning it to the user as a `String`.

What data type does it have in Grakn Core? `double`?

query/QueryManager.java
```diff
@@ -50,17 +58,63 @@ public QueryManager(RPCTransaction rpcTransaction) {
...
+        return match(query, new GraknOptions());
+    }
+
+    public Stream<AnswerGroup<ConceptMap>> match(GraqlMatch.Group query, GraknOptions options) {
+        final QueryProto.Query.Req.Builder request = QueryProto.Query.Req.newBuilder().setMatchGroupReq(
+                QueryProto.Query.MatchGroup.Req.newBuilder().setQuery(query.toString()));
+        return iterateQuery(
+                request, options,
+                res -> res.getQueryRes().getMatchGroupRes().getAnswersList().stream()
+                        .map(e -> {
```
alexjpwalker, line 85: For consistency with other `answer` deserializers this function (the lambda inside `.map`) should live inside `AnswerGroup` and be named `of`

query/QueryManager.java
```diff
@@ -50,17 +58,63 @@ public QueryManager(RPCTransaction rpcTransaction) {
...
+        return match(query, new GraknOptions());
+    }
+
+    public Stream<AnswerGroup<String>> match(GraqlMatch.Group.Aggregate query, GraknOptions options) {
+        final QueryProto.Query.Req.Builder request = QueryProto.Query.Req.newBuilder().setMatchGroupAggregateReq(
+                QueryProto.Query.MatchGroupAggregate.Req.newBuilder().setQuery(query.toString()));
+        return iterateQuery(
+                request, options,
+                res -> res.getQueryRes().getMatchGroupAggregateRes().getAnswersList().stream()
+                        .map(e -> {
```
alexjpwalker, line 103: See above comment

alexjpwalker (changes requested):

concept/answer/AnswerGroup.java
alexjpwalker: (PS I just noticed it is implemented inside `QueryManager` now - we should move it back here for consistency with other `answer` and `Concept` deserializers)

concept/answer/AnswerGroup.java
lolski: I think in Java, we can be generic and therefore we should. We flatten the structure in protobuf because it could not be implemented generically there. I am keeping this implementation because this approach is consistent with what we have in Grakn Core: https://github.com/graknlabs/grakn/blob/master/query/QueryManager.java#L182-L204

What do you think of that?

query/QueryManager.java
lolski: Yeah, I meant to ask you about how to return it to the user. Taking a look at how we're doing the conversion in Grakn Core makes sense.

concept/answer/AnswerGroup.java
alexjpwalker: The thing is that what we have in Grakn Core, most likely, existed before our meeting yesterday. I'm thinking we would update both Core and the clients.

query/QueryManager.java
lolski: Should we put the method in the `grakn.client.concept.Concept` interface? or should we create another `Concept` class in `grakn.client.concept.answer.Concept`?

I am inclined for the former, but it is an `interface` and adding a conversion method that knows about the implementation detail there .

concept/answer/AnswerGroup.java
lolski: Okay, I like that. Because we don't need something as generic as `AnswerGroup<T>` anyway.

query/QueryManager.java
alexjpwalker: We have a `ConceptImpl` class now! Pull from `master` if you don't have in your codebase yet.

lolski: cool!

query/QueryManager.java
alexjpwalker: In Grakn Core it currently returns `Number`, which is either a `long` or a `double`. This is probably the best option as `Number` has the `longValue` and `doubleValue` methods.

query/QueryManager.java
lolski: Addressed in https://github.com/graknlabs/client-java/pull/218/commits/a1a66fd1d280199b60bed6f2ecb2e9c0ee4538d1

concept/answer/AnswerGroup.java
lolski: Addressed in https://github.com/graknlabs/client-java/pull/218/commits/16316dbec63658cb13c503319f3cf662ac6199d8

query/QueryManager.java
lolski: Addressed in https://github.com/graknlabs/client-java/pull/218/commits/16316dbec63658cb13c503319f3cf662ac6199d8

query/QueryManager.java
lolski: Addressed in https://github.com/graknlabs/client-java/pull/218/commits/16316dbec63658cb13c503319f3cf662ac6199d8

query/QueryManager.java
lolski: Addressed in https://github.com/graknlabs/client-java/pull/218/commits/7e6ebb80ca6044a40aba133db2f0a479c7b511c1

concept/answer/Numeric.java
```diff
@@ -0,0 +1,71 @@
+package grakn.client.concept.answer;
+
+import grakn.client.common.exception.GraknClientException;
+import grakn.protocol.AnswerProto;
+
+import javax.annotation.Nullable;
+
+public class Numeric {
```
alexjpwalker, line 8: I think it would be more consistent with the rest of our codebase if we split up `Numeric` into 3 inner classes: `Numeric.Long`, `Numeric.Double` and `Numeric.NaN`.

alexjpwalker (changes requested):

concept/answer/Numeric.java
lolski: And then have `asLong()` and `asDouble()` return `Numeric.Long` and `Numeric.Double` rather than `long` and `double`?

I can see how that's how we're doing it for attribute type, for example, so I can see the consistency. However, it'll make the code more verbose.

What do you guys think @haikalpribadi @alexjpwalker ?

alexjpwalker: > And then have `asLong()` and `asDouble()` return `Numeric.Long` and `Numeric.Double` rather than `long` and `double`?

Yes and I agree with your thoughts too - let's see what Haikal thinks

haikalpribadi: I think it would be too verbose, @alexjpwalker @lolski . I think the user should get the `primitive types` when they call `.longValue()` and `.doubleValue()`.

haikalpribadi: and `NaN` does not even need a class anyways, right.

haikalpribadi: ps: `NaN` goes well with curry :D ugh I'm still hungry

lolski: haha mark as spam

alexjpwalker (approved):

PR #222, opened 2021-01-06
Fix invalid cast for the super type of root types

lriuui0x0:

Currently the super type of entity / relation / attribute / role root type is casted to entity / relation / attribute / role type. However the actual type is thing type. Therefore we cannot cast it to the specific type but keep it as thing type.

This is consistent with server implementation.

- Remove overrides of `getSuperType()` from entity / relation / attribute / role type.

### DIFF
--- a/concept/type/AttributeType.java
+++ b/concept/type/AttributeType.java
@@ -143,9 +143,6 @@ public interface AttributeType extends ThingType {
         void setSupertype(AttributeType attributeType);
 
         @Override
-        AttributeType getSupertype();
-
-        @Override
         Stream<? extends AttributeType> getSubtypes();
 
         @Override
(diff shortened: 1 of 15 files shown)


concept/type/RoleType.java
```diff
@@ -39,12 +39,6 @@ default boolean isRoleType() {
 
     interface Remote extends Type.Remote, RoleType {
 
-        @Override
-        RoleType getSupertype();
-
-        @Override
-        Stream<? extends RoleType> getSupertypes();
```
alexjpwalker, line 46: Pretty sure we should keep these. These are still guaranteed to be `RoleType` and `Stream<? extends RoleType>`. The root RoleType is `relation:role` which is a `RoleType` and its supertype is `null` - it has no supertypes.

alexjpwalker (changes requested):

alexjpwalker (approved):

haikalpribadi (approved):

PR #226, opened 2021-01-08
Bump year in license headers to 2021

vmax:

License headers are currently out of date (2020) for a while. This PR replaces 2020 with 2021 across all the files.

* Replace `2020` with `2021` in source files

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2020 Grakn Labs.
+   Copyright 2021 Grakn Labs.
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


lolski (approved):

LICENSE
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2020 Grakn Labs.
+   Copyright 2021 Grakn Labs.
```
lolski, line 190: Is this expected @haikalpribadi ? We've got an Apache `LICENSE` file, and we're also embedding a copyright statement in it.

haikalpribadi: I don't know @lolski - this looks weird. Do we have it on other Apache license? I don't think so, right? If so we should remove it.

PR #227, opened 2021-01-08
Fix broken Connection and Concept BDD scenarios

alexjpwalker:
## What is the goal of this PR?

Doing a fresh pass over our BDD scenarios exposed a number of assorted issues, which we have now fixed.

## What are the changes implemented in this PR?

- If session fails to close, throw a `GraknClientException`
- Clean databases before running BDD, clean `sessionsParallel` after
- Fix queries not throwing when they should in BDD, by correctly awaiting them
- Make BDD steps always refer to overridden roles by their scoped label

PR #229, opened 2021-01-13
Throw correct exception type while constructing session and transaction

lriuui0x0:

We have fixed a bug which caused incorrect exception type being thrown while constructing session and transaction.

- Catch `StatusRuntimeException` and convert it to `GraknClientException` in `Session` and `Transaction` constructor.

### DIFF
--- a/rpc/RPCSession.java
+++ b/rpc/RPCSession.java
@@ -24,9 +24,11 @@ import grakn.client.Grakn.Session;
 import grakn.client.Grakn.Transaction;
 import grakn.client.GraknClient;
 import grakn.client.GraknOptions;
+import grakn.client.common.exception.GraknClientException;
 import grakn.protocol.GraknGrpc;
 import grakn.protocol.SessionProto;
 import io.grpc.Channel;
+import io.grpc.StatusRuntimeException;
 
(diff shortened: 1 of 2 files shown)


rpc/RPCSession.java
```diff
@@ -45,18 +47,22 @@
...
+            final SessionProto.Session.Open.Req openReq = SessionProto.Session.Open.Req.newBuilder()
+                    .setDatabase(database).setType(sessionType(type)).setOptions(options(options)).build();
+
+            sessionId = blockingGrpcStub.sessionOpen(openReq).getSessionId();
+            pulse = new Timer();
+            isOpen = new AtomicBoolean(true);
+            pulse.scheduleAtFixedRate(this.new PulseTask(), 0, 5000);
+        } catch (StatusRuntimeException e) {
+            throw new GraknClientException(e);
+        }
```
alexjpwalker, line 65: It's worth noting that we should do the same change in `close` - however, I've already done so in this PR: https://github.com/graknlabs/client-java/pull/227; so don't worry about it.

alexjpwalker (approved):

lolski (approved):

PR #230, opened 2021-01-14
Upgrade GraqlSteps (BDD) to Grakn 2.0

alexjpwalker:

Our implementation of BDD, `GraqlSteps`, was out of date, with much of the code being Grakn 1.8 code. We upgraded it to 2.0.

Upgrade GraqlSteps (BDD) to Grakn 2.0


test/behaviour/graql/GraqlSteps.java
```diff
@@ -22,39 +22,47 @@
...
 import static grakn.client.test.behaviour.util.Util.assertThrowsWithMessage;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 public class GraqlSteps {
 
-    private static Stream<ConceptMap> answers;
-    private static Number numericAnswer;
+    private static List<ConceptMap> answers;
```
alexjpwalker, line 62: `Stream<ConceptMap>` was used for efficiency: allowing us to asynchronously run `match` while executing other steps. However, if we wanted to execute two follow-up steps (e.g: identify the answers in two different ways), that would require re-executing the query.

To mitigate this issue, we now collect the entire `Stream` to a `List` in `get answers of graql match`.

test/behaviour/graql/GraqlSteps.java
```diff
@@ -128,238 +136,226 @@ public void graql_delete_throws_exception(String exception, String deleteQuerySt
...
     public void answer_quantity_assertion(int expectedAnswers) {
-        // TODO
-//        assertEquals(
-//                String.format("Expected [%d] answers, but got [%d]", expectedAnswers, answers.count()),
-//                expectedAnswers, answers.count());
+        assertEquals(String.format("Expected [%d] answers, but got [%d]", expectedAnswers, answers.size()),
+                     expectedAnswers, answers.size());
     }
 
-    @Then("concept identifiers are")
```
alexjpwalker, line 180: This step is no longer used. Answer concept matching was simplified to just `uniquely identify answer concepts` a while back.

test/behaviour/graql/GraqlSteps.java
```diff
@@ -128,238 +136,226 @@ public void graql_delete_throws_exception(String exception, String deleteQuerySt
...
-    public void concept_identifiers_are(Map<String, Map<String, String>> identifiers) {
-        for (Map.Entry<String, Map<String, String>> entry : identifiers.entrySet()) {
-            String identifier = entry.getKey();
-            String check = entry.getValue().get("check");
-            String value = entry.getValue().get("value");
+    @Then("uniquely identify answer concepts")
+    public void uniquely_identify_answer_concepts(List<Map<String, String>> answerConcepts) {
```
alexjpwalker, line 193: This implementation, and all remaining implementations further down this file, are copied from Grakn Core.

test/behaviour/graql/GraqlSteps.java
```diff
@@ -369,24 +365,44 @@ public AnswerIdentifierGroup(List<Map<String, String>> answerIdentifierTable, Ma
         }
     }
 
-    private boolean matchAnswer(Map<String, String> answerIdentifiers, ConceptMap answer) {
-
-        if (!(answerIdentifiers).keySet().equals(answer.map().keySet())) {
-            return false;
-        }
+//    private boolean matchAnswer(Map<String, String> answerIdentifiers, ConceptMap answer) {
```
flyingsilverfin, line 368: what is this comment?

test/behaviour/graql/GraqlSteps.java
```diff
@@ -565,9 +583,38 @@ public boolean check(Concept concept) {
         }
 
         public boolean check(Concept concept) {
-            return concept.isAttribute()
-                    && type.equals(concept.asAttribute().asRemote(tx()).getType().getLabel())
-                    && value.equals(concept.asAttribute().getValue().toString());
+            if (!concept.isAttribute()) {
```
flyingsilverfin, line 586: style note: when we can fit into the char limit on one line, we tend to do so and get rid of the {}. However, i there's any block in a series of `if... else if... else` that has to have a multiline {}, all of them use multiline {} to be consistent

flyingsilverfin (approved):

test/behaviour/graql/GraqlSteps.java
alexjpwalker: It's a method used only by Explanation steps, so for consistency with them (& to fix compiler warning), I've commented it out

flyingsilverfin: hmm i can't see the context but is it near the steps that use the method? unclear right now its used for explanations. Might not be clear in core either tbh

alexjpwalker: Added a comment to indicate this

PR #236, opened 2021-01-20
Make the exception message when unable to connect to the server more descriptive

vmax:

Exception message when client cannot connect to server was not descriptive enough ("io exception"). It's changed to a more user-friendly message ("Unable to connect to Grakn Core Server").

Handle the special case of gRPC error - `UNAVAILABLE` (code 14) by supplying helpful `ErrorMessage`. Additionally, store and expose that `ErrorMessage` in `GraknClientException` such that user can distinguish the errors.

### DIFF
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -43,6 +43,8 @@ public abstract class ErrorMessage extends grakn.common.exception.ErrorMessage {
 
         public static final Client CLUSTER_NOT_AVAILABLE =
                 new Client(8, "Attempted connecting to these servers, but none are available: '%s'.");
+        public static final Client SERVER_UNAVAILABLE =
+                new Client(9, "Server is unavailable.");
 
         private static final String codePrefix = "CLI";
         private static final String messagePrefix = "Illegal Client State";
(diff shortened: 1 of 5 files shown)


common/exception/ErrorMessage.java
```diff
@@ -43,6 +43,8 @@ private ErrorMessage(String codePrefix, int codeNumber, String messagePrefix, St
 
         public static final Client CLUSTER_NOT_AVAILABLE =
                 new Client(8, "Attempted connecting to these servers, but none are available: '%s'.");
+        public static final Client SERVER_UNAVAILABLE =
+                new Client(9, "Server is unavailable.");
```
lolski, line 47: Let's pass down the host / port so that the error is descriptive:

```
Unable to connect to Grakn Core Server '%s' 
```

What do you think @alexjpwalker ?

This would be especially useful for client java users

alexjpwalker, line 47: I think "Failed to connect to server." might be more understandable for the user.

alexjpwalker: Actually I prefer Ganesh's version.

vmax: @alexjpwalker addressed in https://github.com/graknlabs/client-java/pull/236/commits/a25f5dfdc396650c909e336a29d6ce1cd49cc018

vmax: @lolski addressed in https://github.com/graknlabs/client-java/pull/236/commits/a25f5dfdc396650c909e336a29d6ce1cd49cc018

alexjpwalker (approved):

lolski (approved):

PR #238, opened 2021-01-21
Add ability to retrieve all rules

alexjpwalker:

Since Rule was refactored in Grakn 2.0 to no longer be a subtype of Concept, we lost the ability to retrieve all rules using `match $x sub rule; get;`.

So, to allow the user to retrieve all rules, we now add `getRules` to the `LogicManager`, enabling use of `tx().logic().getRules()`.

Add getRules to LogicManager

### DIFF
--- a/logic/impl/RuleImpl.java
+++ b/logic/impl/RuleImpl.java
@@ -52,7 +52,7 @@ public class RuleImpl implements Rule {
     }
 
     public static RuleImpl of(LogicProto.Rule ruleProto) {
-        return new RuleImpl(ruleProto.getLabel(), Graql.and(Graql.parsePatterns(ruleProto.getWhen())), Graql.parseVariable(ruleProto.getThen()).asThing());
+        return new RuleImpl(ruleProto.getLabel(), Graql.parsePattern(ruleProto.getWhen()).asConjunction(), Graql.parseVariable(ruleProto.getThen()).asThing());
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


logic/impl/RuleImpl.java
```diff
@@ -52,7 +52,7 @@
     }
 
     public static RuleImpl of(LogicProto.Rule ruleProto) {
-        return new RuleImpl(ruleProto.getLabel(), Graql.and(Graql.parsePatterns(ruleProto.getWhen())), Graql.parseVariable(ruleProto.getThen()).asThing());
+        return new RuleImpl(ruleProto.getLabel(), Graql.parsePattern(ruleProto.getWhen()).asConjunction(), Graql.parseVariable(ruleProto.getThen()).asThing());
```
alexjpwalker, line 55: This change fixes a parsing issue when reading a Rule retrieved remotely.

test/integration/ClientQueryTest.java
```diff
@@ -53,6 +56,7 @@ public static void setUpClass() throws InterruptedException, IOException, Timeou
         grakn = new GraknCoreRunner();
         grakn.start();
         graknClient = GraknClient.core(grakn.address());
+        if (graknClient.databases().contains("grakn")) graknClient.databases().delete("grakn");
```
alexjpwalker, line 59: This change allows this test to be run and succeed against a local instance of Grakn.

flyingsilverfin: nice!

flyingsilverfin (approved):

Issue #180, opened 2020-11-06
When a query fails, send our own Error response rather than a gRPC StatusRuntimeException


alexjpwalker:
## Problem to Solve

Now that we support concurrent operations within the same Transaction, error handling has become awkward on the client side.

In Grakn, any error in a transaction terminates the transaction. This means that pending requests will not be fulfilled.

Suppose there are 3 concurrent requests, but the 2nd request fails because (e.g.) the user attempted to do a Graql Insert in a schema session. Then ideally we would be able to fail the 2nd request with an error like "Attempted data write while session type does not allow it", and fail the 1st and 3rd requests with an error like "The request failed because an error was thrown executing another operation in this transaction."

But we can't, because the gRPC error doesn't mention anywhere which of the 3 requests was the one that triggered the error.

## Current Workaround

So instead, we simply send the error message to all requests, which is misleading for requests 1 and 3 since they had nothing to do with attempting data writes.

## Proposed Solution

We should add our own error response to the protocol containing a request ID. Then, the client will be able to differentiate requests that errored themselves, and requests that have been terminated because of errors in other concurrent operations.

(We could also just prepend the request ID to the error message and parse on the client, but that would be a hack)

PR #243, opened 2021-01-24
Improve failover mechanism for primary and secondary replica connections

lolski:

We have increased resilience by improving the failover mechanism between replicas.

- Remove the terminology leader / non-leader which are Raft specific. We now use the terminology "replica" to refer to a single copy of a database. The active replica that can receive data is now called "primary replica" whereas the passive ones "secondary replica"
- Split `GraknOptions` to `GraknOptions.core()` and `GraknOptions.cluster()`, the later of which contains an option to read from secondary replica
- Increase resilience:
  - Database and cluster discovery will now be re-attempted to all cluster members instead of just to one of them
  - When the cluster have not decided which replica is the primary replica, the client will wait and retry instead of simply failing
- Changed info-level log to debug


Grakn.java
```diff
@@ -114,8 +114,8 @@ public int id() {
 
         enum Type {
             READ(0),
-            READ_REPLICA(2),
-            WRITE(1);
+            WRITE(1),
+            READ_SECONDARY(2);
```
lolski, line 118: The reasoning to rename it to `READ_SECONDARY` is because it allows you to read from *secondary* replicas, whereas `READ` and `WRITE` reads to the *primary* replica.

dependencies/graknlabs/repositories.bzl
```diff
@@ -50,8 +50,8 @@ def graknlabs_dependencies():
 def graknlabs_protocol():
     git_repository(
         name = "graknlabs_protocol",
-        remote = "https://github.com/graknlabs/protocol",
-        tag = "2.0.0-alpha-6", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
+        remote = "https://github.com/lolski/protocol",
```
lolski, line 53: Revert to `graknlabs` when https://github.com/graknlabs/protocol/pull/110 is merged.

common/exception/GraknClientException.java
```diff
@@ -33,15 +34,17 @@ public GraknClientException(String error) {
...
+    public GraknClientException(ErrorMessage error, Object... parameters) {
+        super(error.message(parameters));
         assert !getMessage().contains("%s");
         this.errorMessage = error;
     }
 
     public static GraknClientException of(StatusRuntimeException statusRuntimeException) {
         if (statusRuntimeException.getStatus().getCode() == Status.Code.UNAVAILABLE) {
             return new GraknClientException(ErrorMessage.Client.UNABLE_TO_CONNECT);
+        } else if (statusRuntimeException.getStatus().getCode() == Status.Code.INTERNAL && statusRuntimeException.getStatus().getDescription().contains("[RFT01]")) {
```
lolski, line 46: In this particular if block I want to check if the server throws an exception because there's no leader yet.

The server will throw an exception of code "[RFT01]", and the only way to propagate that information about the exception is by embedding it in the message, hence the need for `getDescription().contains(...)`.

Is there a better way to propagate exceptions than this?

Do you have better ideas?

alexjpwalker: The best alternatives that we have _right now_ would be to either:

- change the protocol and server: don't throw an exception if there's no leader, but rather encode the error in protobuf; or
- add a new method to protocol and server to check if the leader exists - but this is both inefficient and non-atomic.

So in conclusion parsing the error message is the least bad option right now.

There are plans, in the future, to implement a dedicated `Error` message in protocol, which we will be able to interpret without the need for hacks: see https://github.com/graknlabs/client-java/issues/180 .

alexjpwalker (approved):

Grakn.java
alexjpwalker: We agreed in a verbal discussion with Haikal to get rid of the `READ_SECONDARY` transaction type and replace it with a new `Option`.

alexjpwalker (changes requested):
Updating review to reflect recent chat

common/exception/GraknClientException.java
lolski: I've moved the hacky code into its own method and added a TODO: https://github.com/graknlabs/client-java/pull/243/commits/268c884c5fae149d70c0136e643ba117e88c39f9

Grakn.java
lolski: Fixed in https://github.com/graknlabs/client-java/pull/243/commits/fec34cc75e0a53bb8180851d162e01023ce73251, https://github.com/graknlabs/client-java/pull/243/commits/62c43681b78726a34566307317bcae39663475db, https://github.com/graknlabs/client-java/pull/243/commits/8f31ae68c44d647e678f8f364bd283958d931e3b, and https://github.com/graknlabs/client-java/pull/243/commits/b8d3aa5d2ccb64f79e1bed18a9bc0940f05f4eec (I don't know why it took me 4 commits to do so :D )

rpc/RPCSession.java
```diff
@@ -309,17 +374,17 @@ public boolean equals(Object o) {
...
             @Override
             public int hashCode() {
-                return Objects.hash(isLeader, term);
+                return Objects.hash(isPrimary, term);
             }
 
             @Override
             public String toString() {
-                return id + ":" + (isLeader ? "L" : "N") + ":" + term;
+                return id + ":" + (isPrimary ? "P" : "S") + ":" + term;
```
alexjpwalker, line 387: Not 100% sure if this is the right thing to do in the context; do you think we should expand these terms and use `Primary` and `Secondary` instead of `P` and `S`? I don't know how human-readable you want these to be.

rpc/RPCSession.java
```diff
@@ -214,37 +211,105 @@ public String database() {
...
-            throw new GraknClientException(CLUSTER_NOT_AVAILABLE.message(clusterClient.coreClients()));
+        private void sleepWait() {
+            try {
+                Thread.sleep(2000);
+            } catch (InterruptedException e2) {
+                throw new GraknClientException(UNEXPECTED_INTERRUPTION);
+            }
         }
 
         public static class Database {
```
alexjpwalker, line 308: Is this class distinct from the concept of a _database_ (i.e. a knowledge graph?) If so, do you think we can disambiguate?

rpc/RPCSession.java
```diff
@@ -153,44 +155,39 @@ public void run() {
...
-            RPCSession.Core selection = coreSessions.computeIfAbsent(
-                    selected.id(),
-                    key -> {
-                        LOG.info("Opening a session to leader '{}'", selected);
-                        return clusterClient.coreClients().get(key.address()).session(key.database(), this.type, this.options);
-                    }
-            );
-            LOG.info("Opening a transaction of of type '{}' to leader '{}'", type, selected);
-            return selection.transaction(type, options);
+            if (!options.isCluster()) throw new GraknClientException(ILLEGAL_CAST, options);
```
alexjpwalker, line 184: I believe this line is redundant and should be deleted. `GraknOptions.asCluster` already throws if you're performing an illegal cast.
```suggestion
```

alexjpwalker (changes requested):

rpc/RPCSession.java
lolski: Yes, that's intentional. I intend to introduce a symbol that can describe a replica and its state in a concise way such that it can be easily included in the log without cluttering the whole line

```
[INFO] g.c.r.RPCSession - Replica '172.0.01:1729/grakn:P:5' is not responding after three attempts of reconnecting.
```

which would indicate that the replica of `grakn` which sits in the machine `172.0.01:1729`, is a primary replica of term 5 is not responding after three attempts.

They need to be short since we may need to print multiple of them in the same line given that we're working with multiple replicas most of the time.

I think exposing this information is highly useful not just for us but also for the users (eg., the app developers).

I think making them as short as possible (at the cost of being terse) is justified. What do you think @alexjpwalker ?

rpc/RPCSession.java
```diff
@@ -214,37 +211,105 @@ public String database() {
...
+                        LOG.debug("Opening a transaction to primary replica '{}'", database.primaryReplica().id());
+                        return primaryReplicaSession.transaction(type, options);
+                    } catch (GraknClientException e) {
+                        retry++;
+                        if (e.getErrorMessage().equals(CLUSTER_REPLICA_NOT_PRIMARY)) {
+                            LOG.debug("Unable to open a session or transaction", e);
+                            database = databaseDiscover(server);
+                        } else if (e.getErrorMessage().equals(CLUSTER_NO_PRIMARY_REPLICA_YET)) {
+                            LOG.debug("Unable to open a session or transaction", e);
+                            sleepWait();
```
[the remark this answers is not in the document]
lolski: I see what you're saying there. Taking that idea, what if spin it with a different approach which is by adding more context to the name - the function is only used to wait for the primary replica to be selected. Therefore I'll rename it to `waitForPrimaryReplicaSelection()`.

With the extra context in the name I no longer have to add parameters to the function since how long it needs to wait should already be clear. This way we can avoid adding the parameters and therefore making it "a generic util wait function".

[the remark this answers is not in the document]
alexjpwalker: Yeah that seems reasonable.

rpc/RPCSession.java
alexjpwalker: Looks good to me 👍

rpc/RPCSession.java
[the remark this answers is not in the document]
lolski: Fixed in https://github.com/graknlabs/client-java/pull/243/commits/641b5769611a9dd94adfbdff22a355d8b1f94677 and https://github.com/graknlabs/client-java/pull/243/commits/5dd3a4dbf4413e1afcad8ab06f31183505f04912.

rpc/RPCSession.java
lolski: Fixed in https://github.com/graknlabs/client-java/pull/243/commits/8f7cf33539175c537a906d305bdf68b4ff3de612

GraknOptions.java
```diff
@@ -60,4 +60,41 @@ public GraknOptions batchSize(int batchSize) {
...
+    public static GraknOptions.Cluster cluster() {
+        return new Cluster();
+    }
+
+    public boolean isCluster() {
+        return false;
+    }
+
+    public Cluster asCluster() {
+        if (isCluster()) return (Cluster) this;
```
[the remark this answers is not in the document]
lolski: This does look nicer: https://github.com/graknlabs/client-java/pull/243/commits/3041170384fada72b9160b43a15309d1a59a49bd

rpc/RPCSession.java
lolski: Yes, it is distinct from the concept of a "Grakn database" but I think it should be fine since it's already contextualised to be within `RPCSession`.

The name is consistent with the name that we are using in the protocol definition: https://github.com/graknlabs/protocol/blob/master/protobuf/cluster/database.proto.

I do feel that something here can be improved, but it would have to be the overall structure rather than just this one part. I think we should do it as part of the major refactor we're going to do after we're done with Cluster alpha.

What do you think @alexjpwalker ?

alexjpwalker: Perfectly fine with that

alexjpwalker (approved):

PR #248, opened 2021-01-28
Add new package for Cluster and delete Grakn interface

alexjpwalker:

We had Cluster-specific code living in the same classes as Grakn Core code. We made a conscious decision to view Cluster features as an extension of Core features, as the Cluster product is an extension of the Core product. To reflect this shift in mentality we created a new package named `rpc.cluster`. All Cluster functionality (besides that which cannot be easily moved, such as error messages) now lives there.

We also got rid of the Grakn interface, as it was a little odd having a method named `GraknClient.core` that returned `Grakn.Client`. The Client interface is now `GraknClient` and the Session, Transaction and DatabaseManager now live under `GraknClient` (e.g: `GraknClient.Transaction`)

- Move Cluster-specific code to rpc.cluster package
- Delete Grakn and use GraknClient as the root interface for the main classes (Session, Transaction etc)


GraknClient.java
```diff
@@ -19,169 +19,25 @@
...
+    String DEFAULT_ADDRESS = "localhost:1729";
 
-import static grakn.client.common.exception.ErrorMessage.Client.CLUSTER_UNABLE_TO_CONNECT;
-import static grakn.common.collection.Collections.pair;
-
-public class GraknClient {
-    public static final String DEFAULT_ADDRESS = "localhost:1729";
-
-    public static Core core() {
+    static Grakn.Client core() {
```
alexjpwalker, line 28: Note that because this is now an interface, `public` becomes implicit on these methods

lolski:
The changes look good!

I have two questions:

1. For consistency, should we call the classes under cluster `rpc/RPCClientCluster.java`, `rpc/RPCTransactionCluster.java`, and so on? 
2. With regards to the RPC classes, we prefix it with `RPC` (eg., `RPCTransaction`) whereas in Grakn Core we put it as a suffix (eg., `TransactionRPC`). I think we should make it more consistent and pick one style, since there's no reason not to. What do you think?

alexjpwalker:
>     1. For consistency, should we call the classes under cluster `rpc/RPCClientCluster.java`, `rpc/RPCTransactionCluster.java`, and so on?

We could do that. I took out the word RPC because the names were getting a bit long. `RPCClusterDatabaseManager` ...

>     2. With regards to the RPC classes, we prefix it with `RPC` (eg., `RPCTransaction`) whereas in Grakn Core we put it as a suffix (eg., `TransactionRPC`). I think we should make it more consistent and pick one style, since there's no reason not to. What do you think?

That's because `RPCTransaction` and `TransactionRPC` are different things.

An `RPCTransaction` is an implementation of the `Grakn.Transaction` interface using RPC. A `TransactionRPC` is a container object that holds a `RocksTransaction` and handles RPC calls within that transaction.

But I also feel the naming on the server side isn't ideal.

lolski:
@alexjpwalker 

> We could do that. I took out the word RPC because the names were getting a bit long. RPCClusterDatabaseManager ...

Yeah, I see that too. However, let's keep the long name and favour consistency especially in the early days of the new architecture. I think once we've got a more established architecture across repositories we can start making naming tweaks for brevity.

> That's because RPCTransaction and TransactionRPC are different things.
> An RPCTransaction is an implementation of the Grakn.Transaction interface using RPC. A TransactionRPC is a container object that holds a RocksTransaction and handles RPC calls within that transaction.
> But I also feel the naming on the server side isn't ideal.

Okay got it. Let's leave it for now and later we'll discuss it with Grakn Core people.

PR #251, opened 2021-01-28
Add the infrastructure for running BDD against both Grakn Core and Grakn Cluster

lolski:

We have rearchitected the `ConnectionSteps` and `BehaviourTest` BDD classes to support running BDD against both Grakn Core and Grakn Cluster in a generic way.

1. Split `ConnectionSteps` to `ConnectionStepsBase`, `ConnectionStepsCore`, and `ConnectionSteps` cluster.
2. Added `beforeAll` into the connection steps classes. In `ConnectionStepsCore`, the implementation will start Grakn Core whereas in `ConnectionStepsCluster` it will start Grakn Cluster instead.
3. `BehaviourTest` is responsible for stopping the server, with the use of JUnit's `@AfterClass` annotation.
4. Adds a BDD test rule `grakn_core_and_cluster_behaviour_test` that would automatically create two BDD target for one test: one against Grakn Core and another against Grakn Cluster


dependencies/graknlabs/artifacts.bzl
```diff
@@ -29,3 +29,13 @@ def graknlabs_grakn_core_artifacts():
         commit_source = deployment["artifact.snapshot"],
         commit = "b9d1cb52a7e261995d2f3650bbcf2626b4f62c3e",
     )
+
+def graknlabs_grakn_cluster_artifacts():
```
alexjpwalker, line 33: Now that we have both Core and Cluster I think we need to seriously consider if there's a better alternative than downloading 400MB of Grakn only to throw away 2/3rds of it.

I wouldn't know how to achieve this but we should definitely raise an issue I think.

dependencies/graknlabs/repositories.bzl
```diff
@@ -29,9 +29,13 @@ def graknlabs_graql():
...
         name = "graknlabs_common",
-        remote = "https://github.com/graknlabs/common",
-        tag = "2.0.0-alpha-6" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_common
+        remote = "https://github.com/lolski/common",
+        commit = "a2dafd47577cc63cee640ee331472f65722e7c70" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_common
     )
+#    native.local_repository(
+#        name = "graknlabs_common",
+#        path = "/Users/lolski/grakn.ai/graknlabs/common",
+#    )
```
alexjpwalker, line 38: lol

test/behaviour/concept/thing/attribute/BUILD
```diff
@@ -18,34 +18,49 @@
...
 
         # External Maven Dependencies
         "@maven//:junit_junit",
         "@maven//:io_cucumber_cucumber_java",
     ],
-    visibility = ["//visibility:public"],
 )
-
-grakn_java_test(
+grakn_core_and_cluster_behaviour_test(
```
alexjpwalker, line 43: Couple of issues here I think:

Firstly the name doesn't have "java" in it anymore, despite us having Java, Python and NodeJS flavours of this macro.

Secondly the name is really freaking long and I don't feel like adding `core_and_cluster` contributes much. Together, we see Core and Cluster as Grakn. I think `grakn_java_cucumber_test` would be enough, or `grakn_java_bdd_test`, or even just leave it as `grakn_java_test`.

(Now I know it expands, currently, to _two_ `grakn_java_test` targets. So I reckon ideally we would rename _that_ to something else, like `grakn_java_bdd_test_internal` or `grakn_java_bdd_test_specific_server`)

What I do like is the use of a macro in the first place - that's a much better solution than just duplicating `test` everywhere 🙂 Now the hope is that developers are able to figure out that the targets are actually named `test-core` and `test-cluster` (or whatever we named them)

test/behaviour/concept/thing/attribute/BUILD
```diff
@@ -18,34 +18,49 @@
...
+        "AttributeTest.java",
     ],
     test_class = "grakn.client.test.behaviour.concept.thing.attribute.AttributeTest",
+    data = [
+        "@graknlabs_behaviour//concept/thing:attribute.feature",
+    ],
+    native_grakn_artifact_core = "//test:native-grakn-core-artifact",
+    native_grakn_artifact_cluster = "//test:native-grakn-cluster-artifact",
+    connection_steps_core = "//test/behaviour/connection:steps-core",
+    connection_steps_cluster = "//test/behaviour/connection:steps-cluster",
```
alexjpwalker, line 55: It's a little bit arbitrary that the SetUp steps are in `ConnectionSteps`, but given how silly our solution for setup and teardown is anyway, I think we can safely overlook this.

test/behaviour/connection/ConnectionStepsBase.java
```diff
@@ -100,4 +93,15 @@ public synchronized void after() {
...
+    abstract GraknClient createGraknClient(String address);
+
+    void connectionHasBeenOpenedImpl() {
+        assertNotNull(client);
+        assertTrue(client.isOpen());
+    }
+
+    void connectionDoesNotHaveAnyDatabaseImpl() {
+        assertTrue(client.databases().all().isEmpty());
+    }
```
alexjpwalker, line 106: Having to suffix methods with `impl` is also ugly. Does it work if we rename `beforeImpl` and `afterImpl` to `before` and `after` and call `super.before()` in the subclasses?

Moreover, the step `connection does not have any database` should be moved to `DatabaseSteps`. I'm pretty sure nothing will break because _every_ test file uses `DatabaseSteps`.

As for `connection has been opened`, again, can we rename it to match the method in the subclass and call `super.connection_has_been_opened()`?

(Incidentally, did you test simply defining the test steps (`@When("connection has been opened")`) in this file? I assume you tried and Cucumber didn't like it)

alexjpwalker (changes requested):

haikalpribadi (approved):
Approving just for the //dependencies

dependencies/graknlabs/artifacts.bzl
lolski: Why would we throw away 2/3 of it? From that point onwards we are all expected to test both Grakn Core and Cluster right?

In any case, the solution is the fact that we're going to move cross-product testing to the `test` repo. The repo will have to be optimised to not have to always pull huge artifacts.

test/behaviour/concept/thing/attribute/BUILD
lolski: Yes, I was hoping to exchange ideas with regards to the name. I've thought of other names:

- `grakn_bdd_java_test`
- `grakn_behaviour_java_test`

I'm thinking of keeping the suffix `java_test` since it's consistent with Bazel, and with our other rule `host_compatible_java_test`.

I'm leaning towards `grakn_behaviour_java_test`. What do you think?

dependencies/graknlabs/repositories.bzl
lolski: Now you see it, now you don't

test/behaviour/connection/ConnectionStepsBase.java
lolski: > Having to suffix methods with impl is also ugly. Does it work if we rename beforeImpl and afterImpl to before and after and call super.before() in the subclasses?

> Moreover, the step connection does not have any database should be moved to DatabaseSteps. I'm pretty sure nothing will break because every test file uses DatabaseSteps.

> As for connection has been opened, again, can we rename it to match the method in the subclass and call super.connection_has_been_opened()?

all done in https://github.com/graknlabs/client-java/pull/251/commits/82e0af303bbd22fbd370e3537fa2aea7235e617b

test/behaviour/concept/thing/attribute/BUILD
lolski: yeah blood has been shed here

test/behaviour/connection/ConnectionStepsBase.java
lolski: > (Incidentally, did you test simply defining the test steps (@When("connection has been opened")) in this file? I assume you tried and Cucumber didn't like it)

Yes, this is correct. I tried it and it won't work

dependencies/graknlabs/artifacts.bzl
alexjpwalker: Because if you're running on Linux, the Windows and Mac artifacts are of no use to you.

test/behaviour/concept/thing/attribute/BUILD
alexjpwalker: Yep I like `grakn_behaviour_java_test`.

dependencies/graknlabs/repositories.bzl
```diff
@@ -29,10 +29,10 @@ def graknlabs_graql():
 def graknlabs_common():
     git_repository(
         name = "graknlabs_common",
-        remote = "https://github.com/graknlabs/common",
-        tag = "2.0.0-alpha-6" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_common
+        remote = "https://github.com/lolski/common",
+        commit = "a2dafd47577cc63cee640ee331472f65722e7c70" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_common
```
alexjpwalker, line 33: Is there a `common` PR inbound?

dependencies/graknlabs/artifacts.bzl
lolski: Ah yeah I see what you mean.

Okay, let's keep it this way and I'll create an issue for the `test` repo with regards to minimising download size in `grakn` (since we don't yet have the `test` repo).

FYI, the approach that we thought to try is to use multiple `WORKSPACE` file inside the `test` repo. We can have a different Bazel workspace per platform and also per product integration. The permutation will be high and we'll need to have an additional tooling to manage it, but I don't see any other way.

dependencies/graknlabs/repositories.bzl
lolski: Yes, for adding `GraknClusterRunner`. I'm still tidying it up.

alexjpwalker (approved):

PR #252, opened 2021-01-28
Add support for setting prefetch, session timeout, and schema lock timeout in GraknOptions

lolski:

We have added support for setting prefetch, session timeout, and schema lock timeout in GraknOptions.

- Update `GraknOptions` with the new settings

### DIFF
--- a/GraknOptions.java
+++ b/GraknOptions.java
@@ -30,6 +30,20 @@ public class GraknOptions {
     private Boolean infer = null;
     private Boolean explain = null;
     private Integer batchSize = null;
+    private Boolean prefetch = null;
+    private Integer sessionIdleTimeout = null;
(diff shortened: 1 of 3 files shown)


GraknOptions.java
```diff
@@ -23,13 +23,28 @@
 
 import java.util.Optional;
 
-import static grakn.client.common.exception.ErrorMessage.Client.NEGATIVE_BATCH_SIZE;
+import static grakn.client.common.exception.ErrorMessage.Client.NEGATIVE_VALUE_NOT_ALLOWED;
```
alexjpwalker, line 26: Just a note - I think we had this specific message because we ideally want each error message to refer to a specific error. But honestly, I think this may need a broader rethink so this is fine for now.

alexjpwalker (approved):

PR #253, opened 2021-01-29
Support Graql update query

lolski:

We have added support for the update query.

- Updgrade Graql for the `GraqlUpdate` object
- Update Protocol for the `Update` protocol definition
- Added `update` methods in `QueryManager`


alexjpwalker (approved):

haikalpribadi (approved):

alexjpwalker:
There are many failing tests, perhaps because the Grakn Core artifact has not been updated?

PR #262, opened 2021-02-05
Make DatabaseManager more fault tolerant in Cluster

alexjpwalker:

We've introduced a couple of fault tolerance improvements into the Grakn Cluster version of DatabaseManager:

- `contains` and `all` will now try to retrieve information from every node, instead of failing if the first node is down
- `create` and `delete` will now be applied to every working node, even if some nodes are down.

For now, this means `create` will no longer throw if the database already exists, and `delete` will no longer throw if the database doesn't exist.

- `contains` and `all` will now try to retrieve information from every node, instead of failing if the first node is down
- `create` and `delete` will now be applied to every working node, even if some nodes are down.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -20,12 +20,3 @@
 build --incompatible_strict_action_env --javacopt='--release 8'
 run --incompatible_strict_action_env
 test --incompatible_strict_action_env --test_env=PATH
-
-# what is defined in this section will be applied when bazel is invoked like this: bazel ... --config=rbe ...
-build:rbe --project_id=grakn-dev
-build:rbe --remote_cache=cloud.buildbuddy.io
-build:rbe --bes_backend=cloud.buildbuddy.io
-build:rbe --bes_results_url=https://app.buildbuddy.io/invocation/
-build:rbe --tls_client_certificate=/opt/credentials/buildbuddy-cert.pem
-build:rbe --tls_client_key=/opt/credentials/buildbuddy-key.pem
-build:rbe --remote_timeout=3600
--- a/.bazelrc
+++ b/.bazelrc
@@ -19,7 +19,7 @@
 
 build --incompatible_strict_action_env --javacopt='--release 8'
 run --incompatible_strict_action_env
-test --incompatible_strict_action_env
+test --incompatible_strict_action_env --test_env=PATH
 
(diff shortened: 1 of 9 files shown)


GraknClient.java
```diff
@@ -38,12 +38,8 @@ static GraknClient core(String address) {
...
     }
 
-    static GraknClient cluster() {
-        return cluster(DEFAULT_ADDRESS);
-    }
-
-    static GraknClient cluster(String address) {
-        return new RPCGraknClientCluster(address);
+    static GraknClient cluster(String... addresses) {
+        return new RPCGraknClientCluster(addresses);
```
alexjpwalker, line 42: This change allows Cluster clients to be instantiated with failover.

dependencies/graknlabs/repositories.bzl
```diff
@@ -30,7 +30,7 @@ def graknlabs_common():
     git_repository(
         name = "graknlabs_common",
         remote = "https://github.com/graknlabs/common",
-        tag = "2.0.0-alpha-7" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_common
+        commit = "36564a64904eed7d06d6d7c24055ffc70a40aee8" # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_common
```
alexjpwalker, line 33: This increases the time we wait for Grakn to start, in tests, to 15 seconds.

test/behaviour/connection/database/BUILD
```diff
@@ -41,28 +41,27 @@ java_library(
     ],
 )
 
-grakn_behaviour_java_test(
-    name = "test",
+grakn_java_test(
+    name = "test-core",
```
alexjpwalker, line 45: We need to split up this target in order to specify different `@ignore` tags for each.

lolski (approved):

.bazelrc
```diff
@@ -19,7 +19,7 @@
 
 build --incompatible_strict_action_env --javacopt='--release 8'
 run --incompatible_strict_action_env
-test --incompatible_strict_action_env
+test --incompatible_strict_action_env --test_env=PATH
```
alexjpwalker, line 22: This change allows Bazel to use the `lsof` utility. Perhaps, in the future, we should change it and include `lsof` in our Bazel runfiles.

haikalpribadi (approved):

PR #269, opened 2021-02-11
Clean up FailsafeTask

alexjpwalker:

We did some general cleanup in FailsafeTask.

- Make FailsafeTask take in the Database as a constructor parameter, instead of having it as an argument in every method
- Fix ReplicaInfo performing an unnecessary aggregate over the Replicas when finding a preferred secondary

### DIFF
--- a/rpc/cluster/FailsafeTask.java
+++ b/rpc/cluster/FailsafeTask.java
@@ -26,6 +26,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


rpc/cluster/FailsafeTask.java
```diff
@@ -41,9 +42,11 @@
     private static final int FETCH_REPLICAS_MAX_RETRIES = 10;
     private static final int WAIT_FOR_PRIMARY_REPLICA_SELECTION_MS = 2000;
     private final ClientClusterRPC client;
+    private final String database;
 
-    FailsafeTask(ClientClusterRPC client) {
+    FailsafeTask(ClientClusterRPC client, String database) {
```
alexjpwalker, line 47: We were passing `database` in as a parameter to _every_ method, so it was pretty clear it should've been a ctor parameter.

rpc/cluster/ReplicaInfo.java
```diff
@@ -62,7 +62,7 @@ public static ReplicaInfo ofProto(DatabaseProto.Database.Replicas.Res res) {
     Replica preferredSecondaryReplica() {
         return replicas.values().stream()
                 .filter(Replica::isPreferredSecondary)
-                .max(Comparator.comparing(Replica::term))
```
alexjpwalker, line 65: This comparison is simply unnecessary, the server only ever selects one preferred secondary replica.

lriuui0x0 (approved):

PR #270, opened 2021-02-12
Expose replica info when getting Cluster databases

alexjpwalker:

We now expose additional information when fetching databases in Cluster. The database name, server address, primary/secondary state, and term number are all available.

This is a **breaking change**. `client.databases().all()` now returns `List<Database>` (previously it was `List<String>`). To get the old functionality (listing the names), use `client.databases().all().stream().map(GraknClient.Database::name)`.

Additionally, `client.databases().delete()` has been deleted, and superseded by the new syntax: `client.databases().get(name).delete()`.

- Delete `client.databases().delete`, and add `client.databases().get`, which returns a new `Database` object that has a `delete` method
- Add `GraknClient.Cluster` and `GraknClient.DatabaseManager.Cluster`, which return `Database.Cluster` objects when `all` and `get` are called. `Database.Cluster` objects expose their database name, server address, primary/secondary state and term number.


.grabl/automation.yml
```diff
@@ -65,44 +65,39 @@ build:
         export ARTIFACT_USERNAME=$REPO_GRAKN_USERNAME
         export ARTIFACT_PASSWORD=$REPO_GRAKN_PASSWORD
         bazel run @graknlabs_dependencies//distribution/artifact:create-netrc
-        bazel test //test/integration/... --test_output=streamed
+        bazel test //test/integration/... --test_output=errors
```
alexjpwalker, line 68: These tests are way too talkative with streamed output - the output can be 90,000 lines long.

GraknClient.java
```diff
@@ -38,29 +41,74 @@ static GraknClient core(String address) {
...
 
         boolean contains(String name);
 
+        // TODO: Return type should be 'Database' but right now that would require 2 server calls in Cluster
         void create(String name);
 
-        void delete(String name);
+        Database get(String name);
+
+        List<? extends Database> all();
```
alexjpwalker, line 73: Can't use `List<Database>` because `List<Database.Cluster>` is a weaker type constraint than `List<Database>`

GraknClient.java
```diff
@@ -38,29 +41,74 @@ static GraknClient core(String address) {
...
+    }
+
     interface DatabaseManager {
 
         boolean contains(String name);
 
+        // TODO: Return type should be 'Database' but right now that would require 2 server calls in Cluster
         void create(String name);
 
-        void delete(String name);
```
alexjpwalker, line 61: We move `delete` from `DatabaseManager` to the new `Database` interface. This aligns the structure more closely with Grakn Core, which has `delete` also on the `RocksDatabase`.

dependencies/graknlabs/repositories.bzl
```diff
@@ -43,8 +43,8 @@ def graknlabs_dependencies():
 def graknlabs_protocol():
     git_repository(
         name = "graknlabs_protocol",
-        remote = "https://github.com/graknlabs/protocol",
-        tag = "2.0.0-alpha-9", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
+        remote = "https://github.com/alexjpwalker/protocol",
+        commit = "072e1cb3470e1354e308b3eaa6e02355e11636e5", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
```
alexjpwalker, line 47: Update to graknlabs before merging

rpc/cluster/DatabaseClusterRPC.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+class DatabaseClusterRPC implements GraknClient.Database.Cluster {
```
alexjpwalker, line 38: This class used to be called `SessionClusterRPC.Database`. Making it a top-level class feels more correct.

GraknClient.java
```diff
@@ -38,29 +41,74 @@ static GraknClient core(String address) {
...
+        Database get(String name);
```
alexjpwalker, line 71: The main purpose of `get` is to retrieve a `Database` in order to call `delete` on it.

`get` throws `GraknClientException` if the database does not exist.

.grabl/automation.yml
lolski: Right. Is that because of Grakn Core or Grakn Cluster?

.grabl/automation.yml
```diff
@@ -113,7 +108,7 @@ build:
         bazel test //test/behaviour/graql/language/define:test-core --test_output=streamed --jobs=1
         bazel test //test/behaviour/graql/language/undefine:checkstyle --test_output=streamed
         bazel test //test/behaviour/graql/language/undefine:test-core --test_output=streamed --jobs=1
-
+    # TODO: turn on cluster tests once cluster tests pass for definable
```
lolski, line 111: Unrelated to this PR, but do you know why define test fails on cluster?

test/behaviour/connection/database/BUILD
```diff
@@ -30,6 +30,7 @@ java_library(
     visibility = ["//visibility:public"],
     deps = [
         # Internal Package Dependencies
+        "//:client-java",
```
lolski, line 33: Why do we need to depend on `//:client-java` now? Is it because we now return `Database` rather than `String`? Or are there other reasons?

rpc/ClientRPC.java
```diff
@@ -69,4 +72,12 @@ public void close() {
     public Channel channel() {
         return channel;
     }
+
+    static <RES> RES rpcCall(Supplier<RES> req) {
```
lolski, line 76: Do o it weird that now we have to import `ClientRPC` from every class in order for those class to access `rpcCall`? What if we put the method under the common package? Especially since the method is quite generic and free of context

rpc/DatabaseRPC.java
```diff
@@ -0,0 +1,67 @@
...
+    public String name() {
+        return name;
+    }
+
+    @Override
+    public void delete() {
+        rpcCall(() -> blockingGrpcStub.databaseDelete(DatabaseProto.Database.Delete.Req.newBuilder().setName(nonNull(name)).build()));
+    }
+
+    @Override
```
lolski, line 49: Are we implementing `equals` and `hashCode` so that we can compare one RPC object to another?

What if we simply compare the name rather than the object itself? (eg., `rpc1.name().equals(rpc2.name())` instead of `rpc1.equals(rpc2)` 

When it comes to equality checks, I find it a bit too magical when implemented on complex objects such as RPC. 

Would an RPC object for the same database be considered the same object even if the underlying connection is different? Given that the object itself is called "RPC", it may be surprising that we actually do not consider the underlying RPC object for determining equality.

.grabl/automation.yml
alexjpwalker: There isn't a significant difference between Core and Cluster when it comes to our BDD tests.

Most of the logging is:
- printing out every BDD step, including the backgrounds that repeat 100+ times
- printing out every exception stack trace
- printing out whatever else the server says

.grabl/automation.yml
alexjpwalker: I have not looked into it.

test/behaviour/connection/database/BUILD
alexjpwalker: Yeah I think that's literally it haha, we need to import `Database`

rpc/ClientRPC.java
alexjpwalker: That's definitely one option

alexjpwalker: Given that it's only used by the `rpc` package I moved it to `rpc.util.RPCUtils`.

rpc/DatabaseRPC.java
alexjpwalker: Hmm. Yes I can see that that's weird. Let's not override equals here.

alexjpwalker: Deleted `equals`, `hashCode`

lolski (approved):

PR #281, opened 2021-02-18
Add option to enable inference tracing

jmsfltchr:

We introduce a new option to allow tracing the internals of Grakn's reasoner.

- Adds a new option, the flag `traceInference`.

Requires https://github.com/graknlabs/protocol/pull/122

### DIFF
--- a/GraknOptions.java
+++ b/GraknOptions.java
@@ -28,7 +28,7 @@ import static grakn.client.common.exception.ErrorMessage.Internal.ILLEGAL_CAST;
 
 public class GraknOptions {
     private Boolean infer = null;
-    private Boolean inferLog = null;
+    private Boolean traceInference = null;
     private Boolean explain = null;
     private Integer batchSize = null;
     private Boolean prefetch = null;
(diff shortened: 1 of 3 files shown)


alexjpwalker (approved):

GraknOptions.java
```diff
@@ -57,6 +58,15 @@ public GraknOptions infer(boolean infer) {
...
     }
 
+    public Optional<Boolean> inferLog() {
+        return Optional.ofNullable(inferLog);
+    }
+
+    public GraknOptions inferLog(boolean inferLog) {
+        this.inferLog = inferLog;
+        return this;
+    }
```
haikalpribadi, line 68: are we sure this is the best name? not `traceInference` ? if we go with "log", I think it should at least be `logInference`?

haikalpribadi (changes requested):

GraknOptions.java
jmsfltchr: We first went with `inferenceLogging` and changed infer` -> `inference` for consistency, but @alexjpwalker pointed out that `inference` didn't look consistent with the other options. We then went with `inferLog` since from a user perspective this is clearly scoped to relate to the existing `infer` flag.
@alexjpwalker does `traceInference` sound good to you?

alexjpwalker: `traceInference` definitely sounds nicer and is consistent with the other flags, but not with `infer` specifically. Hmmm.

haikalpribadi (approved):

PR #284, opened 2021-02-22
Client tracks and closes its sessions when closed

alexjpwalker:

The `close` method of `GraknClient` now automatically closes all `Session` objects that were opened from that client.

The `close` method of `GraknClient` now automatically closes all `Session` objects that were opened from that client.


rpc/ClientRPC.java
```diff
@@ -48,7 +53,9 @@ public SessionRPC session(String database, GraknClient.Session.Type type) {
 
     @Override
     public SessionRPC session(String database, GraknClient.Session.Type type, GraknOptions options) {
-        return new SessionRPC(this, database, type, options);
+        SessionRPC session = new SessionRPC(this, database, type, options);
+        sessions.put(session.id(), session);
+        return session;
```
flyingsilverfin, line 58: can we just pop an assert in that says the ID is not in the sessions map???

flyingsilverfin (approved):

PR #285, opened 2021-02-23
Recover faster from temporary network outages and server failures

alexjpwalker:

In the event of a network interruption, the client would previously start "automatically" rejecting all requests for a certain period of time. We now force the client to try to reconnect if the user performs any operation, meaning that if the server is restored after a temporary outage, it will now be available immediately to process fresh requests.

After any server outage, always try to reconnect when performing any operation

## Additional information

This PR is based off https://github.com/graknlabs/client-java/pull/284 and should be merged only once that PR is.


test/behaviour/connection/ConnectionStepsBase.java
```diff
@@ -68,6 +71,12 @@ void before() {
     }
 
     void after() {
+        // TODO: Remove this once the server segfault issue is fixed (grakn#6135)
+        try {
+            Thread.sleep(10);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
```
flyingsilverfin, line 79: can we get rid of this now? or at least the TODO

alexjpwalker: Umm I think this is still needed because there is an intermittent issue when closing the sessions at the end of each scenario where the server segfaults. The issue is NOT resolved, to the best of my knowledge.

flyingsilverfin (approved):

PR #294, opened 2021-03-05
Handle gRPC's StatusRuntimeException with status UNKNOWN

lolski:

It is possible to receive a `StatusRuntimeException ` exception with status code `UNKNOWN` when gRPC isn't ready. Therefore, we've added a logic to handle it in the `FailSafeTask` class.

- Treat `StatusRuntimeException` with status code `UNKNOWN` as non-fatal and should be retried.

### DIFF
--- a/rpc/cluster/FailsafeTask.java
+++ b/rpc/cluster/FailsafeTask.java
@@ -65,10 +65,7 @@ abstract class FailsafeTask<TResult> {
             try {
                 return retries == 0 ? run(replica) : rerun(replica);
             } catch (GraknClientException e) {
-                if (CLUSTER_REPLICA_NOT_PRIMARY.equals(e.getErrorMessage()) ||
-                        UNABLE_TO_CONNECT.equals(e.getErrorMessage()) ||
-                        e.getCause() instanceof StatusRuntimeException && ((StatusRuntimeException) e.getCause()).getStatus() == Status.UNKNOWN
-                ) {
+                if (nonFatalException(e)) {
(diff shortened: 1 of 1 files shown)


rpc/cluster/FailsafeTask.java
```diff
@@ -64,7 +65,10 @@ TResult runPrimaryReplica() {
             try {
                 return retries == 0 ? run(replica) : rerun(replica);
             } catch (GraknClientException e) {
-                if (CLUSTER_REPLICA_NOT_PRIMARY.equals(e.getErrorMessage()) || UNABLE_TO_CONNECT.equals(e.getErrorMessage())) {
+                if (CLUSTER_REPLICA_NOT_PRIMARY.equals(e.getErrorMessage()) ||
+                        UNABLE_TO_CONNECT.equals(e.getErrorMessage()) ||
+                        e.getCause() instanceof StatusRuntimeException && ((StatusRuntimeException) e.getCause()).getStatus() == Status.UNKNOWN
+                ) {
```
alexjpwalker, line 71: If gRPC has the good, the bad and the ugly parts, this is definitely a cross between bad and ugly.

However, once we introduce our own `Error` message into protocol, we'll be able to replace that logic with just `if e instanceof StatusRuntimeException`.

So let's go ahead with this for now!

alexjpwalker (approved):

rpc/cluster/FailsafeTask.java
lolski: I've extracted the condition into a function `nonFatalException`

PR #302, opened 2021-03-15
Architecture Refactor

haikalpribadi:

Over time, Client Java has been extended considerably to implement various features and optimisations. It has lead the architecture to a complex codebase which is not so easy to follow and replicate in other languages. We have now refactored Client Java considerably to come up with an architecture that is easy to understand, easy to extend, and easy to replicate in other languages that wish to implement a Grakn Client.

1) Made the Transaction Bidirectional Stream architecture a first class citizen of the architecture, making it obvious and easier to understand and follow. The Bidirectional Stream now gets its own package at `//stream`.
2) Made the two main responsibilities of the bidirectional stream - single response and streamed multipart response - to be explicitly enforced and expressed.
3) Simplified the response parsing logic across the codebase from passing around lambdas to eventually be applied at the lowest level (at "ResponseIterator"), by reversing the logic such that every lower-level class returns the top-level response type and every top level class will unpack the response before returning upwards.
4) Introducing the the "API" package in which all interfaces are aggregated into. This allows us to remove cycles across the classes that was complicated to follow in other language implementations of Grakn Client.
5) Split the monolithic repository (there was only one "package" that compiled all source code) into multiple packages, enforcing and expressing a clear hierarchy. 

The diagram below describes the new architecture of client-java, which should be able to be replicated across every other language implementation of Grakn Clients.

package-structure


concept/thing/AttributeImpl.java
```diff
@@ -99,26 +99,22 @@
...
         }
 
         @Override
         public final Stream<ThingImpl> getOwners() {
-            return thingStream(
-                    ConceptProto.Thing.Req.newBuilder().setAttributeGetOwnersReq(
-                            ConceptProto.Attribute.GetOwners.Req.getDefaultInstance()),
-                    res -> res.getAttributeGetOwnersRes().getThingsList()
-            );
+            return stream(Proto.Thing.Attribute.getOwners(getIID()))
```
alexjpwalker, line 108: Something I notice is we now have to write `getIID` at least once in virtually every concept method - a duplication that was previously avoided by use of `thingStream` and lambda functions.

I think, with this architecture, that is unavoidable.

concept/type/TypeImpl.java
```diff
@@ -77,6 +79,29 @@ public static TypeImpl of(ConceptProto.Type typeProto) {
         }
     }
 
+    public static ConceptProto.Type.Encoding encoding(Type type) {
```
alexjpwalker, line 82: Perhaps we should move this to `Proto.java`.

core/CoreClient.java
```diff
@@ -135,11 +136,11 @@ void reconnect() {
         }
     }
 
-    TransactionRequestBatcher batcher() {
+    RequestTransmitter executor() {
```
alexjpwalker: I love the name `RequestTransmitter`! But should we now rename this:
```suggestion
    RequestTransmitter transmitter() {
```

core/CoreClient.java
```diff
@@ -35,27 +36,27 @@
...
-public class ClientRPC implements GraknClient {
+public class CoreClient implements Client {
 
     private static final String GRAKN_CLIENT_RPC_THREAD_NAME = "grakn-client-rpc";
 
     private final ManagedChannel channel;
-    private final TransactionRequestBatcher batcher;
-    private final DatabaseManagerRPC databases;
-    private final ConcurrentMap<ByteString, SessionRPC> sessions;
+    private final RequestTransmitter batcher;
```
alexjpwalker:
```suggestion
    private final RequestTransmitter transmitter;
```

core/CoreDatabaseManager.java
```diff
@@ -17,51 +17,54 @@
...
-    public DatabaseManagerRPC(ClientRPC client) {
+    public CoreDatabaseManager(CoreClient client) {
         this.client = client;
         blockingGrpcStub = GraknGrpc.newBlockingStub(client.channel());
     }
 
     @Override
-    public boolean contains(String name) {
-        return client.call(() -> blockingGrpcStub.databaseContains(DatabaseProto.Database.Contains.Req.newBuilder().setName(nonNull(name)).build()).getContains());
+    public boolean contains(String n) {
```
alexjpwalker, line 44: hmmm 
```suggestion
    public boolean contains(String name) {
```

(Also applies to `create`, see below)

stream/BidirectionalStream.java
```diff
@@ -0,0 +1,154 @@
...
+            resPartCollector.close(error);
+            try {
+                dispatcher.close();
+            } catch (StatusRuntimeException e) {
+                throw GraknClientException.of(e);
+            }
+        }
+    }
+
+    private StreamObserver<Transaction.Server> responseObserver() {
```
alexjpwalker: I find these clearer to follow when we don't use aliases - ie:
```suggestion
    private StreamObserver<TransactionProto.Transaction.Server> responseObserver() {
```

lolski (approved):

stream/ResponseCollector.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    public synchronized void close(@Nullable StatusRuntimeException error) {
+        collectors.values().forEach(collector -> collector.close(error));
+        collectors.clear();
+    }
+
+    public static class Queue<R> {
+
+        private final BlockingQueue<Either<Response<R>, Done>> responseQueue;
```
alexjpwalker, line 62: I think it might be cleaner to make `Response` a base class, with `Ok` and `Done` as subclasses - `Either` has a fairly unclear API and I also don't think it exists in many languages.

stream/ResponseIterator.java
```diff
@@ -0,0 +1,92 @@
...
+
+public class ResponseIterator implements Iterator<TransactionProto.Transaction.ResPart> {
+
+    private final UUID requestID;
+    private final RequestTransmitter.Dispatcher dispatcher;
+    private final ResponseCollector.Queue<TransactionProto.Transaction.ResPart> responseCollector;
+    private TransactionProto.Transaction.ResPart next;
+    private State state;
+
+    enum State {EMTPY, FETCHED, DONE}
```
alexjpwalker, line 41: My hawk eyes say:
```suggestion
    enum State {EMPTY, FETCHED, DONE}
```

alexjpwalker (changes requested):

cluster/FailsafeTask.java
```diff
@@ -17,54 +17,56 @@
...
-import static grakn.client.common.exception.ErrorMessage.Client.CLUSTER_UNABLE_TO_CONNECT;
-import static grakn.client.common.exception.ErrorMessage.Client.UNABLE_TO_CONNECT;
-import static grakn.client.common.exception.ErrorMessage.Internal.UNEXPECTED_INTERRUPTION;
+import static grakn.client.common.ErrorMessage.Client.CLUSTER_REPLICA_NOT_PRIMARY;
+import static grakn.client.common.ErrorMessage.Client.CLUSTER_UNABLE_TO_CONNECT;
+import static grakn.client.common.ErrorMessage.Client.UNABLE_TO_CONNECT;
+import static grakn.client.common.ErrorMessage.Internal.UNEXPECTED_INTERRUPTION;
 
-abstract class FailsafeTask<TResult> {
+abstract class FailsafeTask<E> {
```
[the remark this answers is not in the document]
haikalpribadi: Let's go with `RESULT`

concept/type/TypeImpl.java
haikalpribadi: We can't do that, because `Type` is in `//api`, and this method signatures depend on it.

common/Proto.java
```diff
@@ -0,0 +1,663 @@
...
+        }
+
+        public static SessionProto.Session.Close.Req close(ByteString sessionID) {
+            return SessionProto.Session.Close.Req.newBuilder().setSessionId(sessionID).build();
+        }
+    }
+
+    public static class Transaction {
+
+        public static TransactionProto.Transaction.Client clientMsg(ArrayList<TransactionProto.Transaction.Req> reqs) {
```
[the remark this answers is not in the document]
haikalpribadi: 👍

stream/ResponseCollector.java
haikalpribadi: let's keep it as discussed in person

alexjpwalker (approved):

api/Client.java
```diff
@@ -0,0 +1,54 @@
...
+
+package grakn.client.api;
+
+import grakn.client.api.database.DatabaseManager;
+
+import javax.annotation.CheckReturnValue;
+
+public interface Client extends AutoCloseable {
+
+    @CheckReturnValue
```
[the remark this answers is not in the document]
haikalpribadi: 👍

Issue #277, opened 2021-02-18
Organise ErrorMessages better


alexjpwalker:
## Description

Our error messages are pretty inconsistent and some of them are odd and misleading.

For example:

### [CLI2] Illegal Client Operation: Unable to connect to Grakn server.
How is this an illegal client operation? Sounds more like a connection error to me.

### [CLI3] Illegal Client Operation: Value cannot be less than 1, was '-42'.
Value? What value? This error message (which I believe refers to batch size) should be more precise.

### [CLI7] Illegal Client Operation: Received a response with unknown request id '00000000-0000-0000-0000-000000000000'
This is definitely our own internal issue, and should probably be an "Internal Error".

### And also...
I'm pretty sure there are cases in the codebase where we actually don't have a dedicated error message and we just throw "whatever".

All of these should be cleaned up.

PR #303, opened 2021-03-17
Clear up redundant methods, reduce code duplication

alexjpwalker:

We cleared up some redundant methods, reduced some code duplication by use of default interface methods, and significantly cleaned up `GraknClientException` by refactoring out some of its constructors.

- Clear out some unused methods
- Use default interface methods where doing so is trivially simple and would reduce code duplication
- Clean out 2 constructors of `GraknClientException` (it had 4, making it a nightmare to replicate in other clients!!)

### DIFF
--- a/api/GraknOptions.java
+++ b/api/GraknOptions.java
@@ -102,7 +102,7 @@ public class GraknOptions {
 
     public GraknOptions batchSize(int batchSize) {
         if (batchSize < 1) {
-            throw new GraknClientException(NEGATIVE_VALUE_NOT_ALLOWED.message(batchSize));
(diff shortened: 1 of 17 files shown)


haikalpribadi (approved):
Approving, but in the future can we replace `new GraknClientExcpetion` with `GraknClientException.of(...)` ? (a) it would be nicer to have builder methods such as `.of()` to do any logic before constructing a class (which should have no logic if possible), and (b) it would be aligned with how we do it in Grakn server.

alexjpwalker:
https://github.com/graknlabs/client-java/issues/277 updated with new information

PR #304, opened 2021-03-19
New Database.schema(), resilient GraknStub, and update Protocol Architecture

haikalpribadi:

There are three key changes in this PR - 1 new feature, 1 refactor & feature, 1 purely refactor.

1) We introduced a new method `schema()` on the `Database` object, which allows you to export the schema of the database and get it back as a `String`. This was a pre-existing feature of Grakn server, but was not easily accessible through the client drivers. It is now defined in our protocol, and readily available through client-java.

2) We had the mechanism to ensure that RPC calls on the GRPC stubs to always (a) ensure that the channel is still alive, and (b) catch any exception that is thrown. We applied this mechanism by calling `CoreClient.call(() -> ...)`, but this was not the best implementation. It was not clear why lower-level objects, such as Session and Database needs to call their parent object Client when they can call the stub themselves, and the method name did not clearly describe what it's responsibility is. There was also no way of guaranteeing that the developers extending the client functionalities would always call `CoreClient.call(..)` for every call they want to do on the GRPC stubs - this made it error prone. We introduced a class `ResilientStub`, which wraps a GRPC stub, and ensures that we apply the aforementioned safety mechanisms strictly. When passing around ResilientStub, developers are now guaranteed to to always perform GRPC stub calls in a safe way.

3) Following the protocol refactors in https://github.com/graknlabs/protocol/pull/130 we have now updated client-java to reflect the new protocol architecture.


cluster/ClusterClient.java
```diff
@@ -42,9 +42,10 @@
 import static grakn.common.collection.Collections.pair;
 
 public class ClusterClient implements GraknClient.Cluster {
+
     private static final Logger LOG = LoggerFactory.getLogger(ClusterClient.class);
     private final Map<String, CoreClient> coreClients;
-    private final Map<String, GraknClusterGrpc.GraknClusterBlockingStub> graknClusterRPCs;
+    private final Map<String, ResilientStub.Cluster> blockingStubs;
```
alexjpwalker, line 48: I like the fact that `ResilientStub` has been introduced. Yet I think we can actually do better than the name `ResilientStub`.

What actually is the purpose of `ResilientStub`? A stub that is resilient, or an API for accessing Grakn Protocol service methods?

I'd say the implementation is a resilient stub, the purpose is a Grakn service.

So how about we rename it to `GraknService`, or `GraknProtoService`, or `ProtoService`, or ...

core/CoreClient.java
```diff
@@ -143,4 +127,19 @@ RequestTransmitter transmitter() {
     void removeSession(CoreSession session) {
         sessions.remove(session.id());
     }
+
+    void reconnect() {
```
lolski, line 131: Can this be removed now?

cluster/ClusterDatabaseManager.java
```diff
@@ -67,8 +67,8 @@ public void create(String name) {
         StringBuilder errors = new StringBuilder();
         for (String address : databaseManagers.keySet()) {
             try {
-                DatabaseProto.Database.Get.Res res = client.coreClient(address).call(
-                        () -> client.graknClusterRPC(address).databaseGet(DatabaseProto.Database.Get.Req.newBuilder().setName(name).build())
+                ClusterDatabaseProto.ClusterDatabaseManager.Get.Res res = client.coreClient(address).call(
```
alexjpwalker, line 70: Oh, I never noticed this before, but shouldn't the construction of the `ClusterDatabaseProto.ClusterDatabaseManager.Get.Req` be moved to `RequestBuilder`?

alexjpwalker, line 70: Also we still appear to be using `client.call` here

lolski (approved):

cluster/ClusterClient.java
```diff
@@ -113,7 +114,7 @@ ClusterSession run(ClusterDatabase.Replica replica) {
         };
     }
 
-    ConcurrentMap<String, ClusterDatabase> clusterDatabases() {
+    ConcurrentMap<String, ClusterDatabase> databaseByAddress() {
```
alexjpwalker, line 117: I love the new name, much clearer about the purpose! But also inaccurate... I believe the keys in this map are database names, not server addresses.

So we should probably rename this to `databaseByName`.

core/CoreClient.java
```diff
@@ -143,4 +127,19 @@ RequestTransmitter transmitter() {
...
+        // TODO: remove
+    }
```
alexjpwalker, line 133: I guess we remove this now?

alexjpwalker (changes requested):

cluster/ClusterClient.java
haikalpribadi: 👍 we've renamed it to GraknStub now

cluster/ClusterClient.java
haikalpribadi: 👍

stream/BidirectionalStream.java
```diff
@@ -48,11 +47,11 @@
     private final RequestTransmitter.Dispatcher dispatcher;
     private final AtomicBoolean isOpen;
 
-    public BidirectionalStream(Channel channel, RequestTransmitter transmitter) {
+    public BidirectionalStream(GraknStub.Core stub, RequestTransmitter transmitter) {
         resPartCollector = new ResponseCollector<>();
         resCollector = new ResponseCollector<>();
         isOpen = new AtomicBoolean(false);
-        dispatcher = transmitter.dispatcher(GraknGrpc.newStub(channel).transaction(new ResponseObserver()));
```
alexjpwalker, line 55: We should be wary of this change and how it affects opening many parallel transactions - excessively reusing Stubs has caused stalling before in client-python. But I don't _think_ we've seen such issues in client-java before, just something to be aware for.

alexjpwalker (approved):

PR #305, opened 2021-03-21
Resilient stub exception handling

lolski:

The `ResilientStub` that was recently introduced wraps gRPC's `StatusRuntimeException` into `GraknClientException`. Therefore we've updated how we handle exception in various places.

- Catch `GraknClientException` instead of `StatusRuntimeException` when we want to recover from the condition
- Remove code that attempts to wrap `StatusRuntimeException` into `GraknClientException` since it's already inside `ResilientStub`


alexjpwalker:
Superseded by https://github.com/graknlabs/client-java/pull/309

PR #307, opened 2021-03-22
Implement Explanations

flyingsilverfin:
Following https://github.com/graknlabs/grakn/pull/6271 and the corresponding protocol change in https://github.com/graknlabs/protocol/pull/131 we implement Explanations, Explainable concept maps, and the explain() query API, which allows users to stream Explanations on demand **note: explain query or transaction option must be set to `true`**

* Implement `Explanation` objects, and extend `ConceptMap` to contain `Explainables`
* Add the `QueryManager.explain(Explainable)` API to retrieve all direct explanations (1-rule layer) 
* update `ClientTest` to include a simple Explanation test

### DIFF
--- a/api/answer/ConceptMap.java
+++ b/api/answer/ConceptMap.java
@@ -20,11 +20,15 @@
 package grakn.client.api.answer;
 
 import grakn.client.api.concept.Concept;
+import grakn.client.common.exception.GraknClientException;
+import grakn.common.collection.Pair;
 
 import javax.annotation.CheckReturnValue;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
+
+import static grakn.client.common.exception.ErrorMessage.Concept.NONEXISTENT_EXPLAINABLE_CONCEPT;
 
 public interface ConceptMap {
 
@@ -37,13 +41,25 @@ public interface ConceptMap {
     @CheckReturnValue
     Concept get(String variable);
 
-    Set<Explainable> explainables();
+    Explainables explainables();
 
-    interface Explainable {
+    interface Explainables {
 
-        String conjunction();
+        Explainable concept(String variable);
 
-        long id();
+        Explainable ownership(String owner, String attribute);
+
+        Map<String, Explainable> explainableConcepts();
+
+        Map<Pair<String, String>, Explainable> explainableOwnerships();
+
+        interface Explainable {
+
+            String conjunction();
+
+            long id();
+
+        }
 
     }
 }
--- a/api/answer/ConceptMap.java
+++ b/api/answer/ConceptMap.java
@@ -47,11 +47,11 @@ public interface ConceptMap {
 
         Explainable ownership(String owner, String attribute);
 
-        Map<String, Explainable> explainableRelations();
+        Map<String, Explainable> relations();
 
-        Map<String, Explainable> explainableAttributes();
+        Map<String, Explainable> attributes();
 
-        Map<Pair<String, String>, Explainable> explainableOwnerships();
+        Map<Pair<String, String>, Explainable> ownerships();
 
     }
 
--- a/api/answer/ConceptMap.java
+++ b/api/answer/ConceptMap.java
@@ -20,15 +20,11 @@
 package grakn.client.api.answer;
 
 import grakn.client.api.concept.Concept;
-import grakn.client.common.exception.GraknClientException;
 import grakn.common.collection.Pair;
 
 import javax.annotation.CheckReturnValue;
 import java.util.Collection;
 import java.util.Map;
-import java.util.Set;
-
-import static grakn.client.common.exception.ErrorMessage.Concept.NONEXISTENT_EXPLAINABLE_CONCEPT;
 
 public interface ConceptMap {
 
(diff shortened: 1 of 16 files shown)


api/query/QueryManager.java
```diff
@@ -72,6 +73,10 @@
 
     Stream<ConceptMap> update(GraqlUpdate query, GraknOptions options);
 
+    Stream<Explanation> explain(ConceptMap.Explainables.Explainable explainable);
```
alexjpwalker, line 76: `ConceptMap.Explainables.Explainable` feels pretty cumbersome.

concept/answer/ConceptMapImpl.java
```diff
@@ -83,4 +112,97 @@ public boolean equals(Object obj) {
...
+                    explainableOwnerships.equals(that.explainableOwnerships);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(explainableConcepts, explainableOwnerships);
+        }
+
+
+        static class ExplainableImpl implements Explainable {
```
alexjpwalker, line 170: `ExplainablesImpl.ExplainableImpl` feels really nasty - can we rename this class to 
```suggestion
        static class Explainable implements Explainables.Explainable {
```

logic/ExplanationImpl.java
```diff
@@ -0,0 +1,95 @@
...
+
+public class ExplanationImpl implements Explanation {
+
+    private final Rule rule;
+    private final Map<String, Set<String>> variableMapping;
+    private final ConceptMap thenAnswer;
+    private final ConceptMap whenAnswer;
+
+    private ExplanationImpl(Rule rule, Map<String, Set<String>> variableMapping, ConceptMap thenAnswer, ConceptMap whenAnswer) {
+        this.rule = rule; this.variableMapping = variableMapping;
```
alexjpwalker, line 43: Keep these on separate lines

api/answer/ConceptMap.java
```diff
@@ -35,4 +40,26 @@
...
+
+        Explainable ownership(String owner, String attribute);
+
+        Map<String, Explainable> explainableConcepts();
+
+        Map<Pair<String, String>, Explainable> explainableOwnerships();
+
+        interface Explainable {
+
+            String conjunction();
```
alexjpwalker, line 58: What is an Explainable's `conjunction`?

api/logic/Explanation.java
```diff
@@ -0,0 +1,37 @@
...
+
+public interface Explanation {
+
+    Rule rule();
+
+    ConceptMap thenAnswer();
+
+    ConceptMap whenAnswer();
+
+    Map<String, Set<String>> variableMapping();
```
alexjpwalker, line 35: What are `thenAnswer`, `whenAnswer` and `variableMapping`?

alexjpwalker (changes requested):

concept/answer/ConceptMapImpl.java
flyingsilverfin: shortened to `ConceptMapImpl.ExplainableImpl`

api/answer/ConceptMap.java
flyingsilverfin: It's the corresponding part of the user's original query (ie the sub-conjunction) that is being explained, mostly for the user to see if they are interactively querying this api

api/logic/Explanation.java
flyingsilverfin: So given a particular `Rule` that is being explained, these are the `ConceptMap` were substituted into the `Rule` when and then respectively, to reach the inference that answered your original answer

concept/answer/ConceptMapImpl.java
flyingsilverfin: i think this is ok

api/logic/Explanation.java
flyingsilverfin: oh and the `variableMapping` is there because your query isn't necesarily in the same variable space as the rule, so this provides that mapping

alexjpwalker (approved):

api/answer/ConceptMap.java
alexjpwalker: Do you think `subquery`, or `queryPart`, or `queryFragment` might be a clearer name? `conjunction` and `disjunction` are very logical terms.

flyingsilverfin: "Explainable Conjunction" is alright i think, the user will see its a subquery pretty quickly

alexjpwalker (approved):

test/integration/ClientQueryTest.java
```diff
@@ -454,17 +460,75 @@ public void parallelQueriesInTransactionTest() {
...
+                            "$x isa person, has name 'Zack'; " +
+                            "$y isa person, has name 'Yasmin'; " +
+                            "(husband: $x, wife: $y) isa marriage;"
+            ).asInsert();
+            tx.query().insert(data);
+            tx.commit();
+        }, WRITE);
+
+        localhostGraknTx(tx -> {
+            GraqlInsert data = Graql.parseQuery(
```
alexjpwalker, line 501: This is unused, right?

PR #308, opened 2021-03-22
Fix error running Graql Update and clean up Concept package

alexjpwalker:

We fixed a possible error when getting the results of a Graql Update query, and cleaned up various unnecessary or confusing methods in the Concept package.

- Fix possible error getting results of an Update query
- Fix inaccurate RequestBuilder method names
- Rename protoRoleTypes, transactionRPC
- Delete AttributeType.isKeyable

### DIFF
--- a/api/concept/type/AttributeType.java
+++ b/api/concept/type/AttributeType.java
@@ -38,9 +38,6 @@ public interface AttributeType extends ThingType {
         return ValueType.OBJECT;
     }
 
-    @CheckReturnValue
-    boolean isKeyable();
-
     @Override
     @CheckReturnValue
     default boolean isAttributeType() {
--- a/common/rpc/RequestBuilder.java
+++ b/common/rpc/RequestBuilder.java
@@ -665,23 +665,23 @@ public class RequestBuilder {
                 ));
             }
 
-            public static ConceptProto.Attribute.Value attributeValueBooleanReq(boolean value) {
+            public static ConceptProto.Attribute.Value protoBooleanAttributeValue(boolean value) {
                 return ConceptProto.Attribute.Value.newBuilder().setBoolean(value).build();
             }
 
-            public static ConceptProto.Attribute.Value attributeValueLongReq(long value) {
+            public static ConceptProto.Attribute.Value protoLongAttributeValue(long value) {
                 return ConceptProto.Attribute.Value.newBuilder().setLong(value).build();
             }
 
-            public static ConceptProto.Attribute.Value attributeValueDoubleReq(double value) {
+            public static ConceptProto.Attribute.Value protoDoubleAttributeValue(double value) {
                 return ConceptProto.Attribute.Value.newBuilder().setDouble(value).build();
             }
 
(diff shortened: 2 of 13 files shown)


api/concept/type/AttributeType.java
```diff
@@ -38,9 +38,6 @@ default ValueType getValueType() {
         return ValueType.OBJECT;
     }
 
-    @CheckReturnValue
-    boolean isKeyable();
```
alexjpwalker, line 42: This was unused and just calling `getValueType().isKeyable()`, so it was unnecessary.

common/rpc/RequestBuilder.java
```diff
@@ -665,23 +665,23 @@ static ByteString byteString(String iid) {
                 ));
             }
 
-            public static ConceptProto.Attribute.Value attributeValueBooleanReq(boolean value) {
```
alexjpwalker, line 668: These methods weren't actually constructing `Req` messages so we took Req out of the name.

concept/thing/RelationImpl.java
```diff
@@ -96,12 +96,12 @@ public RelationTypeImpl getType() {
 
         @Override
         public void addPlayer(RoleType roleType, Thing player) {
-            execute(addPlayerReq(getIID(), protoRoleTypes(roleType), protoThing(player.getIID())));
+            execute(addPlayerReq(getIID(), protoRoleType(roleType), protoThing(player.getIID())));
         }
 
         @Override
         public void removePlayer(RoleType roleType, Thing player) {
-            execute(removePlayerReq(getIID(), protoRoleTypes(roleType), protoThing(player.getIID())));
```
alexjpwalker, line 104: This method was taking in a single RoleType, but was confusingly named `protoRoleTypes` with an S

concept/type/AttributeTypeImpl.java
```diff
@@ -49,36 +49,31 @@
 
     private static final Label ROOT_LABEL = Label.of("attribute");
 
-    AttributeTypeImpl(java.lang.String label, boolean isRoot) {
```
alexjpwalker, line 52: Label was not consistently used across our Concept constructors, which made code harder to reuse.

query/QueryManagerImpl.java
```diff
@@ -138,7 +138,7 @@ public QueryManagerImpl(GraknTransaction.Extended transactionRPC) {
     @Override
     public Stream<ConceptMap> update(GraqlUpdate query, GraknOptions options) {
         return stream(updateReq(query.toString(), options.proto()))
-                .flatMap(rp -> rp.getInsertResPart().getAnswersList().stream())
```
alexjpwalker, line 141: This would have thrown an exception if anyone ever tried to get the answers of an Update query!

api/concept/type/AttributeType.java
flyingsilverfin: good catch!

query/QueryManagerImpl.java
flyingsilverfin: nice haha!

flyingsilverfin (approved):

PR #320, opened 2021-04-07
Encode request IDs as bytes and make isInferred a Local Thing method

alexjpwalker:

Previously the ID of a transaction request was stored as a string, which is an inefficient representation of a UUID. We've changed it to bytes.

Also, `isInferred` is now a Local method on `Thing`.

Store request IDs as bytes, not strings
Make isInferred a Local Thing method


alexjpwalker:
Do not merge until Grakn Core artifact is updated

haikalpribadi (approved):

PR #323, opened 2021-04-14
Fix possible NullPointer errors when connecting to Cluster

alexjpwalker:

We fixed a few possible sources of `NullPointerException` which would be thrown while trying to print other errors when constructing a Cluster client or during Cluster operations.

Fix Null errors when trying to print other errors


alexjpwalker:
Once merged, this update should be immediately propagated to `console`.

lolski (approved):

PR #333, opened 2021-05-17
Make TypeDB Cluster client require TypeDBCredential

lolski:

Make TypeDB Cluster client require `TypeDBCredential`.

- Added a new object `TypeDBCredential` for storing TLS encryption setting (and later user credential)
- Make `TypeDB.clusterClient` require `TypeDBCredential` to be supplied


api/TypeDBOptions.java
```diff
@@ -42,6 +42,8 @@
     private Integer sessionIdleTimeoutMillis = null;
     private Integer schemaLockAcquireTimeoutMillis = null;
 
+    private TypeDBOptions() {}
```
lolski, line 45: I found out a user may accidentally construct the object using the constructor. This shouldn't be allowed - users should construct the object using the static factory method `core()` or `cluster()`.

haikalpribadi (approved):

PR #344, opened 2021-07-29
Simplify the "database get" function of the TypeDB Cluster Client

lolski:

We've removed an unnecessary nested failover logic in order to make the logic easier to follow.

- Make `get(...)` call `db.contains` directly rather than the variant that is wrapped in the failover logic.

### DIFF
--- a/connection/cluster/ClusterDatabase.java
+++ b/connection/cluster/ClusterDatabase.java
@@ -52,10 +52,10 @@ class ClusterDatabase implements Database.Cluster {
         this.databases = new HashMap<>();
         this.replicas = new HashSet<>();
 
-        ClusterDatabaseManager clusterDbMgr = client.databases();
-        for (String address : clusterDbMgr.databaseMgrs().keySet()) {
-            TypeDBDatabaseManagerImpl coreDatabaseMgr = clusterDbMgr.databaseMgrs().get(address);
-            databases.put(address, new TypeDBDatabaseImpl(coreDatabaseMgr, database));
+        ClusterDatabaseManager clusterDBMgr = client.databases();
+        for (String address : clusterDBMgr.databaseMgrs().keySet()) {
+            TypeDBDatabaseManagerImpl databaseMgr = clusterDBMgr.databaseMgrs().get(address);
+            databases.put(address, new TypeDBDatabaseImpl(databaseMgr, database));
         }
     }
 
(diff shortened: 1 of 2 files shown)


connection/cluster/ClusterDatabase.java
```diff
@@ -52,10 +52,10 @@ private ClusterDatabase(String database, ClusterClient client) {
         this.databases = new HashMap<>();
         this.replicas = new HashSet<>();
 
-        ClusterDatabaseManager clusterDbMgr = client.databases();
-        for (String address : clusterDbMgr.databaseMgrs().keySet()) {
-            TypeDBDatabaseManagerImpl coreDatabaseMgr = clusterDbMgr.databaseMgrs().get(address);
-            databases.put(address, new TypeDBDatabaseImpl(coreDatabaseMgr, database));
+        ClusterDatabaseManager clusterDBMgr = client.databases();
+        for (String address : clusterDBMgr.databaseMgrs().keySet()) {
+            TypeDBDatabaseManagerImpl databaseMgr = clusterDBMgr.databaseMgrs().get(address);
```
lolski, line 57: rename `coreDbMgr` to just `dbMgr` here and in several other places, since it is not a Core-specific instance

connection/cluster/ClusterDatabaseManager.java
```diff
@@ -59,21 +59,21 @@ public ClusterDatabaseManager(ClusterClient client) {
 
     @Override
     public boolean contains(String name) {
-        return failsafeTask(name, (stub, coreDbMgr) -> coreDbMgr.contains(name));
+        return failsafeTask(name, (stub, dbMgr) -> dbMgr.contains(name));
```
lolski, line 62: rename `coreDbMgr` to just `dbMgr` here and in several other places, since it is not a Core-specific instance

connection/cluster/ClusterDatabaseManager.java
```diff
@@ -59,21 +59,21 @@ public ClusterDatabaseManager(ClusterClient client) {
...
             return null;
         });
     }
 
     @Override
     public Database.Cluster get(String name) {
-        return failsafeTask(name, (stub, coreDbMgr) -> {
-            if (contains(name)) {
+        return failsafeTask(name, (stub, dbMgr) -> {
+            if (dbMgr.contains(name)) {
```
lolski, line 76: call `dbMgr.contains` directly. The `get` function is already wrapped inside `failsafeTask`, so there's no need to call `contains(...)` which is also wrapped in `failsafeTask`.

alexjpwalker (approved):

PR #346, opened 2021-08-18
Make FailsafeTask an inner class of ClusterClient

lolski:

We've made `FailsafeTask` instances an inner class of `ClusterClient`. `FailSafeTask` makes use and updates the internal state of `ClusterClient` in the event of failover, and therefore are tightly coupled by design. 

- Make `FailsafeTask` internal to `ClusterClient`
- Adds `FailsafeTaskParameter` to encapsulate parameters that are needed for the execution of the task, namely: `ClusterServerClient`, `ClusterServerStub`, and `Replica`


connection/cluster/ClusterClient.java
```diff
@@ -148,9 +149,13 @@ ClusterSession run(ClusterDatabase.Replica replica) {
...
-    // TODO: this is not good - we should not pass an internal object to be modified outside of this class
-    ConcurrentMap<String, ClusterDatabase> databaseByName() {
-        return clusterDatabases;
+    @Nullable
+    ClusterDatabase databaseGet(String name) {
+        return clusterDatabases.get(name);
+    }
+
+    void databasePut(String name, ClusterDatabase database) {
+        clusterDatabases.put(name, database);
```
alexjpwalker, line 158: While this is marginally better than before because of less access to the `HashMap`, I think the overall problem remains - we are still building up the definition of what a `ClusterClient` is by adding stuff to it from outside of the class.

We could solve this issue by adding a callback function to the invocations of `ClusterClient`, say, `onFetchDatabaseReplicas`, but this doesn't feel like the right solution because the actual value of the callback would always be the same - we always want to call `client.databasePut`.

Because of the strong coupling between `ClusterClient` and `FailsafeTask`, and also the awkward positioning of `FailsafeTask` in the `cluster` package (note: every other class in this package starts with the prefix `Cluster`!) I'm starting to feel like `FailsafeTask` should actually be an inner class of `ClusterClient`.

That way, we don't need to make this change at all, because `databaseByName` can be dissolved, as `clusterDatabases` will now be used only within the class itself.

connection/cluster/ClusterDatabaseManager.java
```diff
@@ -45,7 +45,7 @@
 
 public class ClusterDatabaseManager implements DatabaseManager.Cluster {
 
-    private static final Logger LOG = LoggerFactory.getLogger(FailsafeTask.class);
+    private static final Logger LOG = LoggerFactory.getLogger(ClusterClient.FailsafeTask.class);
```
alexjpwalker, line 48: Hmm looks like this was always wrong
```suggestion
    private static final Logger LOG = LoggerFactory.getLogger(ClusterDatabaseManager.class);
```

lolski: Oops, fixed in https://github.com/vaticle/typedb-client-java/pull/346/commits/e4a78991a2e7d340210206d22f8c8770d23c2ac1

alexjpwalker (approved):
Yeah I think this is a lot better isn't it? `FailsafeTask` always felt a bit loose and unspecific. `ClusterClient.FailsafeTask` is a much better name and completely conveys what the thing actually is.

lolski:
@alexjpwalker yup agreed, your suggestion made a lot of sense.

PR #347, opened 2021-08-20
ClusterServerClient owns ClusterServerStub

lolski:

We've made TypeDB Cluster client's `ClusterServerClient` own `ClusterServerStub` instance. This is consistent with TypeDB Client implementation where `TypeDBClientImpl` owns `TypeDBStub`.

Move `ClusterServerStub` instance to be inside `ClusterServerClient`

### DIFF
--- a/connection/cluster/ClusterClient.java
+++ b/connection/cluster/ClusterClient.java
@@ -157,23 +157,23 @@ public class ClusterClient implements TypeDBClient.Cluster {
 
     <RESULT> FailsafeTask<RESULT> createFailsafeTask(
             String database,
-            Function<FailsafeTaskParameter, RESULT> run) {
+            Function<FailsafeTaskParams, RESULT> run) {
         return createFailsafeTask(database, run, run);
     }
 
     <RESULT> FailsafeTask<RESULT> createFailsafeTask(
             String database,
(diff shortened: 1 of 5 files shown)


connection/cluster/ClusterClient.java
```diff
@@ -330,23 +321,17 @@ private TypeDBClientException clusterNotAvailableException() {
     static class FailsafeTaskParameter {
 
         private final ClusterServerClient client;
-        private final ClusterServerStub stub;
         private final ClusterDatabase.Replica replica;
 
-        public FailsafeTaskParameter(ClusterServerClient client, ClusterServerStub stub, ClusterDatabase.Replica replica) {
+        public FailsafeTaskParameter(ClusterServerClient client, ClusterDatabase.Replica replica) {
```
alexjpwalker, line 326: We should rename this to `FailsafeTaskParams`, it encapsulates multiple parameters

lolski: Right. Fixed in https://github.com/vaticle/typedb-client-java/pull/347/commits/ce28757ef2868534e6fda703990f118cf0a3d531.

alexjpwalker (approved):

PR #349, opened 2021-08-23
Describe how to disable debug logging in README

alexjpwalker:

A common complaint when using `typedb-client-java` is the large number of Netty and gRPC logs. They are actually easy to disable, but only if you know how. So we added brief instructions to the README.

Describe how to disable debug logging in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Further documentation: https://docs.vaticle.com/docs/client-api/java
 
 **Q:** I see a large number of Netty and gRPC log messages. How can I disable them?
 
-**A:** 
+**A:** Create a Logback configuration file and set the minimum log level to ERROR. You can do so with the following steps:
 1. Create a file in your `resources` path (`src/main/resources` by default in a Maven project) named `logback.xml`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,3 +61,25 @@ Further documentation: https://docs.vaticle.com/docs/client-api/java
    bazel-bin/com.vaticle.typedb:api.jar
    bazel-bin/pom.xml
    ```
+
+## FAQs
+
+**Q:** I see a large number of Netty and gRPC log messages. How can I disable them?
+
+**A:** 
+1. Create a file in your `resources` path (`src/main/resources` by default in a Maven project) named `logback.xml`.
```
lolski, line 70:
```
**A:**

Make sure that Logback is configured properly. You can do so with the following steps:

1. ...
2. ...
```

lolski: Also, how can we communicate it in the [client docs](https://docs.vaticle.com/docs/client-api/java) as well?

lolski (commented):
Please see my comments.

PR #350, opened 2021-09-02
User token mechanism for faster authentication

lolski:

Authenticating a user by verifying their password introduces a significant overhead, which comes from the fact that the password is hashed with cryptographic hash function.  In effect, opening a new session or transaction becomes much slower.

We've now improved the speed of user-authentication by introducing a mechanism for verifying user credential with cheap "user token" rather than password.

Additionally we've also performed some refactors that has greatly increased the clarity `CoreClient`, `ClusterClient` and the related classes.

- initialising the client renews the token
- resilientCall now renews the token, if the token is no longer valid

### Refactors

Improve the abstraction of the classes CoreClient, ClusterClient, ClusterServerClient:

1. Make `stub` overridable and can be instantiated with `TypeDBStub` or `ClusterServerStub` depending on the client type.
2. Remove connection factory classes.
3. Remove an unneeded `clusterServerStub` in `ClusterServerClient`.


connection/cluster/ClusterServerClient.java
```diff
@@ -22,22 +22,65 @@
...
+                .build();
+    }
+
+    private ManagedChannel tlsChannel(String address, TypeDBCredential credential) {
+        try {
+            SslContext sslContext;
+            if (credential.tlsRootCA().isPresent()) {
+                sslContext = GrpcSslContexts.forClient().trustManager(credential.tlsRootCA().get().toFile()).build();
+            } else {
+                sslContext = GrpcSslContexts.forClient().build();
```
alexjpwalker, line 79: How does this call actually work without a specified `tlsRootCA`? I thought it was necessary to specify one? Or is there some system / application default

connection/cluster/ClusterServerStub.java
```diff
@@ -21,36 +21,58 @@
...
+        this.channel = channel;
+        CredentialEmbedder credentialEmbedder = new CredentialEmbedder();
+        this.blockingStub = TypeDBGrpc.newBlockingStub(channel).withCallCredentials(credentialEmbedder);
+        this.asyncStub = TypeDBGrpc.newStub(channel).withCallCredentials(credentialEmbedder);
+        this.clusterBlockingStub = TypeDBClusterGrpc.newBlockingStub(channel).withCallCredentials(credentialEmbedder);
+        try {
+            ClusterUserTokenProto.ClusterUserToken.Renew.Res res = clusterBlockingStub.userTokenRenew(renewReq(this.credential.username()));
+            token = res.getToken();
+        } catch (StatusRuntimeException e) {
+            // ignore UNAVAILABLE exception
```
alexjpwalker, line 67: Why would we want to ignore this exception? Surely a gRPC error will cause problems regardless of whether it is a connection error or not?

connection/cluster/ClusterServerStub.java
```diff
@@ -85,30 +107,62 @@ public static ClusterServerStub create(String username, String password, Managed
...
+                try {
+                    return function.get();
+                } catch (StatusRuntimeException e3) {
+                    throw TypeDBClientException.of(e3);
+                }
+            } else throw e2;
         }
+    }
+
+    private class CredentialEmbedder extends CallCredentials {
```
alexjpwalker, line 145: How about `TypeDBCallCredentials`? 
```suggestion
    private class TypeDBCallCredentials extends CallCredentials {
```

dependencies/vaticle/repositories.bzl
```diff
@@ -45,8 +45,8 @@ def vaticle_dependencies():
 def vaticle_typedb_protocol():
     git_repository(
         name = "vaticle_typedb_protocol",
-        remote = "https://github.com/vaticle/typedb-protocol",
-        tag = "2.3.0", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
+        remote = "https://github.com/lolski/typedb-protocol",
+        commit = "54fc1734f613b5a5360571896ca65fd00be672a4", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
```
alexjpwalker, line 49: Update to vaticle when protocol PR is merged

alexjpwalker (changes requested):
lalala

common/exception/TypeDBClientException.java
```diff
@@ -43,15 +47,31 @@ public TypeDBClientException(String message, Throwable cause) {
...
+                description != null &&
+                description.contains(CLUSTER_REPLICA_NOT_PRIMARY_ERROR_CODE)
+        ) {
             return new TypeDBClientException(ErrorMessage.Client.CLUSTER_REPLICA_NOT_PRIMARY);
+        } else if (
+                code == Status.Code.UNAUTHENTICATED &&
+                description != null &&
+                description.contains(CLUSTER_TOKEN_CREDENTIAL_INVALID_ERROR_CODE)
+        ) {
+            return new TypeDBClientException(ErrorMessage.Client.CLUSTER_TOKEN_CREDENTIAL_INVALID);
```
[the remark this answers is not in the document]
lolski: I've extract away the condition in order to make them look better: 7ebf4a8b38787cc0204be0b18f866d1a4964de8a

common/rpc/RequestBuilder.java
```diff
@@ -116,11 +118,11 @@ public static ByteString UUIDAsByteString(UUID uuid) {
 
         public static class UserManager {
             public static ClusterUserProto.ClusterUserManager.Contains.Req containsReq(String name) {
-                return ClusterUserProto.ClusterUserManager.Contains.Req.newBuilder().setName(name).build();
+                return ClusterUserProto.ClusterUserManager.Contains.Req.newBuilder().setUsername(name).build();
             }
 
             public static ClusterUserProto.ClusterUserManager.Create.Req createReq(String name, String password) {
```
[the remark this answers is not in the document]
lolski: Yes sounds good. Overall, the term "username" is better than "name" though right? `username` is the terminology that you use to refer to a user's "unique identifier", whereas `name` is more commonly used for referring to their birth name.

common/rpc/RequestBuilder.java
```diff
@@ -107,6 +103,12 @@ public static ByteString UUIDAsByteString(UUID uuid) {
 
     public static class Cluster {
 
+        public static class ClusterUserToken {
```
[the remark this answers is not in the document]
lolski: Fixed in 8135eee8cff67f72a0a711e3c71ff7540b34da7a.

connection/cluster/ClusterServerClient.java
lolski: `tlsRootCA` needs to only be specified if you're creating your own custom TLS certificate, eg., for developing in your local machine.

If you're getting your certificate from major third party vendors such as CloudFlare, you won't need to do it because I think their root CA already comes pre-installed with your OS.

connection/cluster/ClusterServerStub.java
lolski: I went for a name that's more self-explanatory since the terminology "call credential" is just not a very clear name.

The name "credential embedder" clearly indicates that the class is responsible for embedding the credential in the request.

What do you think?

connection/cluster/ClusterServerStub.java
lolski: What the stub is doing here is trying to **pre-acquire** a token of the user to a server such that by the time the user wants to do something, the token is already available.

If the request fail with `UNAVAILABLE`, it indicates a network failure or server crash, and as a result the token from that particular server will not be pre-acquired. We can ignore it since failure is a fact of life in TypeDB Cluster. Failing to pre-acquire a token just means the client will have to reattempt it later.

common/rpc/RequestBuilder.java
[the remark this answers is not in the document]
alexjpwalker: Yes. I agree, username is a less ambiguous term.

connection/cluster/ClusterServerStub.java
alexjpwalker: Hmm, I'm not sure that credential embedder is the best term here. The key thing is that it extends `CallCredentials`, in other words it _is a_ `CallCredentials`. Moreover, the usage of it is:
```java
this.blockingStub = TypeDBGrpc.newBlockingStub(channel).withCallCredentials(credentialEmbedder);
```
Lastly we should ask: is gRPC's name misleading? Here is the javadoc for `CallCredentials`:

https://grpc.github.io/grpc-java/javadoc/io/grpc/CallCredentials.html

The purpose of the class is to "carry credential data that will be propagated to the server via request data".

It is embedded in the request, yes, but that describes what it is used for, not what it is

alexjpwalker (approved):

Issue #351, opened 2021-09-06
IllegalMonitorStateException in LinkedBlockingQueue.take when doing many concurrent Concept API calls


alexjpwalker:
## Description

When doing many concurrent Concept API calls, an IllegalMonitorStateException is thrown.

## Environment

1. OS (where TypeDB server runs): Mac OS 10
2. TypeDB version (and platform): TypeDB 2.3

## Reproducible Steps

Steps to create the smallest reproducible scenario:

1. Import Grabl dataset into TypeDB
2. Use the following test case:
```
@Test
    public void grablTest() throws Throwable {
        TypeDBSession session = null;
        try {
            session = typedbClient.session("grabl", DATA);
            TypeDBTransaction tx = session.transaction(READ);
            final Stream<ConceptMap> answerStream = tx.query().match("match $x isa relation;");
            Collection<CompletableFuture<Object>> futures = new ArrayList<>();
            answerStream.forEach(conceptMap -> {
                conceptMap.concepts().stream().filter(Concept::isRelation).map(Concept::asRelation).forEach(relation -> {
                    futures.add(CompletableFuture.supplyAsync(() -> relation.asRemote(tx).getPlayers().collect(toList())));
                });
            });
            CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join();
        } catch (Exception e) {
            if (session != null) session.close();
            if (e instanceof CompletionException) throw e.getCause();
            else throw e;
        }
    }
```

## Expected Output

The calls should succeed.

## Actual Output

An IllegalMonitorStateException is thrown.
 
## Additional information

This bug does not occur in Java 11, but does occur in Java 14, 15 and 16. However, it is fixed in Java 18 (Early Access at time of writing)

This bug can be worked around by using a `LinkedTransferQueue` instead.

PR #352, opened 2021-09-06
Fix error doing concurrent reads on Java 14-16

alexjpwalker:

Java 14, 15 and 16 all have a bug where `LinkedBlockingQueue.take` can throw an `IllegalMonitorStateException` if you concurrently take from many `LinkedBlockingQueue` (see: #351 )

Although this bug is fixed in Java 18, we need an effective workaround in order to support Java 14-16, so we have changed `ResponseCollector` to be backed by a `LinkedTransferQueue` instead, inflicting a small performance penalty.

Fix error doing concurrent reads on Java 14-16

### DIFF
--- a/stream/ResponseCollector.java
+++ b/stream/ResponseCollector.java
@@ -31,7 +31,6 @@ import java.util.UUID;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.LinkedTransferQueue;
 
 import static com.vaticle.typedb.client.common.exception.ErrorMessage.Client.TRANSACTION_CLOSED;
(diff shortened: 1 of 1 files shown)


stream/ResponseCollector.java
```diff
@@ -64,7 +65,8 @@ public synchronized void close(@Nullable StatusRuntimeException error) {
         private final BlockingQueue<Either<Response<R>, Done>> responseQueue;
 
         Queue() {
-            responseQueue = new LinkedBlockingQueue<>();
+            // TODO: We should be using LinkedBlockingQueue here, however there is a bug in Java 14-16: see #351
```
lolski, line 68: Should we make it a bit more concise:

```
// TODO: switch LinkedTransferQueue to LinkedBlockingQueue once issue #351 is fixed
```

alexjpwalker: I'm actually considering getting rid of the "TODO" entirely @lolski . Even if _we_ upgrade to Java 18 in the future, there's no guarantee that our users will. Therefore I think it's safer to keep `LinkedTransferQueue` and allow ourselves to support Java 14-16 for the foreseeable future (at least 2 years)

lolski: I think we can expect the bug to be fixed on a newer Java 14-16. You know how Java 8 gets tons of minor updates? I would expect the same for Java 14-16.

alexjpwalker: Alright, changed

lolski (approved):

PR #358, opened 2021-10-04
Refactor client creation methods

lolski:

We've refactored client creation methods, minimising unnecessary boilerplate.

- Remove `create` factory methods of `CoreClient`, `ClusterClient`, and `ClusterServerClient`.
- Reordered methods used only in the constructor to be next to the constructor following our convention.

### DIFF
--- a/connection/cluster/ClusterServerStub.java
+++ b/connection/cluster/ClusterServerStub.java
@@ -70,8 +70,29 @@ public class ClusterServerStub extends TypeDBStub {
         }
     }
 
-    public static ClusterServerStub create(TypeDBCredential credential, ManagedChannel channel) {
-        return new ClusterServerStub(channel, credential);
+    private CallCredentials createCallCredentials() {
+        return new CallCredentials() {
+            private final Metadata.Key<String> TOKEN_FIELD = Metadata.Key.of("token", ASCII_STRING_MARSHALLER);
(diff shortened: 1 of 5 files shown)


connection/cluster/ClusterServerStub.java
```diff
@@ -70,8 +70,29 @@
...
+                        headers.put(TOKEN_FIELD, token);
+                    } else {
+                        headers.put(PASSWORD_FIELD, credential.password());
+                    }
+                    applier.apply(headers);
+                });
+            }
+
+            @Override
+            public void thisUsesUnstableApi() { }
```
alexjpwalker, line 94: I know this wasn't added in this commit, but do we know what this is all about? `@Override public void thisUsesUnstableApi() { }`??

lolski: It is something that we are forced to implement if we're to implement the `CallCredentials` interface.

I could not find any documentation about why we need to to implement the method,  but according to their Javadoc, I assume it's gRPC's way of forcing us to acknowledge that by using `CallCredentials` we're effectively using an API considered unstable.

Upon researching, I found an established library, Apache Arrow, which also needs to [override it](https://github.com/apache/arrow/blob/a8eb73699b32ae36b2dd218e3eb969ec2cebd449/java/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/CallCredentialAdapter.java#L49-L52).

PR #364, opened 2021-12-14
Add and test transaction timeouts

flyingsilverfin:

We implement options for transaction timeout configuration (https://github.com/vaticle/typedb/issues/6487). We also add BDD implementation for new steps required to test transaction timeout and configuring transaction options.

* Add option for transaction timeout (default: 5 mins)
* fix build by adding missing maven dependencies from common
* fix a bug where errors submitted to the response queue were ignored (cleared immediately) - these are now drained from the response queue 
* add missing BDD steps for session and transaction option configuration

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -126,6 +126,7 @@ vaticle_typedb_artifact()
 vaticle_typedb_cluster_artifact()
 
 # Load maven
+load("@vaticle_typedb_common//dependencies/maven:artifacts.bzl", vaticle_typedb_common_artifacts = "artifacts")
 load("@vaticle_typeql_lang_java//dependencies/maven:artifacts.bzl", vaticle_typeql_lang_java_artifacts = "artifacts")
 load("@vaticle_factory_tracing//dependencies/maven:artifacts.bzl", vaticle_factory_tracing_artifacts = "artifacts")
 load("//dependencies/maven:artifacts.bzl", vaticle_typedb_client_java_artifacts = "artifacts", vaticle_typedb_client_java_overrides = "overrides")
@@ -137,6 +138,7 @@ load("//dependencies/maven:artifacts.bzl", vaticle_typedb_client_java_artifacts 
 load("@vaticle_dependencies//library/maven:rules.bzl", "maven")
 maven(
     vaticle_factory_tracing_artifacts +
+    vaticle_typedb_common_artifacts +
     vaticle_typeql_lang_java_artifacts +
     vaticle_dependencies_tool_maven_artifacts +
     vaticle_typedb_client_java_artifacts +
--- a/api/TypeDBOptions.java
+++ b/api/TypeDBOptions.java
@@ -40,6 +40,7 @@ public class TypeDBOptions {
     private Boolean prefetch = null;
     private Integer prefetchSize = null;
     private Integer sessionIdleTimeoutMillis = null;
+    private Integer transactionTimeoutMillis = null;
     private Integer schemaLockAcquireTimeoutMillis = null;
 
     private TypeDBOptions() {
@@ -141,6 +142,18 @@ public class TypeDBOptions {
         return Optional.ofNullable(schemaLockAcquireTimeoutMillis);
     }
 
+    public TypeDBOptions transactionTimeoutMillis(int transactionTimeoutMillis) {
+        if (transactionTimeoutMillis < 1) {
+            throw new TypeDBClientException(NEGATIVE_VALUE_NOT_ALLOWED, transactionTimeoutMillis);
+        }
+        this.transactionTimeoutMillis = transactionTimeoutMillis;
+        return this;
+    }
+
+    public Optional<Integer> transactionTimeoutMillis() {
+        return Optional.ofNullable(transactionTimeoutMillis);
+    }
+
     public TypeDBOptions schemaLockAcquireTimeoutMillis(int schemaLockAcquireTimeoutMillis) {
         if (schemaLockAcquireTimeoutMillis < 1) {
             throw new TypeDBClientException(NEGATIVE_VALUE_NOT_ALLOWED, schemaLockAcquireTimeoutMillis);
@@ -164,6 +177,7 @@ public class TypeDBOptions {
         prefetchSize().ifPresent(builder::setPrefetchSize);
         prefetch().ifPresent(builder::setPrefetch);
         sessionIdleTimeoutMillis().ifPresent(builder::setSessionIdleTimeoutMillis);
+        transactionTimeoutMillis().ifPresent(builder::setTransactionTimeoutMillis);
         schemaLockAcquireTimeoutMillis().ifPresent(builder::setSchemaLockAcquireTimeoutMillis);
         if (isCluster()) asCluster().readAnyReplica().ifPresent(builder::setReadAnyReplica);
 
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -34,30 +34,32 @@ public abstract class ErrorMessage extends com.vaticle.typedb.common.exception.E
                 new Client(2, "The session has been closed and no further operation is allowed.");
         public static final Client TRANSACTION_CLOSED =
                 new Client(3, "The transaction has been closed and no further operation is allowed.");
+        public static final Client TRANSACTION_CLOSED_WITH_ERRORS =
+                new Client(4, "The transaction has been closed and no further operation is allowed because of error(s): '%s'.");
         public static final Client UNABLE_TO_CONNECT =
-                new Client(4, "Unable to connect to TypeDB server.");
+                new Client(5, "Unable to connect to TypeDB server.");
         public static final Client NEGATIVE_VALUE_NOT_ALLOWED =
-                new Client(5, "Value cannot be less than 1, was: '%d'.");
+                new Client(6, "Value cannot be less than 1, was: '%d'.");
         public static final Client MISSING_DB_NAME =
-                new Client(6, "Database name cannot be null.");
+                new Client(7, "Database name cannot be null.");
         public static final Client DB_DOES_NOT_EXIST =
-                new Client(7, "The database '%s' does not exist.");
+                new Client(8, "The database '%s' does not exist.");
         public static final Client MISSING_RESPONSE =
-                new Client(8, "Unexpected empty response for request ID '%s'.");
+                new Client(9, "Unexpected empty response for request ID '%s'.");
         public static final Client UNKNOWN_REQUEST_ID =
-                new Client(9, "Received a response with unknown request id '%s'.");
+                new Client(10, "Received a response with unknown request id '%s'.");
         public static final Client CLUSTER_NO_PRIMARY_REPLICA_YET =
-                new Client(10, "No replica has been marked as the primary replica for latest known term '%d'.");
+                new Client(11, "No replica has been marked as the primary replica for latest known term '%d'.");
         public static final Client CLUSTER_UNABLE_TO_CONNECT =
-                new Client(11, "Unable to connect to TypeDB Cluster. Attempted connecting to the cluster members, but none are available: '%s'.");
+                new Client(12, "Unable to connect to TypeDB Cluster. Attempted connecting to the cluster members, but none are available: '%s'.");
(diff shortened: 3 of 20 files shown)


WORKSPACE
```diff
@@ -126,6 +126,7 @@ vaticle_typedb_artifact()
 vaticle_typedb_cluster_artifact()
 
 # Load maven
+load("@vaticle_typedb_common//dependencies/maven:artifacts.bzl", vaticle_typedb_common_artifacts = "artifacts")
```
flyingsilverfin, line 129: we need to load `common` maven requirements as well

api/TypeDBOptions.java
```diff
@@ -40,6 +40,7 @@
     private Boolean prefetch = null;
     private Integer prefetchSize = null;
     private Integer sessionIdleTimeoutMillis = null;
+    private Integer transactionTimeoutMillis = null;
```
flyingsilverfin, line 43: new transaction option: `transactionTimeoutMillis`

api/TypeDBOptions.java
```diff
@@ -164,6 +177,7 @@ public Cluster asCluster() {
         prefetchSize().ifPresent(builder::setPrefetchSize);
         prefetch().ifPresent(builder::setPrefetch);
         sessionIdleTimeoutMillis().ifPresent(builder::setSessionIdleTimeoutMillis);
+        transactionTimeoutMillis().ifPresent(builder::setTransactionTimeoutMillis);
```
flyingsilverfin, line 180: set it in grpc message

common/exception/ErrorMessage.java
```diff
@@ -34,30 +34,32 @@ private ErrorMessage(String codePrefix, int codeNumber, String messagePrefix, St
                 new Client(2, "The session has been closed and no further operation is allowed.");
         public static final Client TRANSACTION_CLOSED =
                 new Client(3, "The transaction has been closed and no further operation is allowed.");
+        public static final Client TRANSACTION_CLOSED_WITH_ERRORS =
+                new Client(4, "The transaction has been closed and no further operation is allowed because of error(s): '%s'.");
```
flyingsilverfin, line 38: we now report transactions that were closed with errors separately

connection/TypeDBTransactionImpl.java
```diff
@@ -109,17 +110,23 @@ private Res execute(Req.Builder request, boolean batch) {
...
-        if (!isOpen()) throw new TypeDBClientException(TRANSACTION_CLOSED);
+        if (!isOpen()) drainAndThrow();
         return bidirectionalStream.stream(request);
     }
 
+    private void drainAndThrow() {
+        if (bidirectionalStream.hasErrors()) {
+            throw new TypeDBClientException(TRANSACTION_CLOSED_WITH_ERRORS, bidirectionalStream.drainErrors());
+        } else throw new TypeDBClientException(TRANSACTION_CLOSED);
+    }
```
flyingsilverfin, line 128: drain the errors from the queue and throw them all

stream/ResponseCollector.java
```diff
@@ -56,7 +58,17 @@ public ResponseCollector() {
...
+
+    public boolean hasErrors() {
+        return collectors.values().stream().anyMatch(collector -> collector.responseQueue.stream()
+                .anyMatch(msg -> msg.isSecond() && msg.second().error().isPresent()));
+    }
+
+    public List<StatusRuntimeException> errors() {
+        List<StatusRuntimeException> errors = new ArrayList<>();
+        collectors.values().forEach(collectors -> errors.addAll(collectors.drainErrors()));
+        return errors;
```
flyingsilverfin, line 71: collect all errors from all collectors

test/behaviour/connection/ConnectionStepsBase.java
```diff
@@ -51,7 +56,14 @@
...
     public static Map<CompletableFuture<TypeDBSession>, List<CompletableFuture<TypeDBTransaction>>> sessionsParallelToTransactionsParallel = new HashMap<>();
-    private static boolean isBeforeAllRan = false;
+    public static TypeDBOptions sessionOptions;
+    public static TypeDBOptions transactionOptions;
+    static boolean isBeforeAllRan = false;
+
+    public static final Map<String, BiConsumer<TypeDBOptions, Integer>> optionSetters = map(
+            pair("session-idle-timeout-millis", TypeDBOptions::sessionIdleTimeoutMillis),
+            pair("transaction-timeout-millis", TypeDBOptions::transactionTimeoutMillis)
+    );
```
flyingsilverfin, line 66: we set up only the initial options we need to configure

test/behaviour/connection/ConnectionStepsBase.java
```diff
@@ -72,6 +84,8 @@ void before() {
         assertNotNull(address);
         client = createTypeDBClient(address);
         client.databases().all().forEach(Database::delete);
+        sessionOptions = createOptions().infer(true);
+        transactionOptions = createOptions().infer(true);
```
flyingsilverfin, line 88: we always turn inference on - this was extracted from lower level implementation where we opened the transaction

test/behaviour/connection/transaction/TransactionTest.java
```diff
@@ -32,7 +32,7 @@
         plugin = "pretty",
         glue = "com.vaticle.typedb.client.test.behaviour",
         features = "external/vaticle_typedb_behaviour/connection/transaction.feature",
-        tags = "not @ignore and not @ignore-typedb-client-java and not @ignore-typedb"
```
flyingsilverfin, line 35: `@ignore-typedb` should only be for things ignored within TypeDB core (and maybe cluster?)

WORKSPACE
alexjpwalker: why?

connection/TypeDBTransactionImpl.java
alexjpwalker: I don't think this method name is intuitive - how about `throwTransactionClosedError()`? It's scoped to the class, and when looking at the class as a whole, that name makes the purpose clearer.

Also, `hasErrors` feels unnecessary here (looking at its implementation, too, it seems largely copied from/to `drainErrors`); I think we can delete `hasErrors` and write
```java
List<StatusRuntimeException> errors = bidirectionalStream.drainErrors();
if (errors.isEmpty()) throw new TypeDBClientException(TRANSACTION_CLOSED);
else throw new TypeDBClientException(TRANSACTION_CLOSED_WITH_ERRORS, errors);
```

dependencies/maven/artifacts.snapshot
```diff
@@ -137,12 +143,14 @@
 @maven//:org_objenesis_objenesis
 @maven//:org_objenesis_objenesis_2_5
 @maven//:org_slf4j_jcl_over_slf4j
-@maven//:org_slf4j_jcl_over_slf4j_1_7_20
+@maven//:org_slf4j_jcl_over_slf4j_1_7_32
```
alexjpwalker, line 146: This is to address vulnerability CVE-2021-44228 that was recently exposed in Log4J.

test/behaviour/connection/transaction/TransactionTest.java
alexjpwalker: Yep - this was probably added to disable a particular failing test a while back, and should since have been removed

test/behaviour/connection/ConnectionStepsCluster.java
```diff
@@ -45,6 +47,7 @@ void beforeAll() {
         }
         server.start();
         TypeDBSingleton.setTypeDBRunner(server);
+        ConnectionStepsBase.isBeforeAllRan = true;
```
alexjpwalker, line 50: This isn't necessary, right? Since the implementation of `before` in `ConnectionStepsBase` already sets `isBeforeAllRan` to `true`,

(having said that, `isBeforeAllRan` being set in a method that isn't `beforeAll` feels pretty dodgy, so perhaps there is a better solution here)

test/behaviour/connection/ConnectionStepsCore.java
```diff
@@ -44,6 +46,7 @@ void beforeAll() {
         }
         server.start();
         TypeDBSingleton.setTypeDBRunner(server);
+        ConnectionStepsBase.isBeforeAllRan = true;
```
alexjpwalker, line 49: see comment on `isBeforeAllRan` in `ConnectionStepsCluster`

test/behaviour/connection/ConnectionStepsBase.java
```diff
@@ -114,4 +131,8 @@ void connection_does_not_have_any_database() {
         assertNotNull(client);
         assertTrue(client.isOpen());
     }
+
+    void wait_seconds(int seconds) throws InterruptedException {
```
alexjpwalker, line 135: Feels like this should be in some kind of `UtilSteps`.. but then it'd be all alone, so ¯\\\_(ツ)_/¯

test/behaviour/connection/session/SessionSteps.java
```diff
@@ -152,4 +155,12 @@ public void sessions_in_parallel_have_databases(List<String> names) {
 
         CompletableFuture.allOf(assertions).join();
     }
+
+    @Given("set session option {word} to: {int}")
+    public void set_session_option_to(String option, int value) {
```
alexjpwalker, line 160: We have options that are not `int`-valued. Even though we aren't testing them yet, I imagine we'll do so soon, so we should future-proof this step by renaming it to
```java
@Given("set session option {word} to int: {int}")
```

test/behaviour/connection/transaction/TransactionSteps.java
```diff
@@ -262,6 +262,17 @@ private void for_each_session_in_parallel_transactions_in_parallel_are(Consumer<
         CompletableFuture.allOf(assertions.toArray(new CompletableFuture[0])).join();
     }
 
+    // ===================================//
+    // transaction configuration          //
+    // ===================================//
+
+    @Given("set transaction option {word} to: {int}")
```
alexjpwalker, line 269: same comment as the session equivalent above

alexjpwalker (changes requested):

WORKSPACE
flyingsilverfin: because we depend on all of common, and common now has a sub-library (without its own package) that requires maven dependencies!

connection/TypeDBTransactionImpl.java
flyingsilverfin: good ideas

test/behaviour/connection/ConnectionStepsCluster.java
flyingsilverfin: cleaned it up

test/behaviour/connection/ConnectionStepsBase.java
flyingsilverfin: Actually let me do that,  it'll probably grow

test/behaviour/connection/ConnectionStepsCluster.java
flyingsilverfin: Urgh actually im going to revert this, and when we upgrade Cucumber we can use a new version that has `@BeforeAll` and `@AfterAll`

test/behaviour/connection/session/SessionSteps.java
flyingsilverfin: Implemented this by always using strings and parsing them in the option setters

stream/BidirectionalStream.java
```diff
@@ -116,6 +118,16 @@ private void close(@Nullable StatusRuntimeException error) {
         }
     }
 
+    public boolean hasErrors() {
```
alexjpwalker, line 121: this method should now be unused & redundant

stream/ResponseCollector.java
```diff
@@ -56,7 +58,17 @@ public ResponseCollector() {
 
     public synchronized void close(@Nullable StatusRuntimeException error) {
         collectors.values().forEach(collector -> collector.close(error));
-        collectors.clear();
+    }
+
+    public boolean hasErrors() {
```
alexjpwalker, line 63: same as the other `hasErrors` above

alexjpwalker (changes requested):

alexjpwalker (approved):

PR #366, opened 2021-12-20
Upgrade grpc to 1.43.0 and netty to 4.1.61.Final

flyingsilverfin:

We silence false-positive grpc stream errors by upgrading grpc and netty. This propagages work started by  https://github.com/vaticle/dependencies/pull/339.

* Upgrade grpc to `1.43.0`, and use its transitive requirement netty `4.1.61.Final`
* Remove maven overrides pinning grpc to 4.1.38.Final after confirming it does not cause a performance regression (with a quick test on typedb-benchmark), and that Factory can use the new grpc/netty versions without issue.
* update TypeDB Core artifact


lolski:
Note: all TypeDB Cluster CI jobs will fail because it hasn't been upgraded to the latest TypeDB Core.

PR #367, opened 2022-01-20
Propagate exceptions throughout transactions

flyingsilverfin:

We align with Client NodeJS version 2.6.1 (mostly the work in https://github.com/vaticle/typedb-client-nodejs/pull/197), which implements a better error propagation mechanism: when an exception occurs, we store it against all the transaction's active transmit queues to retrieve whenever the user tries to perform an operation in the transaction anywhere, ensuring the user always sees the originating error.

* store errors received from gRPC against each receive queue
* return a new exception type, `transaction is closed with errors`, which throws the errors from all queues (note that this can be duplicate if there are multiple open transmit queues that have been given the same error)
* we clean up queues that are no longer needed, so we minimise the number of times the user sees the same exception
* update terminology around `ResponseCollector`, calling internal collectors `response_queues` instead of `collectors` (a collector made of collectors is a bit strange)


stream/BidirectionalStream.java
```diff
@@ -68,14 +68,14 @@ public BidirectionalStream(TypeDBStub stub, RequestTransmitter transmitter) {
...
-        return new Single<>(queue);
+        return new Single<>(requestID, queue, this);
     }
 
     public Stream<ResPart> stream(Req.Builder request) {
         UUID requestID = UUID.randomUUID();
         ResponseCollector.Queue<ResPart> collector = resPartCollector.queue(requestID);
         dispatcher.dispatch(request.setReqId(UUIDAsByteString(requestID)).build());
-        ResponsePartIterator iterator = new ResponsePartIterator(requestID, collector, dispatcher);
+        ResponsePartIterator iterator = new ResponsePartIterator(requestID, collector, this);
```
flyingsilverfin, line 78: now add the stream to the iterator/Single so they can call back when they are finished in order to clean up

stream/BidirectionalStream.java
```diff
@@ -118,22 +118,40 @@ private void close(@Nullable StatusRuntimeException error) {
...
 
-    public List<StatusRuntimeException> drainErrors() {
-        List<StatusRuntimeException> errors = new ArrayList<>(resCollector.drainErrors());
-        errors.addAll(resPartCollector.drainErrors());
+    void singleDone(UUID requestID) {
+        resCollector.remove(requestID);
+    }
+
+    void iteratorDone(UUID requestID) {
+        resPartCollector.remove(requestID);
```
flyingsilverfin, line 126: separate methods for separate collectors

stream/BidirectionalStream.java
```diff
@@ -118,22 +118,40 @@ private void close(@Nullable StatusRuntimeException error) {
...
+
+    public List<StatusRuntimeException> getErrors() {
+        List<StatusRuntimeException> errors = new ArrayList<>(resCollector.getErrors());
+        errors.addAll(resPartCollector.getErrors());
         return errors;
     }
 
+    RequestTransmitter.Dispatcher dispatcher() {
+        return dispatcher;
+    }
```
flyingsilverfin, line 137: we've replaced the `ResPartIterator` argument which is `dispatcher` with a getter for `dispatcher()` on the stream, which the Iterator now has access to

stream/ResponseCollector.java
```diff
@@ -46,64 +47,63 @@ public ResponseCollector() {
...
         }
 
         public void put(R response) {
             responseQueue.add(Either.first(new Response<>(response)));
         }
 
         public void close(@Nullable StatusRuntimeException error) {
-            responseQueue.add(Either.second(new Done(error)));
+            this.error = error;
+            responseQueue.add(Either.second(new Done()));
```
flyingsilverfin, line 106: hold on to error to be returned later like in node

stream/BidirectionalStream.java
alexjpwalker: Just a fairly general comment, not limited to this change:

How aligned are the implementations of tx exception propagation in the 3 clients?

I'm pretty sure client-python had just one `done` method here, and client-nodejs doesn't appear to have any.

flyingsilverfin: yeah python has only one collector for all single and iterator, so it doesn't need split methods. I'm going to go back to update client-nodejs so it's in one of these two styles next (but won't release it)

alexjpwalker (approved):

PR #368, opened 2022-01-21
Store error on transaction stream

flyingsilverfin:

We store any exceptions received against the transaction stream (as well as query streams) in order to propagate the error to any future transaction operations, not just open query streams.

* store errors against query streams and transaction streams, so we can error on transaction operations against a closed transaction

### DIFF
--- a/stream/BidirectionalStream.java
+++ b/stream/BidirectionalStream.java
@@ -34,6 +34,7 @@ import io.grpc.stub.StreamObserver;
 import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Stream;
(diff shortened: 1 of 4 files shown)


stream/ResponsePartIterator.java
```diff
@@ -92,8 +92,11 @@ public boolean hasNext() {
...
     public TransactionProto.Transaction.ResPart next() {
-        if (!hasNext()) throw new NoSuchElementException();
-        state = State.EMPTY;
-        return next;
+        if (stream.getError().isPresent()) throw TypeDBClientException.of(stream.getError().get());
+        else if (!hasNext()) throw new NoSuchElementException();
+        else {
+            state = State.EMPTY;
+            return next;
+        }
```
flyingsilverfin, line 100: we make the next() idempotently throw the same error if the user repeatedly calls it

stream/BidirectionalStream.java
```diff
@@ -108,6 +111,7 @@ public void close() {
 
     private void close(@Nullable StatusRuntimeException error) {
         if (isOpen.compareAndSet(true, false)) {
+            this.error = error;
```
flyingsilverfin, line 114: store the closure error against the bidirectional stream to prevent future transaction operations

alexjpwalker (approved):

PR #369, opened 2022-01-27
Don't delete response collectors in a transaction

alexjpwalker:

We no longer delete response collectors in a transaction after receiving a response to a "single" request, or receiving a "DONE" message in a stream. This fixes a possible error when loading 50+ answers in one query and then performing a second query.

See https://github.com/vaticle/typedb-client-python/pull/250, which this PR is a copy of.


common/exception/ErrorMessage.java
```diff
@@ -47,7 +47,7 @@ private ErrorMessage(String codePrefix, int codeNumber, String messagePrefix, St
         public static final Client MISSING_RESPONSE =
```
alexjpwalker, line 47: See comments on https://github.com/vaticle/typedb-client-python/pull/250/files, which this PR is a copy of.

test/integration/ClientQueryTest.java
```diff
@@ -518,6 +521,23 @@ public void testSimpleExplanation() {
         }, READ, TypeDBOptions.core().infer(true).explain(true));
     }
 
+    @Test
+    public void testStreaming() {
```
alexjpwalker, line 525: With these options (the default in TypeDB at time of writing), the server may respond with: 50 answers -> CONTINUE -> 1 answer [compensating for latency] -> DONE. The client will respond to CONTINUE with STREAM to keep iterating, and the server responds to STREAM with a 2nd DONE message.

This is expected and should be handled correctly (ie: ignored) by the client.

stream/ResponsePartIterator.java
```diff
@@ -60,7 +60,6 @@ private boolean fetchAndCheck() {
             case STREAM_RES_PART:
                 switch (resPart.getStreamResPart().getState()) {
                     case DONE:
-                        stream.iteratorDone(requestID);
```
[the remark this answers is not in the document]
flyingsilverfin: nvm!!

flyingsilverfin (approved):

PR #372, opened 2022-03-10
Query iterator only throws exeption once

flyingsilverfin:

We revert previous changes from #368 and #364, which made query queues and iterators throw the same error idempotently if there was one. However, this goes counter to standard usages of iterators and queues, which are not meant to behave idempotently (each item is only returned once, and if they have an error they should no longer be used). 

* remove idempotent error state of collectors and queues, which back query iterators

### DIFF
--- a/stream/BidirectionalStream.java
+++ b/stream/BidirectionalStream.java
@@ -76,7 +76,7 @@ public class BidirectionalStream implements AutoCloseable {
         UUID requestID = UUID.randomUUID();
         ResponseCollector.Queue<ResPart> collector = resPartCollector.queue(requestID);
         dispatcher.dispatch(request.setReqId(UUIDAsByteString(requestID)).build());
-        ResponsePartIterator iterator = new ResponsePartIterator(requestID, collector, this);
+        ResponsePartIterator iterator = new ResponsePartIterator(requestID, collector, dispatcher);
         return StreamSupport.stream(spliteratorUnknownSize(iterator, ORDERED | IMMUTABLE), false);
     }
 
--- a/stream/ResponseCollector.java
+++ b/stream/ResponseCollector.java
@@ -26,6 +26,7 @@ import com.vaticle.typedb.common.collection.Either;
 import io.grpc.StatusRuntimeException;
 
 import javax.annotation.Nullable;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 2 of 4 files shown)


stream/BidirectionalStream.java
```diff
@@ -76,7 +76,7 @@ public BidirectionalStream(TypeDBStub stub, RequestTransmitter transmitter) {
         UUID requestID = UUID.randomUUID();
         ResponseCollector.Queue<ResPart> collector = resPartCollector.queue(requestID);
         dispatcher.dispatch(request.setReqId(UUIDAsByteString(requestID)).build());
-        ResponsePartIterator iterator = new ResponsePartIterator(requestID, collector, this);
+        ResponsePartIterator iterator = new ResponsePartIterator(requestID, collector, dispatcher);
```
flyingsilverfin, line 79: we no longer rely on the bidirectional stream to retrieve the error, so we can revert this to only pass in the dispatcher rather than the entire stream

stream/ResponseCollector.java
```diff
@@ -61,20 +61,18 @@ synchronized void close(@Nullable StatusRuntimeException error) {
...
         }
 
         public R take() {
             try {
                 Either<Response<R>, Done> response = responseQueue.take();
                 if (response.isFirst()) return response.first().message();
-                else if (this.error != null) throw new TypeDBClientException(TRANSACTION_CLOSED_WITH_ERRORS, error);
-                else throw new TypeDBClientException(TRANSACTION_CLOSED);
+                else if (!response.second().error().isPresent()) throw new TypeDBClientException(TRANSACTION_CLOSED);
+                else throw TypeDBClientException.of(response.second().error().get());
```
flyingsilverfin, line 75: the error is only thrown once, after which the `TRANSACTION_CLOSED` error is thrown

stream/ResponsePartIterator.java
```diff
@@ -91,8 +91,7 @@ public boolean hasNext() {
 
     @Override
     public TransactionProto.Transaction.ResPart next() {
-        if (stream.getError().isPresent()) throw TypeDBClientException.of(stream.getError().get());
```
flyingsilverfin, line 94: the error, if there is one, will be throw via the `fetchAndCheck()` method which will read the error from it's internal queue

flyingsilverfin:
PR Review:

This change comes from noticing that the query iterator was throwing `TRANSACTION_CLOSED_WITH_ERROR`, when it should just throw the error. Through inspection we also see that we had made the query iterator throw errors idempotently, which goes counter to standard usages of iterators and queue (each item is only returned once, and if they have an error they should no longer be used). We now revert to the old behaviour which fit the required iterator behaviour.

lolski (approved):

PR #375, opened 2022-04-13
Upgrade to Bazel 5

lolski:

Make the repository use Bazel 5. Bazel 4 and below only work with the deprecated Python 2, which is no longer included in some operating systems such as the latest Mac OS X.

1. Upgrade Bazel to 5.1.1.
2. Upgrade all `@vaticle` dependencies
3. Log session pulse failure to make debugging easier.


connection/TypeDBSessionImpl.java
```diff
@@ -164,7 +167,8 @@ public void run() {
             boolean alive = false;
             try {
                 alive = stub().sessionPulse(pulseReq(sessionID)).getAlive();
-            } catch (TypeDBClientException ignored) {
+            } catch (TypeDBClientException e) {
+                LOG.debug("Unable to send session pulse", e);
```
lolski, line 171: Log session pulse to make debugging easier

test/behaviour/typeql/TypeQLSteps.java
```diff
@@ -544,7 +544,7 @@ public void each_answer_satisfies(String templatedQuery) {
         }
     }
 
-    @Then("each answer does not satisfy")
+    @Then("templated typeql match; throws exception")
```
lolski, line 547: Update step name based on the latest version of `@typedb_behaviour`

PR #380, opened 2022-05-24
Explanation Conclusion should include non-retrievable concepts

jmsfltchr:

Introduce a different answer type for conclusions inside explanations. This permits returning any concept, not only retrievable concepts, which is important for returning a thorough explanation due to unification.

- Swap `ConceptMap` for a `Map`

### DIFF
--- a/api/logic/Explanation.java
+++ b/api/logic/Explanation.java
@@ -31,16 +31,10 @@ public interface Explanation {
 
     Rule rule();
 
-    ConclusionAnswer conclusion();
+    Map<String, Concept> conclusion();
 
     ConceptMap condition();
 
     Map<String, Set<String>> variableMapping();
-
-    interface ConclusionAnswer {
-
-        Map<String, Concept> concepts();
-
-    }
 
 }
--- a/api/logic/Explanation.java
+++ b/api/logic/Explanation.java
@@ -22,6 +22,7 @@
 package com.vaticle.typedb.client.api.logic;
 
 import com.vaticle.typedb.client.api.answer.ConceptMap;
+import com.vaticle.typedb.client.api.concept.Concept;
 
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


api/logic/Explanation.java
```diff
@@ -30,7 +31,7 @@ public interface Explanation {
 
     Rule rule();
 
-    ConceptMap conclusion();
+    Map<String, Concept> conclusion();
```
alexjpwalker, line 34: I'm a little confused. A `ConceptMap` **is** a `Map<String, Concept>`. What is the distinction here?

jmsfltchr: Yes I see the confusion here, there are two distinctions.
1. A `ConceptMap` can contain only retrievable variables on the server, whereas a `conclusion` can have non-retrievable variables. So, the actual variable names that can occur in a `conclusion` is a superset of those that can occur in a `ConceptMap`.
2. A `ConceptMap` has `Explainables`. We never have this for a `Conclusion`, so we have just been leaving this empty so far when using a `ConceptMap`.

The change then is due to the above given that there is a semantic difference between the two. This is also less confusing for us devs, because if we used `ConceptMap` for `conclusion` here then it would be inconsistent with the server. The user would also see an `Explainables` that is always empty.

haikalpribadi:
We've concluded that this change, along with the Protocol prerequisite change, does not arrive at meaningful data structure. Replacing ConceptMap with Map<Str,Concept> was not meaningful given that we still provide users with the same information but with different APIs. The only change is that the users are now restricted from calling `.explanations()` but that wasn't bad to begin with, given that the return was empty. Using ConceptMap, albeit providing richer API (with `.explanations()`) than necessary was not bad given that it's still consistent. We think the there's more to this discussion, in that we need to: a) look at the overall design of ConceptMap with with a fresh perspective, and b) review more of what reasoner's new data structure actually needs. Both will need to happen in the future, and thus for now, we should stick with the current API we have.

PR #389, opened 2022-07-15
Fix memory leak in long-running sessions

dmitrii-ubskii:

A long running session used to keep the reference to each closed transaction, continually growing in size. These references are inaccessible from outside the session object. The problem was compounded in the write transactions, when the caller did not collect the responses received after insert queries, which were held onto by the closed transaction objects.

We now dispose of the closed transaction objects and free their memory.

Transaction now keeps a backreference to the session that spawned it and notifies it when it is closed. The session releases the reference to the transaction and allows the GC to free the memory.

### DIFF
--- a/connection/TypeDBSessionImpl.java
+++ b/connection/TypeDBSessionImpl.java
@@ -161,6 +161,10 @@ public class TypeDBSessionImpl implements TypeDBSession {
         }
     }
 
+    public void closed(TypeDBTransaction.Extended typeDBTransaction) {
+        transactions.remove(typeDBTransaction);
+    }
+
     private class PulseTask extends TimerTask {
 
         @Override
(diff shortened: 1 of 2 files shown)


connection/TypeDBSessionImpl.java
```diff
@@ -161,6 +161,10 @@ public void close() {
         }
     }
 
+    public void closed(TypeDBTransaction.Extended typeDBTransaction) {
+        transactions.remove(typeDBTransaction);
+    }
```
dmitrii-ubskii, line 166: I think this is okay to do even with the `transactions.forEach(TypeDBTransaction.Extended::close);` on L152 in `close()`, based on the similar construction in `SessionService` in typedb core.

flyingsilverfin: agree - the locking is used to not open transactions when the session is concurrently closed. This operation should always be fine though.

flyingsilverfin: I think this can not be `public`?

dmitrii-ubskii: You're right, this can be package access.

flyingsilverfin (approved):

PR #393, opened 2022-10-21
Serialise queries without pretty printing

flyingsilverfin:

To avoid injecting unintended tab characters into user attributes containing new lines, we utilise TypeQL's new unformatted to string conversion for serialising queries to the server.

Inserting a string with a newline character `insert $x isa blob, has value "Hello \n World"` led to the following formatted TypeQL query:
```
insert
$x isa blob,
  has value "Hello
  World";
```
which on the server side is parsed as: `insert $x isa blobg, has value "Hello \n\t World", which is not the user's actual string! To avoid this, we now serialise using TypeQL's new `toString(pretty = false)` method which does no indentation or newlines, and send this query to the server.

In the new format, the server would receive:
```
insert $x isa blob, has value "Hello 
World";
```

Which does encode exactly what the user intends.

### DIFF
--- a/dependencies/vaticle/artifacts.bzl
+++ b/dependencies/vaticle/artifacts.bzl
@@ -29,7 +29,7 @@ def vaticle_typedb_artifact():
         artifact_name = "typedb-server-{platform}-{version}.{ext}",
         tag_source = deployment["artifact.release"],
         commit_source = deployment["artifact.snapshot"],
-        commit = "ee90e02d4d74e082912059d31369e836400ff73b",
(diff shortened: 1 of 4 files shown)


haikalpribadi (approved):

lolski:
Merging this PR despite `test-behaviour-writeable-cluster` failure, which will be fixed once we have upgraded TypeDB Cluster to the latest TypeDB ([see this issue](https://github.com/vaticle/typedb-cluster/issues/431)).

PR #399, opened 2023-01-24
Add transitive dependency on netty-handler to the connection package

jamesreprise:

In line with bumping our netty dependencies (https://github.com/vaticle/dependencies/pull/415), we need to include a transitive dependency that has been introduced on netty's side between the version we used to depend on and the version we now depend on.

Add `@maven//:io_netty_netty_handler` to our dependencies for the connection module.

## Additional info

We should merge the linked dependencies PR and bump the commit of dependencies we are depending on before merging this PR.


lolski (approved):

connection/BUILD
```diff
@@ -46,6 +46,7 @@ java_library(
         "@maven//:io_grpc_grpc_netty",
         "@maven//:io_grpc_grpc_stub",
         "@maven//:io_netty_netty_all",
+        "@maven//:io_netty_netty_handler",
```
lolski, line 49: should it be under `deps` or `runtime_deps`?

jamesreprise: `deps`! It's needed at build time:
```
ERROR: /Users/james/Repos/typedb-client-java/connection/BUILD:26:13: Building connection/libconnection.jar (15 source files) failed: (Exit 1): java failed: error executing command external/remotejdk11_macos/bin/java -XX:-CompactStrings '--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED' '--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED' ... (remaining 17 arguments skipped)
connection/cluster/ClusterServerClient.java:68: error: [strict] Using type io.netty.handler.ssl.SslContext from an indirect dependency (TOOL_INFO: "external/maven/v1/https/repo1.maven.org/maven2/io/netty/netty-handler/4.1.87.Final/netty-handler-4.1.87.Final.jar").
                SslContext sslContext;
                ^
```

PR #400, opened 2023-02-27
Cluster password policy and user management support

jamesreprise:

Add functionality relevant to the changes made in https://github.com/vaticle/typedb-cluster/pull/456:
* Optional password expiry in days for each user.
* The ability for admins to set passwords of other users
* User password update, which requires old and new password

We've added the functionality that is being added in a recent TypeDB Cluster pull request:
* Support updating password now requiring both the old and the new password.
* Add support for retrieving the user password expiry in days. This is optional, depending on whether password expiration is enabled on the server.
* Allow updating password as the admin requiring only a new password.

### DIFF
--- a/api/user/UserManager.java
+++ b/api/user/UserManager.java
@@ -34,5 +34,9 @@ public interface UserManager {
 
     void create(String username, String password);
 
+    void delete(String username);
+
     Set<User> all();
+
+    void passwordSet(String username, String password);
 }
--- a/api/user/UserManager.java
+++ b/api/user/UserManager.java
@@ -27,9 +27,6 @@ import java.util.Set;
 public interface UserManager {
 
     @CheckReturnValue
-    User get(String username);
-
-    @CheckReturnValue
     boolean contains(String username);
 
     void create(String username, String password);
@@ -39,4 +36,7 @@ public interface UserManager {
     Set<User> all();
 
     void passwordSet(String username, String password);
+
+    @CheckReturnValue
+    User get(String username);
 }
--- a/api/user/UserManager.java
+++ b/api/user/UserManager.java
@@ -33,10 +33,10 @@ public interface UserManager {
 
     void delete(String username);
 
+    @CheckReturnValue
+    User get(String username);
+
     Set<User> all();
 
     void passwordSet(String username, String password);
-
-    @CheckReturnValue
-    User get(String username);
 }
--- a/common/rpc/RequestBuilder.java
+++ b/common/rpc/RequestBuilder.java
@@ -137,9 +137,7 @@ public class RequestBuilder {
 
             public static ClusterUserProto.ClusterUserManager.PasswordSet.Req passwordSetReq(String username, String password) {
                 return ClusterUserProto.ClusterUserManager.PasswordSet.Req.newBuilder()
-                        .setUsername(username)
-                        .setPassword(password)
-                        .build();
+                        .setUsername(username).setPassword(password).build();
             }
 
             public static ClusterUserProto.ClusterUserManager.Get.Req getReq(String username) {
@@ -151,10 +149,7 @@ public class RequestBuilder {
 
             public static ClusterUserProto.ClusterUser.PasswordUpdate.Req passwordUpdateReq(String username, String oldPassword, String newPassword) {
                 return ClusterUserProto.ClusterUser.PasswordUpdate.Req.newBuilder()
-                        .setUsername(username)
-                        .setOldPassword(oldPassword)
-                        .setNewPassword(newPassword)
-                        .build();
+                        .setUsername(username).setOldPassword(oldPassword).setNewPassword(newPassword).build();
             }
 
             public static ClusterUserProto.ClusterUser.Token.Req tokenReq(String username) {
--- a/common/rpc/RequestBuilder.java
+++ b/common/rpc/RequestBuilder.java
@@ -147,9 +147,9 @@ public class RequestBuilder {
 
         public static class User {
 
-            public static ClusterUserProto.ClusterUser.PasswordUpdate.Req passwordUpdateReq(String username, String oldPassword, String newPassword) {
+            public static ClusterUserProto.ClusterUser.PasswordUpdate.Req passwordUpdateReq(String username, String passwordOld, String passwordNew) {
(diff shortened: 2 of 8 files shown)


common/rpc/RequestBuilder.java
```diff
@@ -127,24 +127,39 @@ public static ClusterUserProto.ClusterUserManager.Create.Req createReq(String us
...
         public static class User {
 
-            public static ClusterUserProto.ClusterUser.Password.Req passwordReq(String username, String password) {
-                return ClusterUserProto.ClusterUser.Password.Req.newBuilder().setUsername(username).setPassword(password).build();
+            public static ClusterUserProto.ClusterUser.PasswordUpdate.Req passwordUpdateReq(String username, String oldPassword, String newPassword) {
+                return ClusterUserProto.ClusterUser.PasswordUpdate.Req.newBuilder()
+                        .setUsername(username)
+                        .setOldPassword(oldPassword)
+                        .setNewPassword(newPassword)
+                        .build();
```
flyingsilverfin, line 157: if you notice our style is to inline things that fit in one line without overflow

connection/cluster/ClusterUserManager.java
```diff
@@ -73,16 +73,70 @@ public void create(String username, String password) {
...
-                                .usersAll(allReq()).getNamesList().stream()
-                                .map(name -> new ClusterUser(client, name))
-                                .collect(Collectors.toSet())
+                (parameter) -> {
+                    Set<User> users = new HashSet<>();
+                    List<ClusterUserProto.User> protoUsers = parameter.client().stub().usersAll(allReq()).getUsersList();
+
+                    for (ClusterUserProto.User user: protoUsers) {
+                        if (user.getPasswordExpiryOptCase() ==
+                                ClusterUserProto.User.PasswordExpiryOptCase.PASSWORDEXPIRYOPT_NOT_SET) {
```
flyingsilverfin, line 98: I recon we sohuld add a `User.of(Proto)` method to hide this `if` case away into the right domain

connection/cluster/ClusterUserManager.java
```diff
@@ -73,16 +73,70 @@ public void create(String username, String password) {
...
+                SYSTEM_DB,
+                (parameter) -> {
+                    com.vaticle.typedb.protocol.ClusterUserProto.User user = parameter.client().stub().usersGet(getReq(username)).getUser();
+                    if (user.getPasswordExpiryOptCase() ==
+                            ClusterUserProto.User.PasswordExpiryOptCase.PASSWORDEXPIRYOPT_NOT_SET) {
+                        return new ClusterUser(client, user.getUsername(), Optional.empty());
+                    }
+                    else {
+                        return new ClusterUser(client, user.getUsername(), Optional.of(user.getPasswordExpiryDays()));
+                    }
```
flyingsilverfin, line 125: Ok at this point - notice we have repeated code, then we realise there is a better domain to put the construction of a `ClusterUser` - the `ClusterUser` itself!

test/behaviour/connection/user/UserSteps.java
```diff
@@ -63,9 +63,14 @@ public void users_create(String username, String password) {
...
-    public void user_password(String username, String password) {
-        getClient().users().get(username).password(password);
+    @Then("users delete: {word}")
+    public void users_delete(String username) {
+        getClient().users().delete(username);
+    }
+
+    @Then("user password update: {word}, {word}, {word}")
+    public void user_password_update(String username, String oldPassword, String newPassword) {
+        getClient().users().get(username).passwordUpdate(oldPassword, newPassword);
```
flyingsilverfin, line 73: what about one for `users_password_set`?

common/rpc/RequestBuilder.java
jamesreprise: This does overflow all on one line unfortunately. Something like this?
```
return ClusterUserProto.ClusterUserManager.PasswordSet.Req.newBuilder()
                        .setUsername(username).setPassword(password).build();
```

connection/cluster/ClusterUserManager.java
jamesreprise: Agreed, implemented.

connection/cluster/ClusterUserManager.java
jamesreprise: Agreed, implemented.

test/behaviour/connection/user/UserSteps.java
jamesreprise: Haven't done the behaviour tests yet, apologies.

jamesreprise: It seems as though the user behaviour steps aren't run in the current set of automation jobs in vaticle/typedb-client-java - we should decide what we want to do with these.

api/user/UserManager.java
```diff
@@ -26,13 +26,17 @@
...
     void create(String username, String password);
 
+    void delete(String username);
+
     Set<User> all();
+
+    void passwordSet(String username, String password);
+
+    @CheckReturnValue
+    User get(String username);
```
flyingsilverfin, line 41: it's pop this back up to the top, or at least near `all`

connection/cluster/ClusterUserManager.java
```diff
@@ -73,16 +69,47 @@ public void create(String username, String password) {
...
     @Override
     public Set<User> all() {
         ClusterClient.FailsafeTask<Set<User>> failsafeTask = client.createFailsafeTask(
                 SYSTEM_DB,
-                (parameter) ->
-                        parameter.client().stub()
-                                .usersAll(allReq()).getNamesList().stream()
-                                .map(name -> new ClusterUser(client, name))
-                                .collect(Collectors.toSet())
+                (parameter) -> parameter.client().stub().usersAll(allReq()).getUsersList().stream().map(
```
flyingsilverfin, line 88: I'd probably put a line break before one of the `.` instead of the inside the `map()` backets

flyingsilverfin (commented):
Left 2 minor comments you can address - approving now conditional on those being changed.

flyingsilverfin (approved):

jamesreprise:
> Left 2 minor comments you can address - approving now conditional on those being changed.

Done! Quantity of commits was due to rebasing on master.

PR #404, opened 2023-03-22
Concept map json

dmitrii-ubskii:
**For the title of this PR:** please follow the grammatical rules of a usual publication title, without capitalisation (except for the first letter). Thus, the title should NOT CONTAIN CODE: no dots, no parentheses, no backticks, no brackets, etc. It needs to be distinctive (not detailed) and succinct (not lengthy). Details of this PR will go in the description. **For the description of this PR:** please replace every line in curly brackets ( { like this } ) with an appropriate description following the guidance. Finally, **please remove this paragraph**.

{ In the form of a paragraph (only use bullet points if strictly necessary), please describe the goal of this PR, why they are valuable to achieve, and reference the related GitHub issues. This section will be automatically compiled into the release notes, so please:
  - describe the impact of the change in this PR to the _user_ of this repository (e.g. end user, contributor, developer).
  - describe the new product behaviour in _present tense_, and the old behaviour and how it's been changed in _past tense_.
  - Use the _Royal We_: _"We"_ made changes, not _"I"_ made changes. }

{ Please explain what you implemented, why your changes are the best way to achieve the goal(s) above. Please describe every method, class and package, by explaining:
  - its responsibility, 
  - how it's expected to behave, and 
  - how it relates to the adjacent methods/classes/packages it interacts with. 

This would allow the reviewer to understand your intentions in the code much better. If you're adding new classes, make sure these explanations are also included in the class header comments. Last but not least, please reference the GitHub issues to be automatically closed, such like 'closes #number'. }

### DIFF
--- a/api/answer/ConceptMap.java
+++ b/api/answer/ConceptMap.java
@@ -39,6 +39,9 @@ public interface ConceptMap {
     @CheckReturnValue
     Concept get(String variable);
 
+    @CheckReturnValue
+    String toJSON();
+
     Explainables explainables();
 
     interface Explainables {
--- a/test/behaviour/concept/serialization/SerializationSteps.java
+++ b/test/behaviour/concept/serialization/SerializationSteps.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2022 Vaticle
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.vaticle.typedb.client.test.behaviour.concept.serialization;
+
+import com.eclipsesource.json.Json;
+import com.eclipsesource.json.JsonValue;
+import com.vaticle.typedb.client.api.answer.ConceptMap;
+import io.cucumber.java.en.Then;
+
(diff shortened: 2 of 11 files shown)


api/answer/ConceptMap.java
```diff
@@ -39,6 +41,13 @@ public interface ConceptMap {
     @CheckReturnValue
     Concept get(String variable);
 
+    @CheckReturnValue
+    default JsonObject toJSON() {
+        JsonObject object = Json.object();
+        map().forEach((resVar, resConcept) -> object.add(resVar, resConcept.toJSON()));
+        return object;
+    }
```
dmitrii-ubskii, line 49: All implementations of `toJSON()` are done in the `api` package. Otherwise the implementation would be duplicated between e.g. `ThingImpl` and `ThingImpl.Remote`, since `...Impl.Remote` don't extend their enclosing counterparts.

test/behaviour/concept/serialization/SerializationSteps.java
```diff
@@ -0,0 +1,78 @@
...
+        if (left.size() != right.size()) return false;
+        return left.names().stream().allMatch((name) -> JSONValuesAreEqual(left.get(name), right.get(name)));
+    }
+
+    private boolean JSONArraysAreEqual(JsonArray left, JsonArray right) {
+        if (left.size() != right.size()) return false;
+        for (int i = 0; i < left.size(); i++)
+            if (!JSONValuesAreEqual(left.get(i), right.get(i))) return false;
+        return true;
+    }
```
dmitrii-ubskii, line 77: `JsonObject`s are ordered, and by default the comparison would respect the order of keys. Since we don't specify the order, we have to define our own equality check for `JsonValue`s.

benjaminasmall (approved):

test/behaviour/concept/serialization/SerializationTest.java
```diff
@@ -0,0 +1,58 @@
...
+import io.cucumber.junit.CucumberOptions;
+import org.junit.runner.RunWith;
+
+@RunWith(Cucumber.class)
+@CucumberOptions(
+        strict = true,
+        plugin = "pretty",
+        glue = "com.vaticle.typedb.client.test.behaviour",
+        features = "external/vaticle_typedb_behaviour/concept/serialization.feature",
+        tags = "not @ignore and not @ignore-typedb"
```
flyingsilverfin, line 35: Can you double check the `@ignore-typedb` is correct here? I think we have one for client-java or clients specifically or something? Might as well do a drive-by of the other BDD tests in this repo while you're at it

PR #409, opened 2023-04-11
Introduce Values to support expressions

krishnangovindraj:

Introduce the 'Value' type, which is returned as the result of an expression's computation. This change follows from https://github.com/vaticle/typeql/pull/260, which outlines the capabilities of the new expression syntax.

Values (representing any of Long, Double, Boolean, String, or DateTime) are returned as part of `ConceptMap` answers and are subtypes of `Concept` for the time being. Their main API is made of the `.getValue()` method and `.getValueType()` method, along with all the standard safe downcasting methods to convert a `Concept` into a `Value`, using `Concept.isValue()` and `Concept.asValue()`.

We also move the import location of `ValueType` from being nested in `AttributeType` to `Concept`.

* Introduces the `Value` concept and the required `ValueImpl` that implements it
* Refactor `ValueType` to no longer live within `AttributeType` - now it exists in `Concept.ValueType`
* Updates the test framework for tests involving values, including the new `ExpressionTest` behaviour scenarios, which we also add to CI

### DIFF
--- a/concept/value/ValueImpl.java
+++ b/concept/value/ValueImpl.java
@@ -23,7 +23,6 @@ package com.vaticle.typedb.client.concept.value;
 
 import com.vaticle.typedb.client.api.TypeDBTransaction;
 import com.vaticle.typedb.client.api.concept.Concept;
-import com.vaticle.typedb.client.api.concept.type.Type.ValueType;
 import com.vaticle.typedb.client.api.concept.value.Value;
 import com.vaticle.typedb.client.common.exception.TypeDBClientException;
 import com.vaticle.typedb.client.concept.ConceptImpl;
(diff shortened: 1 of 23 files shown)


krishnangovindraj:
Tests are red because the cluster artifact doesn't support arithmetic.

concept/value/ValueImpl.java
```diff
@@ -0,0 +1,189 @@
...
+
+    @Override
+    public ValueImpl<?> asValue() {
+        return this;
+    }
+
+    @Override
+    public Concept.Remote asRemote(TypeDBTransaction transaction) {
+        throw new TypeDBClientException(VALUECONCEPT_HAS_NO_REMOTE);
+    }
```
krishnangovindraj, line 77: Value concepts are ephemeral and have no remote.

flyingsilverfin (approved):

alexjpwalker (approved):

PR #414, opened 2023-04-27
Implement native calls from Client Java to Client Rust

dmitrii-ubskii:

We replace the direct implementation of the client side of the TypeDB Protocol in Java with native calls to the Rust client over JNI.

- merge `TypeDBCLient.Core` and `.Cluster` into single `TypeDBClient`, following the protocol merge;
- remove `Remote` variants of `Concept`s; Concept API calls now require a `Transaction` to be provided;
- make `Value` "generic" over its value type, which allows us to collapse all the `Attribute.<Type>`s into a single `Attribute` class;
- associate each top-level class (`Client`, `Session`, `Concept`, etc.) with its native representation on the Rust side;
- replace Java method implementations with calls using the native representation over JNI.


common/NativeObject.java
```diff
@@ -19,31 +19,16 @@
...
-public interface QueryFuture<T> {
-
-    T get();
-
-    default <U> QueryFuture<U> map(Function<T, U> function) {
-        return new Mapped<>(this, function);
+public abstract class NativeObject<T> {
+    static {
+        System.loadLibrary("typedb_client_jni");
     }
```
dmitrii-ubskii, line 27: Common superclass for any class that has a native representation. This ensures that the shared library is loaded before we use any of them.

concept/answer/ConceptMapImpl.java
```diff
@@ -23,208 +23,175 @@
...
+import static com.vaticle.typedb.client.jni.typedb_client.explainables_get_attribute;
+import static com.vaticle.typedb.client.jni.typedb_client.explainables_get_attributes_keys;
+import static com.vaticle.typedb.client.jni.typedb_client.explainables_get_ownership;
+import static com.vaticle.typedb.client.jni.typedb_client.explainables_get_ownerships_keys;
+import static com.vaticle.typedb.client.jni.typedb_client.explainables_get_relation;
+import static com.vaticle.typedb.client.jni.typedb_client.explainables_get_relations_keys;
+import static com.vaticle.typedb.client.jni.typedb_client.explainables_to_string;
+
+public class ConceptMapImpl extends NativeObject<com.vaticle.typedb.client.jni.ConceptMap> implements ConceptMap {
+    private final int hash;
```
flyingsilverfin, line 54: let's replace this with a proper (lazily cached) hash code using a computed hash over the streamed entries' hashes

let's apply this logic to all the `Answer` types

api/TypeDBTransaction.java
```diff
@@ -23,16 +23,15 @@
...
 import com.vaticle.typedb.client.api.query.QueryManager;
-import com.vaticle.typedb.protocol.TransactionProto;
+import com.vaticle.typedb.client.common.exception.TypeDBClientException;
 
 import java.util.function.Consumer;
 import javax.annotation.CheckReturnValue;
-import java.util.stream.Stream;
 
-public interface TypeDBTransaction extends AutoCloseable {
+import static com.vaticle.typedb.client.common.exception.ErrorMessage.Concept.BAD_ENUM_VALUE;
```
flyingsilverfin, line 32: I think we can use `Internal.Illegal_State` for these

api/TypeDBTransaction.java
```diff
@@ -71,11 +70,20 @@ enum Type {
...
 
-        public static Type of(int value) {
-            for (Type t : values()) {
-                if (t.id == value) return t;
+        public static Type of(com.vaticle.typedb.client.jni.TransactionType transactionType) {
+            switch (transactionType) {
+                case Read: return READ;
+                case Write: return WRITE;
+            }
+            throw new TypeDBClientException(BAD_ENUM_VALUE);
```
flyingsilverfin, line 78: we tend to put the `throw` in the `default` block, please apply throughout

api/concept/Value.java
```diff
@@ -0,0 +1,139 @@
...
+import static com.vaticle.typedb.client.jni.typedb_client.value_new_double;
+import static com.vaticle.typedb.client.jni.typedb_client.value_new_long;
+import static com.vaticle.typedb.client.jni.typedb_client.value_new_string;
+
+public class Value extends NativeObject<com.vaticle.typedb.client.jni.Value> {
+    private final int hash;
+
+    public Value(com.vaticle.typedb.client.jni.Value value) {
+        super(value);
+        this.hash = toString().hashCode();
```
flyingsilverfin, line 54: Let's use well-defined hash and equality for the `Concept` layer objects, since they should be pretty simple. They must be aligned: `object.equals(obj)` => `object.hashCode() == obj.hashCode()`

TypeDB.java
```diff
@@ -21,41 +21,26 @@
...
-        return new CoreClient(address, parallelisation);
-    }
-
-    public static TypeDBClient.Cluster clusterClient(String address, TypeDBCredential credential) {
-        return new ClusterClient(set(address), credential);
-    }
-
-    public static TypeDBClient.Cluster clusterClient(String address, TypeDBCredential credential, int parallelisation) {
-        return new ClusterClient(set(address), credential, parallelisation);
+    public static TypeDBConnection coreClient(String address) {
```
flyingsilverfin, line 35: I think we should continue to use `TypeDBClient`, since in Java the 'model' is still that it goes Client -> Session -> Transaction. In Rust we have `DatabaseManager(connection)`, so it's a different model.

concept/ConceptManagerImpl.java
```diff
@@ -21,45 +21,44 @@
...
+import static com.vaticle.typedb.client.jni.typedb_client.concepts_get_relation_type;
+import static com.vaticle.typedb.client.jni.typedb_client.concepts_get_schema_exceptions;
+import static com.vaticle.typedb.client.jni.typedb_client.concepts_put_attribute_type;
+import static com.vaticle.typedb.client.jni.typedb_client.concepts_put_entity_type;
+import static com.vaticle.typedb.client.jni.typedb_client.concepts_put_relation_type;
+import static com.vaticle.typedb.client.jni.typedb_client.schema_exception_code;
+import static com.vaticle.typedb.client.jni.typedb_client.schema_exception_message;
 
 public final class ConceptManagerImpl implements ConceptManager {
+    public final com.vaticle.typedb.client.jni.Transaction transaction;
```
flyingsilverfin, line 58: I have a feeling we should try to minimise the exposure of JNI objects as much as possible. Does this field _have_ to be public?

concept/thing/RelationImpl.java
```diff
@@ -25,119 +25,70 @@
...
+        relation_remove_role_player(((ConceptManagerImpl) transaction.concepts()).transaction,
+                nativeObject, ((RoleTypeImpl) roleType).nativeObject, ((ThingImpl) player).nativeObject);
     }
 
     @Override
-    public final RelationImpl asRelation() {
-        return this;
+    public Stream<ThingImpl> getPlayers(TypeDBTransaction transaction, RoleType... roleTypes) {
+        return relation_get_players_by_role_type(((ConceptManagerImpl) transaction.concepts()).transaction,
+                nativeObject, Arrays.stream(roleTypes).map(rt -> ((RoleTypeImpl) rt).nativeObject).toArray(com.vaticle.typedb.client.jni.Concept[]::new)).stream().map(ThingImpl::of);
```
flyingsilverfin, line 75: So the Rust side of this is a "ConceptIterator", what does our SWIG layer give us, a Java iterator?

concept/type/ThingTypeImpl.java
```diff
@@ -28,275 +28,278 @@
...
 
-        @Override
-        public Stream<AttributeTypeImpl> getOwns(Set<TypeQLToken.Annotation> annotations) {
-            return stream(getOwnsReq(getLabel(), protoAnnotations(annotations)))
-                    .flatMap(rp -> rp.getThingTypeGetOwnsResPart().getAttributeTypesList().stream())
-                    .map(AttributeTypeImpl::of);
-        }
+    @Override
+    public Stream<AttributeTypeImpl> getOwns(TypeDBTransaction transaction) {
+        return getOwns(transaction, Transitivity.Transitive, set());
```
flyingsilverfin, line 193: use `emptySet()` and `emptyList()` when you use an empty immutable set/list to avoid unneeded allocations

concept/type/ThingTypeImpl.java
```diff
@@ -28,275 +28,278 @@
...
-            return transactionExt.concepts().getThingType(getLabel().name()) == null;
+        public Stream<? extends ThingImpl> getInstances(TypeDBTransaction transaction) {
+            return null;
         }
 
         @Override
-        public final String getSyntax() {
-            return execute(getSyntaxReq(getLabel())).getThingTypeGetSyntaxRes().getSyntax();
+        public Stream<? extends ThingImpl> getInstancesExplicit(TypeDBTransaction transaction) {
+            return null;
```
flyingsilverfin, line 302: This is strange, we didn't have a `Root` type before at all right? First of all `instances()` should still return things, not null, and secondly, why do we have a new structure we didn't have before?

connection/TypeDBConnectionImpl.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    public boolean isOpen() {
+        return isOpen;
+    }
+
+    @Override
+    public User user() {
+        return users().getCurrentUser();
```
flyingsilverfin, line 67: I dont think we should have this public on both `Client` and `UserManager`. I think it's actually better on here -- can we hide the public `getCurrentUser()` API and just leave it on the Impl?

connection/TypeDBConnectionImpl.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    public TypeDBSession session(String database, TypeDBSession.Type type, TypeDBOptions options) {
+        return new TypeDBSessionImpl(databases().get(database), type, options);
+    }
+
+    @Override
+    public void close() {
+        isOpen = false;
+        nativeObject.delete();
```
flyingsilverfin, line 93: Interesting, can this ever fail & throw an exception?

connection/TypeDBSessionImpl.java
```diff
@@ -95,8 +56,8 @@ public Type type() {
     }
 
     @Override
-    public TypeDBDatabaseImpl database() {
-        return database;
+    public String database_name() {
```
flyingsilverfin, line 59: why can't we return the database itself?

user/BUILD
```diff
@@ -24,28 +24,26 @@ package(default_visibility = ["//visibility:public"])
...
-
-        # External dependencies from Maven
-        "@maven//:com_google_code_findbugs_jsr305",
-        "@maven//:com_google_protobuf_protobuf_java",
-        "@maven//:io_grpc_grpc_stub",
-        "@maven//:io_grpc_grpc_api",
-        "@maven//:org_slf4j_slf4j_api"
+        "@vaticle_typedb_client_rust//:typedb_client_jni",
+    ],
+    runtime_deps = [
```
flyingsilverfin, line 37: nice catch

test/behaviour/connection/user/UserSteps.java
```diff
@@ -21,45 +21,42 @@
...
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
-
 public class UserSteps {
-
-    private TypeDBClient.Cluster getClient() {
-        assert client.isCluster();
-        return (TypeDBClient.Cluster) client;
+    private TypeDBConnection getConnection() {
```
flyingsilverfin, line 38: these all need to be renamed back to `client`

logic/RuleImpl.java
```diff
@@ -23,164 +23,73 @@
...
     @Override
     public String getLabel() {
-        return label;
+        return rule_get_label(nativeObject);
     }
 
     @Override
     public Conjunction<? extends Pattern> getWhen() {
-        return when;
+        return TypeQL.parsePattern(rule_get_when(nativeObject)).asConjunction();
```
flyingsilverfin, line 55: I wonder if we  should establish a pattern of caching these moderately expensive lookup+computation steps. I think users would think of these as simple getters that should be 'free', so I think we should...?

dependencies/vaticle/repositories.bzl
```diff
@@ -25,35 +25,46 @@ def vaticle_dependencies():
...
+        commit = "b641f6a3eb390dc9a854d64ea7e4c4e4c27f5ab9", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typeql
+    )
+
+def vaticle_typedb_client_rust():
+#    git_repository(
+#        name = "vaticle_typedb_client_rust",
+#        remote = "https://github.com/dmitrii-ubskii/typedb-client-rust",
+#        branch = "ffi",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
+#    )
+    native.local_repository(
```
flyingsilverfin, line 51: note!

connection/TypeDBTransactionImpl.java
```diff
@@ -81,78 +64,60 @@ public TypeDBOptions options() {
...
     public boolean isOpen() {
-        return bidirectionalStream.isOpen();
+        if (!nativeObject.isOwned()) return false;
+        else return transaction_is_open(nativeObject);
     }
 
     @Override
     public ConceptManager concepts() {
-        return conceptMgr;
+        return new ConceptManagerImpl(nativeObject);
```
flyingsilverfin, line 73: Does it make sense to return a _new_ concept manager each time or should we should we cache and return the same one all the time?

connection/TypeDBConnectionImpl.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    public TypeDBSession session(String database, TypeDBSession.Type type, TypeDBOptions options) {
+        return new TypeDBSessionImpl(databases().get(database), type, options);
+    }
+
+    @Override
+    public void close() {
+        isOpen = false;
```
flyingsilverfin, line 92: This should probably be an atomic boolean-style protection block to avoid RACY double-free of the native object, right?

flyingsilverfin (changes requested):
I commented on stuff from yesterday for the first two or three comments, please ignore.
I've also left off reading the tests, I think those should be OK. Please flag anything you want to look at to me explicitly :)

connection/TypeDBTransactionImpl.java
dmitrii-ubskii: It's a simple projection, which shouldn't be expensive.

concept/thing/RelationImpl.java
dmitrii-ubskii: Yep.

connection/TypeDBConnectionImpl.java
dmitrii-ubskii: Agreed, but it's not possible unless `User` is placed under `Connection`

connection/TypeDBConnectionImpl.java
dmitrii-ubskii: All good points. This should call Rust's `force_close`.

connection/TypeDBTransactionImpl.java
dmitrii-ubskii: Actually scratch that, this was needed before `nativeObject` became `final`, we can cache now.

connection/TypeDBConnectionImpl.java
dmitrii-ubskii: Discussed and fixed

`close()` is not equivalent to `finalize()`, it only shuts down the network agent without cleaning up the client data in Rust

api/concept/thing/Attribute.java
```diff
@@ -25,86 +25,72 @@
...
     @CheckReturnValue
     default boolean isString() {
-        return false;
+        return getValue().isString();
     }
 
     @CheckReturnValue
     default boolean isDateTime() {
-        return false;
+        return getValue().isDateTime();
```
flyingsilverfin, line 80: let's remove these and just have them on `Value`

api/logic/Explanation.java
```diff
@@ -23,17 +23,14 @@
...
-
     Rule rule();
 
     ConceptMap conclusion();
 
     ConceptMap condition();
 
-    Map<String, Set<String>> variableMapping();
-
+    Set<String> getMapping(String var);
```
flyingsilverfin, line 35: This isn't clear what the key `var` is going to get - is it a variable from the `condition` or the `conclusion`? Maybe we should update the naming?

concept/answer/ConceptMapGroupImpl.java
```diff
@@ -24,52 +24,55 @@
...
     }
 
     @Override
     public int hashCode() {
+        if (hash == 0) hash = computeHash();
         return hash;
     }
+
+    private int computeHash() {
+        return Objects.hash(owner(), conceptMaps().collect(Collectors.toList()));
```
flyingsilverfin, line 76: dont want to inline this?

connection/TypeDBTransactionImpl.java
```diff
@@ -81,78 +72,60 @@ public TypeDBOptions options() {
...
     @Override
-    public Res execute(Req.Builder request) {
-        return execute(request, true);
-    }
-
-    private Res execute(Req.Builder request, boolean batch) {
-        return query(request, batch).get();
+    public void commit() {
+        if (!nativeObject.isOwned()) throw new TypeDBClientException(TRANSACTION_CLOSED);
+        transaction_commit(nativeObject.released());
```
flyingsilverfin, line 103: let's leave a comment with a small clarification

query/QueryManagerImpl.java
```diff
@@ -46,23 +43,24 @@
...
+import static com.vaticle.typedb.client.jni.typedb_client.query_insert;
+import static com.vaticle.typedb.client.jni.typedb_client.query_match;
+import static com.vaticle.typedb.client.jni.typedb_client.query_match_aggregate;
+import static com.vaticle.typedb.client.jni.typedb_client.query_match_group;
+import static com.vaticle.typedb.client.jni.typedb_client.query_match_group_aggregate;
+import static com.vaticle.typedb.client.jni.typedb_client.query_undefine;
+import static com.vaticle.typedb.client.jni.typedb_client.query_update;
 
 public final class QueryManagerImpl implements QueryManager {
+    private final com.vaticle.typedb.client.jni.Transaction transaction;
```
flyingsilverfin, line 60: Let's call these `nativeTransaction` variable names throughout

connection/TypeDBDatabaseManagerImpl.java
```diff
@@ -24,53 +24,44 @@
...
-import static com.vaticle.typedb.client.common.rpc.RequestBuilder.Core.DatabaseManager.createReq;
+import static com.vaticle.typedb.client.jni.typedb_client.database_manager_new;
+import static com.vaticle.typedb.client.jni.typedb_client.databases_all;
+import static com.vaticle.typedb.client.jni.typedb_client.databases_contains;
+import static com.vaticle.typedb.client.jni.typedb_client.databases_create;
+import static com.vaticle.typedb.client.jni.typedb_client.databases_get;
 import static java.util.stream.Collectors.toList;
 
 public class TypeDBDatabaseManagerImpl implements DatabaseManager {
+    private final com.vaticle.typedb.client.jni.DatabaseManager databaseManager;
```
flyingsilverfin, line 39: `nativeDatabaseMgr`

api/query/QueryManager.java
```diff
@@ -52,16 +51,16 @@ public interface QueryManager {
     Stream<ConceptMap> match(String query, TypeDBOptions options);
 
     @CheckReturnValue
-    QueryFuture<Numeric> match(TypeQLMatch.Aggregate query);
+    Numeric match(TypeQLMatch.Aggregate query);
```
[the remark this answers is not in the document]
flyingsilverfin: I think the architecture of the `QueryFuture` replacements should be to have them do a single-shot future, like a single-element stream. It's async "enough", just like a match stream is "async" in that it computes and only blocks when calling `commit()` or `hasNext()`/`next()`

The goal is to enable users to fire off a bunch of queries at once, then collect the answers together later

api/TypeDBSession.java
```diff
@@ -51,31 +51,26 @@ public interface TypeDBSession extends AutoCloseable {
...
+            this.nativeObject = nativeObject;
+
+            this.isSchema = nativeObject == com.vaticle.typedb.client.jni.SessionType.Schema;
         }
 
         public static Type of(com.vaticle.typedb.client.jni.SessionType sessionType) {
             switch (sessionType) {
                 case Data: return DATA;
                 case Schema: return SCHEMA;
+                default: throw new TypeDBClientException(UNEXPECTED_NATIVE_VALUE);
```
dmitrii-ubskii, line 72: for loop

api/concept/type/ThingType.java
```diff
@@ -129,12 +131,12 @@ default ThingType asThingType() {
...
     String getSyntax(TypeDBTransaction transaction);
 
-    public class Annotation extends NativeObject<com.vaticle.typedb.client.jni.Annotation> {
+    class Annotation extends NativeObject<com.vaticle.typedb.client.jni.Annotation> {
         private final int hash;
 
         private Annotation(com.vaticle.typedb.client.jni.Annotation annotation) {
             super(annotation);
-            this.hash = toString().hashCode();
+            this.hash = Boolean.hashCode(isKey());
```
dmitrii-ubskii: `Objects.hash`

api/logic/Explanation.java
```diff
@@ -23,10 +23,16 @@
...
 public interface Explanation {
     Rule rule();
 
     ConceptMap conclusion();
 
     ConceptMap condition();
+
+    Set<String> mappedVariables();
+
+    Set<String> conclusionMapping(String var);
```
dmitrii-ubskii: `queryVariables()`
`queryVariableMapping(var)`

concept/type/ThingTypeImpl.java
```diff
@@ -277,27 +277,45 @@ public ThingTypeImpl getSupertype(TypeDBTransaction transaction) {
...
+                            transaction.concepts().getRootEntityType().getSubtypes(transaction),
+                            transaction.concepts().getRootRelationType().getSubtypes(transaction)),
+                    transaction.concepts().getRootAttributeType().getSubtypes(transaction)).map(tt -> (ThingTypeImpl) tt);
         }
 
         @Override
         public Stream<? extends ThingTypeImpl> getSubtypesExplicit(TypeDBTransaction transaction) {
-            return null;
+            return Stream.concat(
+                    Stream.of(transaction.concepts().getRootEntityType(), transaction.concepts().getRootRelationType(), transaction.concepts().getRootAttributeType()),
```
dmitrii-ubskii, line 295: just this

concept/type/ThingTypeImpl.java
```diff
@@ -277,27 +277,45 @@ public ThingTypeImpl getSupertype(TypeDBTransaction transaction) {
...
         }
 
         @Override
         public Stream<? extends ThingImpl> getInstancesExplicit(TypeDBTransaction transaction) {
-            return null;
+            return Stream.concat(
+                    Stream.concat(
+                            transaction.concepts().getRootEntityType().getInstancesExplicit(transaction),
+                            transaction.concepts().getRootRelationType().getInstancesExplicit(transaction)),
+                    transaction.concepts().getRootAttributeType().getInstancesExplicit(transaction)).map(t -> (ThingImpl) t);
```
dmitrii-ubskii: empty

flyingsilverfin (approved):

PR #417, opened 2023-05-16
Check BDD value equality using native types

dmikhalin:

In BDD steps implementation we checked the equality of all values through the conversion to `String`s. It might be incorrect when comparing `Double`s especially if one of these values is a result of arithmetics. 
Now we compare `Double`s by their absolute error. 
For consistency we compare all values using their native types equality.

- We modified `ValueUniquenessCheck` class in order to compare values using their native types equality.
- We modified `KeyUniquenessCheck` class: we find the desired key attribute label and compare each type immediately instead of converting all values to `String`.

### DIFF
--- a/test/behaviour/typeql/TypeQLSteps.java
+++ b/test/behaviour/typeql/TypeQLSteps.java
@@ -666,7 +666,7 @@ public class TypeQLSteps {
                 case LONG:
                     return value.equals(attribute.asLong().getValue().toString());
                 case DOUBLE:
-                    return precisionEquals(Double.parseDouble(value),  attribute.asDouble().getValue(), 1e9);
+                    return precisionEquals(Double.parseDouble(value), attribute.asDouble().getValue());
                 case STRING:
                     return value.equals(attribute.asString().getValue());
                 case DATETIME:
@@ -707,7 +707,7 @@ public class TypeQLSteps {
                     case LONG:
                         return value.equals(key.asLong().getValue().toString());
                     case DOUBLE:
-                        return precisionEquals(Double.parseDouble(value),  key.asDouble().getValue(), 1e9);
+                        return precisionEquals(Double.parseDouble(value), key.asDouble().getValue());
                     case STRING:
                         return value.equals(key.asString().getValue());
                     case DATETIME:
@@ -727,7 +727,7 @@ public class TypeQLSteps {
         }
     }
 
-    private static Boolean precisionEquals(Double a, Double b, Double epsilon) {
-        return Math.abs(a - b) < epsilon;
+    private static Boolean precisionEquals(Double a, Double b) {
+        return Math.abs(a - b) / Math.abs(a) < 1.0e6;
     }
 }
--- a/test/behaviour/typeql/TypeQLSteps.java
+++ b/test/behaviour/typeql/TypeQLSteps.java
@@ -697,12 +697,11 @@ public class TypeQLSteps {
                 return false;
             }
 
-            Optional<? extends Attribute<?>> opt_key = concept.asThing().asRemote(tx()).getHas(set(KEY))
+            Optional<? extends Attribute<?>> keyOpt = concept.asThing().asRemote(tx()).getHas(set(KEY))
                     .filter(attr -> attr.getType().getLabel().equals(type)).findFirst();
-            if (opt_key.isEmpty())
-                return false;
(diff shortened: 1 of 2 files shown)


flyingsilverfin:
Ok so some comments:
1. Let's make the `precisionEqual` a functionality available in TypeDB common:
```
public class Double {

    public static final double EPSILON = 1e-6;

    public static boolean equalsApproximate(double first, double second) {
        return equalsApproximate(first, second, EPSILON);
    }

    public static boolean equalsApproximate(double first, double second, double epsilon) {
        return Math.abs(first - second) < epsilon;
    }
}

```

Dmitrii and I derived this after some discussion in the office. This will be our centralised double-equality function in TypeDB core, cluster, and client-java (and any TypeDB Java repository in the future).

You should add this class in a PR in `@vaticle_typedb_common/util/Double`. When that's done you can update the dependency on typedb-common this repo and import that the new function above.

2. When you're making changes like this, make sure you're applying the logic of the change consistently: the goal was to make equality use machine-native equality rather than strings, where possible. As such you should also parse the "boolean" and "long" comparisons, the same way double & datetime area already done. Only Strings should use String comparison! This is the most consistent approach.

test/behaviour/typeql/TypeQLSteps.java
```diff
@@ -694,34 +695,34 @@ public boolean check(Concept concept) {
             }
 
             Set<Attribute<?>> keys = concept.asThing().asRemote(tx()).getHas(true).collect(Collectors.toSet());
-            HashMap<Label, String> keyMap = new HashMap<>();
 
             for (Attribute<?> key : keys) {
-                String keyValue;
+                if (!key.getType().getLabel().equals(type))
+                    continue;
```
flyingsilverfin, line 701: I would do the filtering for this on line 697 using the Java streams API: `...getHas(true).filter(attr -> attr.getType().getLabel().equals(type)).findFirst()`. This returns an `Optional`, which if it doesn't exist you can immediately throw an useful exception

test/behaviour/typeql/TypeQLSteps.java
```diff
@@ -694,34 +695,34 @@ public boolean check(Concept concept) {
...
-                        break;
+                        return value.equals(key.asString().getValue());
                     case DATETIME:
-                        keyValue = key.asDateTime().getValue().toString();
-                        break;
+                        LocalDateTime dateTime;
+                        try {
+                            dateTime = LocalDateTime.parse(value);
+                        } catch (DateTimeParseException e) {
+                            dateTime = LocalDate.parse(value).atStartOfDay();
```
flyingsilverfin, line 716: why do we need this?

dmikhalin: The `DATETIME` value in the step could be written as DateTime or Date. This code snippet already existed in `ValueUniquenessCheck` method, I just copied it to consistency.

test/behaviour/typeql/TypeQLSteps.java
```diff
@@ -693,35 +695,33 @@ public boolean check(Concept concept) {
...
-                    case DATETIME:
-                        keyValue = key.asDateTime().getValue().toString();
-                        break;
-                    case OBJECT:
-                    default:
-                        throw new ScenarioDefinitionException("Unrecognised value type " + key.getType().getValueType());
-                }
-
-                keyMap.put(key.getType().getLabel(), keyValue);
+            Optional<? extends Attribute<?>> opt_key = concept.asThing().asRemote(tx()).getHas(true)
```
flyingsilverfin, line 698: In Java we don't use `_` for variable names, it's camelCase. Also I would call this `keyOpt` instead of `optKey` I think

test/behaviour/typeql/TypeQLSteps.java
```diff
@@ -693,35 +695,33 @@ public boolean check(Concept concept) {
...
+                    .filter(attr -> attr.getType().getLabel().equals(type)).findFirst();
+            if (opt_key.isEmpty())
+                return false;
```
flyingsilverfin, line 701: you can put these on one line

flyingsilverfin (approved):
I've approved but please make the remaining suggested changes before the merge :)

flyingsilverfin:
It looks like you'll need to rebase this PR off Master since there's a conflict!

flyingsilverfin (approved):

PR #421, opened 2023-05-27
Introduce protocol versioning

flyingsilverfin:

We use a new protocol API to perform a "connection open". This API does server-side protocol version compatibility checks, and replaces our previous need to get all databases to check that the connection is available.

The user will receive an error about a protocol version mismatch if they are using a client-server combination that are not using exactly compatible protocols. 
1. The server will raise an error if the client tries to connect with a mismatching protocol version
2. The client will raise an error if it tries to connect to the server and the server does not have that API

Both errors imply a client-server mismatch and the error will suggest this as a fix.

This change depends on https://github.com/vaticle/typedb-protocol/pull/185, which means that this client will implement protocol version 1.

* Refactor `validateConnectionOrThrow` into `openConnection`, was then inlined into the constructors of `CoreClient` and `ClusterServerClient`
* Call the new `openConnection` with the protocol Version set when constructing a client
* Throw a useful message if the server returns an `unimplemented` error

### DIFF
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -28,40 +28,42 @@ public abstract class ErrorMessage extends com.vaticle.typedb.common.exception.E
     }
 
     public static class Client extends ErrorMessage {
+        public static final Client RPC_METHOD_UNAVAILABLE =
+                new Client(1, "The server does not support this method, please check the client-server compatibility.\n'%s'.");
         public static final Client CLIENT_CLOSED =
-                new Client(1, "The client has been closed and no further operation is allowed.");
+                new Client(2, "The client has been closed and no further operation is allowed.");
         public static final Client SESSION_CLOSED =
-                new Client(2, "The session has been closed and no further operation is allowed.");
+                new Client(3, "The session has been closed and no further operation is allowed.");
         public static final Client TRANSACTION_CLOSED =
-                new Client(3, "The transaction has been closed and no further operation is allowed.");
+                new Client(4, "The transaction has been closed and no further operation is allowed.");
         public static final Client TRANSACTION_CLOSED_WITH_ERRORS =
-                new Client(4, "The transaction has been closed with error(s): \n%s.");
+                new Client(5, "The transaction has been closed with error(s): \n%s.");
         public static final Client UNABLE_TO_CONNECT =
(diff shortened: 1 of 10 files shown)


common/exception/ErrorMessage.java
```diff
@@ -28,40 +28,42 @@ private ErrorMessage(String codePrefix, int codeNumber, String messagePrefix, St
     }
 
     public static class Client extends ErrorMessage {
+        public static final Client RPC_METHOD_UNAVAILABLE =
+                new Client(1, "The server does not support this method, please check the client-server compatibility:\n'%s'.");
```
alexjpwalker, line 32: This error message could be friendlier, something like
"The server does not support this method. Please ensure that the TypeDB Client and TypeDB Server versions are compatible ..."

common/exception/TypeDBClientException.java
```diff
@@ -85,6 +93,11 @@ private static boolean isPasswordCredentialExpired(StatusRuntimeException status
                 statusRuntimeException.getStatus().getDescription() != null &&
                 statusRuntimeException.getStatus().getDescription().contains(CLUSTER_PASSWORD_CREDENTIAL_EXPIRED_ERROR_CODE);
     }
+
+    private static boolean isUnimplementedMethod(StatusRuntimeException statusRuntimeException) {
+        return statusRuntimeException.getStatus().getCode() == Status.Code.UNIMPLEMENTED;
```
alexjpwalker, line 98: This is a sufficient but not necessary condition for a client-server mismatch error. It's also possible that we send a request message with an RPC code that _is_ implemented, but carries a different semantic meaning in this protocol version.

Do we perform some kind of handshake anywhere to verify the protocol version? I notice `OpenReq` now sends a version number. Does that mean the server will decline a connection request from an invalid client?

flyingsilverfin: Yep, and it throws basically the equivalent error message if it is a mismatch :) between the two of them we should get pretty good coverage of client-protocol mismatches -> useful errors the users can resolve by themselves.

common/exception/ErrorMessage.java
flyingsilverfin: Will push an update, thank you!

PR #422, opened 2023-06-02
Split client construction and connection validation

flyingsilverfin:

To prevent null pointer errors during failover tasks being executed, we go back to the previous model of first creating a client, then opening and validating the connection.

* Split client constructor and client opening internally, so that cluster-server-client can be retained and opened even if the connection is not available at the time.

### DIFF
--- a/connection/TypeDBClientImpl.java
+++ b/connection/TypeDBClientImpl.java
@@ -37,6 +37,7 @@ import java.util.concurrent.TimeUnit;
 
 import static com.vaticle.typedb.client.common.exception.ErrorMessage.Client.CLIENT_NOT_OPEN;
 import static com.vaticle.typedb.client.common.exception.ErrorMessage.Internal.ILLEGAL_CAST;
+import static com.vaticle.typedb.client.common.rpc.RequestBuilder.Connection.openReq;
 import static com.vaticle.typedb.common.util.Objects.className;
 
 public abstract class TypeDBClientImpl implements TypeDBClient {
@@ -46,7 +47,7 @@ public abstract class TypeDBClientImpl implements TypeDBClient {
     private final RequestTransmitter transmitter;
     private final TypeDBDatabaseManagerImpl databaseMgr;
     private final ConcurrentMap<ByteString, TypeDBSessionImpl> sessions;
-    protected boolean connectionValidated;
+    private boolean connectionValidated;
 
     protected TypeDBClientImpl(int parallelisation) {
         NamedThreadFactory threadFactory = NamedThreadFactory.create(TYPEDB_CLIENT_RPC_THREAD_NAME);
@@ -56,7 +57,10 @@ public abstract class TypeDBClientImpl implements TypeDBClient {
         connectionValidated = false;
     }
 
-    protected abstract void validateConnection();
+    public void validateConnection() {
+        stub().connectionOpen(openReq());
+        connectionValidated = true;
+    }
 
     @Override
     public boolean isOpen() {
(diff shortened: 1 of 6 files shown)


connection/cluster/ClusterClient.java
```diff
@@ -93,13 +93,14 @@ private Map<String, ClusterServerClient> createClients(TypeDBCredential credenti
         Map<String, ClusterServerClient> clients = new HashMap<>();
         boolean available = false;
         for (String address : addresses) {
+            ClusterServerClient client = new ClusterServerClient(address, credential, parallelisation);
             try {
-                ClusterServerClient client = new ClusterServerClient(address, credential, parallelisation);
-                clients.put(address, client);
+                client.open();
```
flyingsilverfin, line 98: this 'attempt's to open the connection, but we retain the client afterward anyway

connection/cluster/ClusterServerClient.java
```diff
@@ -52,12 +52,17 @@ class ClusterServerClient extends TypeDBClientImpl {
...
+
+    protected void open() {
         try {
             stub.connectionOpen(openReq());
         } catch (Exception e) {
             close();
             throw e;
         }
+        isOpen = true;
     }
```
dmitrii-ubskii, line 66: This appears identical with the `CoreClient` implementation and should probably be implemented on `TypeDBClientImpl`.

dmitrii-ubskii (changes requested):

connection/TypeDBClientImpl.java
```diff
@@ -45,12 +46,21 @@ public abstract class TypeDBClientImpl implements TypeDBClient {
...
         databaseMgr = new TypeDBDatabaseManagerImpl(this);
         sessions = new ConcurrentHashMap<>();
+        connectionValidated = false;
+    }
+
+    protected abstract void validateConnection();
+
+    @Override
+    public boolean isOpen() {
+        return !channel().isShutdown() && connectionValidated;
```
flyingsilverfin, line 63: "open" now means the channel is available and the connection has been validated

dmitrii-ubskii: I think these should be in different accessors now. There's currently no way to check whether the client is in a "destroyed" state (i.e. `.close()` was explicitly called), or if it's in a "not yet validated" state where it can be reopened.

connection/TypeDBClientImpl.java
```diff
@@ -108,13 +114,14 @@ void removeSession(TypeDBSessionImpl session) {
 
     @Override
     public void close() {
-        try {
-            sessions.values().forEach(TypeDBSessionImpl::close);
-            channel().shutdown().awaitTermination(10, TimeUnit.SECONDS);
-            transmitter.close();
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
+        if (isOpen()) {
```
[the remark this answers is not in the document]
dmitrii-ubskii: See my previous comment. If the client has been opened but not validated (network error, whatever else) the channel will not be shut down.

connection/cluster/ClusterClient.java
```diff
@@ -320,6 +322,12 @@ private void waitForPrimaryReplicaSelection() {
             }
         }
 
+        private ClusterServerClient fetchOpenServerClient(String address) {
+            ClusterServerClient serverClient = clusterServerClient(address);
+            if (!serverClient.isOpen()) serverClient.validateConnection(); // may throw exception
```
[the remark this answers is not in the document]
dmitrii-ubskii: Here too, `isOpen()` represents both transient network failures, for which we want to reattempt a handshake, and a forcibly closed client, for which we don't.

dmitrii-ubskii (changes requested):

dmitrii-ubskii (approved):

PR #428, opened 2023-07-13
Merge TypeDB Rust driver into TypeDB Java driver repository

dmitrii-ubskii:

We merge driver-rust into driver-java (which is using driver-rust as the backend) with the ultimate goal of unifying all drivers under the same typedb-driver repo at a later date.

- move the Java driver implementation into the top-level `java/` package;
- move the Rust driver implementation into the top-level `rust/` package;
- merge the two git trees, erring on the side of driver-java as the source of truth for e.g. version and dependencies.


flyingsilverfin:
High level comments in 1 place:
1. Deploy github at `//:deploy-github` doesn't exist, only used in the release workflow but should exist. It should probably deploy all languages at once to github?
2. You can move the README into the `//java` package and update its contents to reflect the correct paths. Add a new one at the root for now that points to each sub-package. We can edit it a bit later

Note: you could also comment out the `release` steps and we re-activate them later, but it might be a larger jump to do later

PR #431, opened 2023-07-24
Improve error handling with cluster connection issues

krishnangovindraj:
Improves handing of errors encountered while connecting to typedb cluster.

1. If we cannot connect to any node during client initialisation, We append the individual reasons to the general `CLUSTER_UNABLE_TO_CONNECT` exception message.

2. We extend support for translating grpc `StatusRuntimeException`s. The following error codes were introduced for the following scenarios
* Unencrypted server:
  - Connecting with an encrypted client throws `CLUSTER_SERVER_NOT_ENCRYPTED` (typedb-cluster#445)
* Encrypted server
  - Connecting with a server whose identity cannot be verified (e.g., wrong CA certificate is passed to the client) throws `CLUSTER_SSL_CERTIFICATE_NOT_VALIDATED` 
- Connecting to a server through a hostname which does not appear on its certificate throws `CLUSTER_SSL_CERTIFICATE_INVALID_FOR_HOSTNAME`
- Other SSLHandshakeExceptions throw a generic `CLUSTER_SSL_HANDSHAKE_FAILED` exception, wrapping the cause.
  - Connecting with an unencrypted client is hard to detect, since the `StatusRuntimeException` has no cause and a vague message: "Network closed for unknown reason"
  - We still throw the general `CLUSTER_UNABLE_TO_CONNECT`. 
  - When this is during client initialisation (usually the case), the individual reasons are appended to the message.

- fixes vaticle/typedb-cluster#445
- fixes vaticle/typedb-cluster#492


krishnangovindraj (commented):
self review

connection/cluster/ClusterClient.java
```diff
@@ -294,13 +304,17 @@ private ClusterDatabase.Replica seekPrimaryReplica() {
...
+                String serverAddress = clusterAddresses.get((roundRobinIndex + i) % clusterAddresses.size());
                 try {
                     LOG.debug("Fetching replica info from {}", serverAddress);
                     ClusterDatabaseProto.ClusterDatabaseManager.Get.Res res = fetchValidatedServerClient(serverAddress)
                             .stub().databasesGet(getReq(database));
                     ClusterDatabase clusterDatabase = ClusterDatabase.of(res.getDatabase(), ClusterClient.this);
-                    clusterDatabases.put(database, clusterDatabase);
+                    if (!isNonExistentRaftClusterInfo(clusterDatabase)) {
+                        clusterDatabases.put(database, clusterDatabase);
+                    }
```
krishnangovindraj, line 317: Don't cache nonExistentRaftClusterInfo responses.

krishnangovindraj (commented):
Self review

common/exception/TypeDBClientException.java
```diff
@@ -98,6 +109,25 @@ private static boolean isUnimplementedMethod(StatusRuntimeException statusRuntim
...
 
+    private static TypeDBClientException tryGetRstStreamCause(StatusRuntimeException sre) {
+        Throwable cause = null;
+        if (sre.getStatus().getCode() == Status.Code.UNAVAILABLE) {
+            if (sre.getCause() instanceof javax.net.ssl.SSLHandshakeException) {
+                if (sre.getCause().getMessage().contains("PKIX path building failed")) {
+                    return new TypeDBClientException(sre.getCause(), CLUSTER_SSL_CERTIFICATE_NOT_VALIDATED);
+                } else if (sre.getCause().getMessage().contains("No subject alternative DNS name")) {
+                    return new TypeDBClientException(sre.getCause(), CLUSTER_SSL_CERTIFICATE_INVALID_FOR_HOSTNAME);
+                } else return new TypeDBClientException(sre.getCause(), CLUSTER_SSL_HANDSHAKE_FAILED, sre.getMessage());
```
krishnangovindraj, line 120: I introduce an exception for this branch. Should I let it fall through to the end instead?

krishnangovindraj (commented):
self review, after removing the cluster database changes

common/exception/TypeDBClientException.java
```diff
@@ -48,7 +53,13 @@ public TypeDBClientException(ErrorMessage error, Object... parameters) {
         this.errorMessage = error;
     }
 
-    public TypeDBClientException(String message, Throwable cause) {
+    public TypeDBClientException(Throwable cause, ErrorMessage error, Object... parameters) {
+        super(error.message(parameters), cause);
+        assert !getMessage().contains("%s");
```
flyingsilverfin, line 58: I don't think this should be possible since the string formatting should fail if the number of args doesn't match? Otherwise I think this should move into the `error.message()` call since we'd like to encapsulate that behaviour in the creation of the message from the error

test/integration/ClusterEncryptionTest.java
```diff
@@ -0,0 +1,126 @@
...
+            try (TypeDBClient.Cluster client = TypeDB.clusterClient(clusterRunner.externalAddresses(), ENCRYPTED_CREDENTIALS)) {
+                LOG.info("Encrypted client connected successfully");
+            } catch (Exception e) {
+                fail("The encrypted client with correct CA failed to connect.");
+            }
+
+            try (TypeDBClient.Cluster client = TypeDB.clusterClient(clusterRunner.externalAddresses(), MISCONFIGURED_CREDENTIALS)) {
+                fail("The encrypted client with wrong CA connected (should have failed).");
+            } catch (TypeDBClientException e) {
+                Assert.assertEquals(CLUSTER_SSL_CERTIFICATE_NOT_VALIDATED, e.getErrorMessage());
```
flyingsilverfin, line 88: static imports can be used

test/integration/ClusterEncryptionTest.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    private void createCluster(int nodes, Map<String, String> serverOpts) {
+        if (clusterRunner == null) shutDownCluster();
+        clusterRunner = TypeDBClusterRunner.create(Paths.get("."), nodes, serverOpts);
+    }
+
+    @Test
+    public void encrypted_server_connection_tests() {
+        createCluster(1, Collections.singletonMap("--server.encryption.enable", "true"));
```
flyingsilverfin, line 76: Do you have to tell the server to use the `CLUSTER_CA` anywhere? Right now you're using the default cluster CA? If so, you should be accessing that one instead of having a copy of it as `dev-root-ca.pem`, or you can inject the `dev-root-ca.pem` into the cluster via the configuration in `createCluster()`

connection/cluster/ClusterClient.java
```diff
@@ -75,19 +75,22 @@ public ClusterClient(Set<String> initAddresses, TypeDBCredential credential, int
...
+                    LOG.warn("Unable to fetch list of all servers from server {}.", server);
+                    if (e.getCause() != null) causes += "\t* " + server + ": " + e.getCause().getMessage() + "\n";
                 } else {
                     throw e;
                 }
             }
         }
-        throw new TypeDBClientException(CLUSTER_UNABLE_TO_CONNECT, String.join(",", servers));
+        String description = causes.isBlank() ? String.join(",", servers) : "Reasons:[\n" + causes + "]";
+        throw new TypeDBClientException(CLUSTER_UNABLE_TO_CONNECT, description);
```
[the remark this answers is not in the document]
flyingsilverfin: I'm not going to review this here since it depends on the other PR going in first with this change

common/exception/TypeDBClientException.java
```diff
@@ -98,6 +109,25 @@ private static boolean isUnimplementedMethod(StatusRuntimeException statusRuntim
         return statusRuntimeException.getStatus().getCode() == Status.Code.UNIMPLEMENTED;
     }
 
+    private static TypeDBClientException tryGetRstStreamCause(StatusRuntimeException sre) {
+        if (sre.getStatus().getCode() == Status.Code.UNAVAILABLE) {
+            if (sre.getCause() instanceof javax.net.ssl.SSLHandshakeException) {
+                if (sre.getCause().getMessage().contains("PKIX path building failed")) {
```
flyingsilverfin, line 115: oof okay is there no better way to do this rather than checking the strings :'( What if they change their messages

common/exception/TypeDBClientException.java
flyingsilverfin: This seems ok to me!

common/exception/TypeDBClientException.java
```diff
@@ -98,6 +109,25 @@ private static boolean isUnimplementedMethod(StatusRuntimeException statusRuntim
...
+                    return new TypeDBClientException(sre.getCause(), CLUSTER_SSL_CERTIFICATE_NOT_VALIDATED);
+                } else if (sre.getCause().getMessage().contains("No subject alternative DNS name")) {
+                    return new TypeDBClientException(sre.getCause(), CLUSTER_SSL_CERTIFICATE_INVALID_FOR_HOSTNAME);
+                } else return new TypeDBClientException(sre.getCause(), CLUSTER_SSL_HANDSHAKE_FAILED, sre.getMessage());
+            } else if (sre.getCause() instanceof io.netty.handler.ssl.NotSslRecordException) {
+                return new TypeDBClientException(sre.getCause(), CLUSTER_SERVER_NOT_ENCRYPTED);
+            } else if (sre.getStatus().getDescription() != null &&
+                    sre.getStatus().getDescription().contains("Network closed for unknown reason")) {
+                // Too generic a reason to throw. Wrap it in UNABLE_TO_CONNECT so the client tries other nodes.
+                return new TypeDBClientException(new TypeDBClientException(sre.getCause(), CLUSTER_CONNECTION_CLOSED_UKNOWN), UNABLE_TO_CONNECT);
```
flyingsilverfin, line 125: It's awkward to have an exception in an exception, can we directly use `CLOSED_UNKNOWN` alone?

flyingsilverfin (changes requested):

krishnangovindraj (commented):
replies & revisions

common/exception/TypeDBClientException.java
flyingsilverfin: Ok maybe you can update that one too ;)

common/exception/TypeDBClientException.java
flyingsilverfin: Sure, let's go with that

common/exception/TypeDBClientException.java
flyingsilverfin: I think this one you'd need to take up with Ganesh while im away or wait till next week

test/integration/ClusterEncryptionTest.java
flyingsilverfin: I think we should then be providing the (absolute?) path of the `dev-root-ca.pem` to use into the runner using the options the same way you're overriding the `server.encryption.enable` option

common/exception/TypeDBClientException.java
krishnangovindraj: Flagging to revisit

krishnangovindraj (commented):
Addressed most comments.

common/exception/ErrorMessage.java
```diff
@@ -66,6 +66,16 @@ public static class Client extends ErrorMessage {
...
                 new Client(19, "Expired password credential.");
+        public static final ErrorMessage CLUSTER_SERVER_NOT_ENCRYPTED =
+                new Client(20, "Unable to connect to TypeDB Cluster: attempting an encrypted connection to an unencrypted server.");
+        public static final ErrorMessage CLUSTER_SSL_HANDSHAKE_FAILED =
+                new Client(21, "SSL handshake with TypeDB Cluster failed: %s.");
+        public static final ErrorMessage CLUSTER_SSL_CERTIFICATE_NOT_VALIDATED =
+                new Client(22, "SSL handshake with TypeDB Cluster failed: the server's identity could not be verified.");
+        public static final ErrorMessage CLUSTER_SSL_CERTIFICATE_INVALID_FOR_HOSTNAME =
+                new Client(23, "SSL handshake with TypeDB Cluster failed: the server's certificate is not valid for the provided hostname.");
+        public static final ErrorMessage CLUSTER_CONNECTION_CLOSED_UKNOWN =
```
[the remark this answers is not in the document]
krishnangovindraj: We don't always have the information available where we encounter these problems.
What we can do is to catch these where expected and wrap them in an another UNABLE_TO_CONNECT_TO_NODE exception.

test/integration/ClientEncryptionTest.java
```diff
@@ -0,0 +1,138 @@
...
+    private static final TypeDBCredential MISCONFIGURED_CREDENTIALS = new TypeDBCredential(USERNAME, PASSWORD, WRONG_CA);
+    private static final TypeDBCredential UNENCRYPTED_CREDENTIALS = new TypeDBCredential(USERNAME, PASSWORD, false);
+
+    private TypeDBClusterRunner clusterRunner;
+
+    @After
+    public void shutDownCluster() {
+        if (clusterRunner != null) {
+            clusterRunner.stop();
+            clusterRunner.deleteFiles();
```
flyingsilverfin, line 76: Check if this is already included in the cluster runner cleanup

flyingsilverfin (approved):
Approved pending 1 investigation

test/integration/ClientEncryptionTest.java
krishnangovindraj: There's no cleanup.
Just a `reset` or a `deleteFiles`
This feels more explicit.

PR #432, opened 2023-07-26
Improve interaction with TypeDB cluster databases 

krishnangovindraj:
We refactor the the client's interaction with TypeDB cluster for more clarity. 
We also fix a bug so that opening a session to a non-existent database on TypeDB cluster fails immediately with a "database not found exception" rather than failing after 10 seconds with a connection error.

* Client is now explicitly aware of cluster's response when it has no replicator for a given database.
* Refactors `ClusterDatabaseManager` to runAnyReplica or runPrimaryReplica as required rather than relying on a utility function which always tries to runAnyReplica first, then runsPrimaryReplica if the node throws "Not primary replica"
- fixes vaticle/typedb-cluster#234

### DIFF
--- a/connection/cluster/ClusterClient.java
+++ b/connection/cluster/ClusterClient.java
@@ -41,10 +41,10 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
-import java.util.stream.Collectors;
 
 import static com.vaticle.typedb.client.common.exception.ErrorMessage.Client.CLUSTER_REPLICA_NOT_PRIMARY;
 import static com.vaticle.typedb.client.common.exception.ErrorMessage.Client.CLUSTER_UNABLE_TO_CONNECT;
+import static com.vaticle.typedb.client.common.exception.ErrorMessage.Client.DB_DOES_NOT_EXIST;
 import static com.vaticle.typedb.client.common.exception.ErrorMessage.Client.UNABLE_TO_CONNECT;
 import static com.vaticle.typedb.client.common.exception.ErrorMessage.Internal.UNEXPECTED_INTERRUPTION;
 import static com.vaticle.typedb.client.common.rpc.RequestBuilder.Cluster.DatabaseManager.getReq;
@@ -77,7 +77,7 @@ public class ClusterClient implements TypeDBClient.Cluster {
     }
 
     private Set<String> fetchCurrentAddresses(Set<String> servers) {
-        List<Pair<String, Throwable>> perServerExceptions = new ArrayList<>();
+        Map<String, Throwable> perServerExceptions = new HashMap<>();
         for (String server : servers) {
             try (ClusterServerClient client = new ClusterServerClient(server, credential, parallelisation)) {
                 client.validateConnection();
@@ -85,7 +85,7 @@ public class ClusterClient implements TypeDBClient.Cluster {
             } catch (TypeDBClientException e) {
                 if (UNABLE_TO_CONNECT.equals(e.getErrorMessage())) {
                     LOG.warn("Unable to fetch list of all servers from server {}.", server);
-                    if (e.getCause() != null) perServerExceptions.add(new Pair<>(server, e));
+                    if (e.getCause() != null) perServerExceptions.put(server, e);
                 } else {
                     throw e;
                 }
@@ -94,10 +94,11 @@ public class ClusterClient implements TypeDBClient.Cluster {
 
         String description;
         if (!perServerExceptions.isEmpty()) {
-            List<String> reasons = perServerExceptions.stream()
-                    .map(cause -> String.format("\t- %s: %s", cause.first(), cause.second().getCause().getMessage()))
-                    .collect(Collectors.toList());
-            description = "Reasons:[\n" + String.join("\n", reasons) + "\n]";
+            StringBuilder reasons = new StringBuilder("Reasons:[\n");
+            perServerExceptions.entrySet().stream()
+                    .map(entry -> String.format("\t- %s: %s\n", entry.getKey(), entry.getValue().getCause().getMessage()))
+                    .forEach(reasons::append);
+            description = reasons.append("]").toString();
         } else description = String.join(",", servers);
         throw new TypeDBClientException(CLUSTER_UNABLE_TO_CONNECT, description);
     }
@@ -222,7 +223,6 @@ public class ClusterClient implements TypeDBClient.Cluster {
         private final int PRIMARY_REPLICA_TASK_MAX_RETRIES = 10;
         private final int FETCH_REPLICAS_MAX_RETRIES = 10;
         private final int WAIT_FOR_PRIMARY_REPLICA_SELECTION_MS = 2000;
-
         private final String database;
 
         private FailsafeTask(String database) {
@@ -297,6 +297,8 @@ public class ClusterClient implements TypeDBClient.Cluster {
                 ClusterDatabase clusterDatabase = fetchDatabaseReplicas();
                 if (clusterDatabase.primaryReplica().isPresent()) {
                     return clusterDatabase.primaryReplica().get();
+                } else if (isNonExistentRaftClusterInfo(clusterDatabase)) {
+                    throw new TypeDBClientException(DB_DOES_NOT_EXIST, this.database);
                 } else {
                     waitForPrimaryReplicaSelection();
                     retries++;
@@ -312,7 +314,9 @@ public class ClusterClient implements TypeDBClient.Cluster {
                     ClusterDatabaseProto.ClusterDatabaseManager.Get.Res res = fetchValidatedServerClient(serverAddress)
                             .stub().databasesGet(getReq(database));
                     ClusterDatabase clusterDatabase = ClusterDatabase.of(res.getDatabase(), ClusterClient.this);
-                    clusterDatabases.put(database, clusterDatabase);
+                    if (!isNonExistentRaftClusterInfo(clusterDatabase)) {
+                        clusterDatabases.put(database, clusterDatabase);
+                    }
                     return clusterDatabase;
                 } catch (TypeDBClientException e) {
                     if (UNABLE_TO_CONNECT.equals(e.getErrorMessage())) {
@@ -332,6 +336,11 @@ public class ClusterClient implements TypeDBClient.Cluster {
             } catch (InterruptedException e) {
                 throw new TypeDBClientException(UNEXPECTED_INTERRUPTION);
             }
+        }
+
+        private boolean isNonExistentRaftClusterInfo(ClusterDatabase clusterDatabase) {
+            return clusterDatabase.primaryReplica().isEmpty() &&
+                    clusterDatabase.replicas().stream().allMatch(replica -> replica.term() == -1);
         }
 
         private ClusterServerClient fetchValidatedServerClient(String address) {
--- a/connection/cluster/ClusterDatabaseManager.java
+++ b/connection/cluster/ClusterDatabaseManager.java
@@ -59,15 +59,27 @@ public class ClusterDatabaseManager implements DatabaseManager.Cluster {
 
     @Override
     public boolean contains(String name) {
-        return failsafeTask(name, (stub, dbMgr) -> dbMgr.contains(name));
+        try {
+            return failsafeTask(name, ((stub, dbMgr) -> dbMgr.contains(name))).runPrimaryReplica();
+        } catch (TypeDBClientException e) {
+            if (e.getErrorMessage().equals(DB_DOES_NOT_EXIST)) return false;
+            else throw e;
+        }
     }
 
     @Override
     public void create(String name) {
-        failsafeTask(name, (stub, dbMgr) -> {
+        ClusterClient.FailsafeTask<Void> failsafeTask = failsafeTask(name, (stub, dbMgr) -> {
             dbMgr.create(name);
             return null;
         });
+        try {
+            failsafeTask.runAnyReplica();
+        } catch (TypeDBClientException e) {
+            if (CLUSTER_REPLICA_NOT_PRIMARY.equals(e.getErrorMessage())) {
+                failsafeTask.runPrimaryReplica();
+            } else throw e;
+        }
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


connection/cluster/ClusterClient.java
```diff
@@ -322,14 +336,19 @@ private void waitForPrimaryReplicaSelection() {
             }
         }
 
+        private boolean isNonExistentRaftClusterInfo(ClusterDatabase clusterDatabase) {
+            return clusterDatabase.primaryReplica().isEmpty() &&
+                    clusterDatabase.replicas().stream().allMatch(replica -> replica.term() == -1);
+        }
```
krishnangovindraj, line 342: When a replicator does not exist on a node, it returns an entry with no primary and all terms set to -1.

connection/cluster/ClusterDatabaseManager.java
```diff
@@ -59,25 +59,58 @@ public ClusterDatabaseManager(ClusterClient client) {
...
+                    name,
+                    parameter -> task.apply(
+                            parameter.client().stub(),
+                            parameter.client().databases()
+                    )
+            ).runPrimaryReplica();
+        } catch (TypeDBClientException e) {
+            if (e.getErrorMessage().equals(DB_DOES_NOT_EXIST)) return false;
+            else throw e;
+        }
```
krishnangovindraj, line 74: Gets rid of the failSafeTask wrapper function (below in red). The wrapper would try running the task against all replicas. If it then receives a NOT_PRIMARY_REPLICA` exception, it retries on the primary replica.

We update each task to go to be more explicit in where it should run.

connection/cluster/ClusterDatabaseManager.java
```diff
@@ -59,25 +59,58 @@ public ClusterDatabaseManager(ClusterClient client) {
...
+                        parameter.client().databases()
+                )
+        );
+        try {
+            failsafeTask.runAnyReplica();
+        } catch (TypeDBClientException e) {
+            if (CLUSTER_REPLICA_NOT_PRIMARY.equals(e.getErrorMessage())) {
+                failsafeTask.runPrimaryReplica();
+            } else throw e;
+        }
```
krishnangovindraj, line 96: **If the database never existed**, there is no primary, so we must run it on any replica. This may fail in two ways:
1. While creating replicators on each node: 
2. When writing the database `CREATE` management record to the raft log. 
* If this is caused by the (highly probable) scenario where the node the client is connected not being voted primary (for the newly created replica), we transparently retry on whichever node is the primary.

In all other cases, We require the user to retry. This is safe because database creation is idempotent.

If the database did exist, we should be running create on the primary.
The runAnyReplica call will fail with `CLUSTER_REPLICA_NOT_PRIMARY` and retry on the primary.

connection/cluster/ClusterClient.java
```diff
@@ -294,13 +304,17 @@ private ClusterDatabase.Replica seekPrimaryReplica() {
...
+                String serverAddress = clusterAddresses.get((roundRobinIndex + i) % clusterAddresses.size());
                 try {
                     LOG.debug("Fetching replica info from {}", serverAddress);
                     ClusterDatabaseProto.ClusterDatabaseManager.Get.Res res = fetchValidatedServerClient(serverAddress)
                             .stub().databasesGet(getReq(database));
                     ClusterDatabase clusterDatabase = ClusterDatabase.of(res.getDatabase(), ClusterClient.this);
-                    clusterDatabases.put(database, clusterDatabase);
+                    if (!isNonExistentRaftClusterInfo(clusterDatabase)) {
+                        clusterDatabases.put(database, clusterDatabase);
+                    }
```
krishnangovindraj, line 317: Don't cache nonExistentRaftClusterInfo responses.

connection/cluster/ClusterClient.java
```diff
@@ -285,6 +293,8 @@ private ClusterDatabase.Replica seekPrimaryReplica() {
                 ClusterDatabase clusterDatabase = fetchDatabaseReplicas();
                 if (clusterDatabase.primaryReplica().isPresent()) {
                     return clusterDatabase.primaryReplica().get();
+                } else if (isNonExistentRaftClusterInfo(clusterDatabase)) {
+                    throw new TypeDBClientException(DB_DOES_NOT_EXIST, this.database);
```
krishnangovindraj, line 297: If any of the nodes does not have a replicator, the database does not exist.

In the cluster code: The typedb database is created only when the a database `CREATE` record is committed to the raft log. This is appended to raft log only in ClusterDatabaseManager.create only if creating the replicator succeeded on all nodes. ClusterDatabaseManager.create which is only called through TypeDBService - Hence run on the node which is processing the client request.

dependencies/vaticle/repositories.bzl
```diff
@@ -52,8 +52,8 @@ def vaticle_typedb_protocol():
 def vaticle_typedb_behaviour():
     git_repository(
         name = "vaticle_typedb_behaviour",
-        remote = "https://github.com/vaticle/typedb-behaviour",
-        commit = "90b4addcd71a398e9e52e322021c49310e66a47a" # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
+        remote = "https://github.com/krishnangovindraj/typedb-behaviour",
+        commit = "0dee0f7785416edd441f02be0df53572ff55df4d" # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
```
krishnangovindraj: enables the database delete test on a non-existent database.

dependencies/vaticle/repositories.bzl
```diff
@@ -52,8 +52,8 @@ def vaticle_typedb_protocol():
 def vaticle_typedb_behaviour():
     git_repository(
         name = "vaticle_typedb_behaviour",
-        remote = "https://github.com/vaticle/typedb-behaviour",
-        commit = "90b4addcd71a398e9e52e322021c49310e66a47a" # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
+        remote = "https://github.com/krishnangovindraj/typedb-behaviour",
+        commit = "14a2e9843ea4757d9c10d78a137881d49f9ac79e" # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
```
krishnangovindraj, line 56: test for deleting non existent database + 
create->delete->create test

krishnangovindraj (commented):
self-review

connection/cluster/ClusterClient.java
```diff
@@ -75,19 +79,22 @@ public ClusterClient(Set<String> initAddresses, TypeDBCredential credential, int
...
+                    LOG.warn("Unable to fetch list of all servers from server {}.", server);
+                    if (e.getCause() != null) causes += "\t* " + server + ": " + e.getCause().getMessage() + "\n";
                 } else {
                     throw e;
                 }
             }
         }
-        throw new TypeDBClientException(CLUSTER_UNABLE_TO_CONNECT, String.join(",", servers));
+        String description = causes.isBlank() ? String.join(",", servers) : "Reasons:[\n" + causes + "]";
+        throw new TypeDBClientException(CLUSTER_UNABLE_TO_CONNECT, description);
```
krishnangovindraj, line 97: More descriptive but very adhoc. Please suggest improvements.

krishnangovindraj (commented):
Flag ad-hoc fix for review.

flyingsilverfin:
Given how nuanced this is I think I'd prefer to review this in person. Is it ok if we do it monday/tuesday?

krishnangovindraj:
I think Dmitrii can do at least the first few rounds of this one. He's got a pretty good picture from his time with the rust client.

connection/cluster/ClusterClient.java
```diff
@@ -210,11 +217,12 @@ abstract class FailsafeTask<RESULT> {
         private final int PRIMARY_REPLICA_TASK_MAX_RETRIES = 10;
         private final int FETCH_REPLICAS_MAX_RETRIES = 10;
         private final int WAIT_FOR_PRIMARY_REPLICA_SELECTION_MS = 2000;
-
         private final String database;
+        private int roundRobinIndex;
```
dmitrii-ubskii, line 221: This should not be per `FailsafeTask`. As it stands, we always begin by querying the first address in the list.

connection/cluster/ClusterClient.java
dmitrii-ubskii: Do we need the first check here?

connection/cluster/ClusterClient.java
```diff
@@ -294,13 +304,17 @@ private ClusterDatabase.Replica seekPrimaryReplica() {
         }
 
         private ClusterDatabase fetchDatabaseReplicas() {
-            for (String serverAddress : clusterServerClients.keySet()) {
+            roundRobinIndex = (roundRobinIndex + 1) % clusterAddresses.size();
+            for (int i=0 ; i < clusterAddresses.size() ; i++) {
+                String serverAddress = clusterAddresses.get((roundRobinIndex + i) % clusterAddresses.size());
```
dmitrii-ubskii, line 309: It might be worth it to also advance the round robin index on failure here. Otherwise, if node 0 is unavailable, we effectively query node 1 twice before querying node 2.

dmitrii-ubskii (changes requested):

connection/cluster/ClusterClient.java
krishnangovindraj: May not be, but it's a tighter condition.

connection/cluster/ClusterDatabaseManager.java
krishnangovindraj: Unsure about doing any running within the function. For brevity, I've reintroduced `failSafeTask` which only creates it but does no running. Please re-review.

connection/cluster/ClusterClient.java
```diff
@@ -67,6 +70,7 @@ public ClusterClient(Set<String> initAddresses, TypeDBCredential credential, int
         this.credential = credential;
         this.parallelisation = parallelisation;
         Set<String> currAddresses = fetchCurrentAddresses(initAddresses);
+        clusterAddresses = new ArrayList<>(initAddresses);
```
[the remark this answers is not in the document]
krishnangovindraj: Eep. Thanks!
Updated.

.bazelrc
```diff
@@ -21,7 +21,7 @@
 
 try-import ./.bazel-remote-cache.rc
 
-build --incompatible_strict_action_env --java_language_version=11 --javacopt='--release 11'
-run --incompatible_strict_action_env
-test --incompatible_strict_action_env --test_env=PATH --cache_test_results=no
+build --incompatible_strict_action_env --java_language_version=11 --javacopt='--release 11' --java_runtime_version=remotejdk_11
```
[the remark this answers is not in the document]
krishnangovindraj: Works with it, Doesn't work without it :(

connection/cluster/ClusterClient.java
krishnangovindraj: After discussion, we decided to leave this here so a single task's `fetchDatabaseReplicas` retries are guaranteed to go roundRobin without interfering with each other.

krishnangovindraj (commented):
Responses & changes.

connection/cluster/ClusterDatabaseManager.java
dmitrii-ubskii: That's even better, nicely done.

connection/cluster/ClusterClient.java
dmitrii-ubskii: To clarify, the existence of a preferred replica makes this a non-issue.

connection/cluster/ClusterClient.java
```diff
@@ -315,12 +314,14 @@ private ClusterDatabase fetchDatabaseReplicas() {
...
                         clusterDatabases.put(database, clusterDatabase);
                     }
+                    roundRobinIndex = (roundRobinIndex + i + 1) % clusterAddresses.size();
                     return clusterDatabase;
                 } catch (TypeDBClientException e) {
                     if (UNABLE_TO_CONNECT.equals(e.getErrorMessage())) {
                         LOG.debug("Failed to fetch replica info for database '" + database + "' from " +
                                 serverAddress + ". Attempting next server.", e);
                     } else {
+                        roundRobinIndex = (roundRobinIndex + i + 1) % clusterAddresses.size();
```
dmitrii-ubskii, line 324: This would arguably be simpler if we had a `finally { roundRobinIndex = (roundRobinIndex + 1) % clusterAddresses.size(); }`. Then we can simply `clusterAddresses.get(roundRobinIndex);` at the start of each iteration.

dmitrii-ubskii (approved):
Minor comments. Keep in mind that the repositories still link to your fork of behaviour!

connection/cluster/ClusterClient.java
krishnangovindraj: Fair. Doesn't this suffer from the "Querying node 1 twice in a row, if node 0 is unreachable" problem?

krishnangovindraj: Updated.
I think I shied away from it initially  because it had two variables being updated independently in a loop.

connection/cluster/ClusterClient.java
```diff
@@ -75,19 +81,29 @@ public ClusterClient(Set<String> initAddresses, TypeDBCredential credential, int
     }
 
     private Set<String> fetchCurrentAddresses(Set<String> servers) {
+        List<Pair<String, Throwable>> perServerExceptions = new ArrayList<>();
```
flyingsilverfin, line 84: Map

flyingsilverfin (approved):

connection/cluster/ClusterClient.java
```diff
@@ -312,7 +314,9 @@ private ClusterDatabase fetchDatabaseReplicas() {
                     ClusterDatabaseProto.ClusterDatabaseManager.Get.Res res = fetchValidatedServerClient(serverAddress)
                             .stub().databasesGet(getReq(database));
                     ClusterDatabase clusterDatabase = ClusterDatabase.of(res.getDatabase(), ClusterClient.this);
-                    clusterDatabases.put(database, clusterDatabase);
+                    if (!isNonExistentRaftClusterInfo(clusterDatabase)) {
+                        clusterDatabases.put(database, clusterDatabase);
+                    }
```
krishnangovindraj, line 319: (anchor for conversation about rust driver)

PR #424, opened 2023-06-23
Test datetime attribute insert and match time-zone invariant

shiladitya-mukherjee:
## What is the goal of this PR?

Added step implementation for changing environment timezone for Java, and fixed the parsing of datetime. We hence enabled the BDD tests to check timezone-invariance of inserting and reading datetime attributes.

## What are the changes implemented in this PR?

- Added "set time-zone is: {time_zone_label}" which sets environment timezone to time_zone_label.
- Fixed datetime attribute parsing by bounding the number of splits of the attribute by ":" to 2.

PR #434, opened 2023-08-02
Implement expressions and top-level values in client-rust

dmitrii-ubskii:

We update the Rust driver to support the features introduced in driver Java since v2.19 in effort to preserve the feature set when transitioning from the JVM-native implementation to the Rust JNI implementation.

Changes in master since branching (#417):

Reimplemented in Rust and made available in Java over JNI:
- #409 
- #421 
- 1f396a6 Improve method unavailable error message
- #430 
- #431: partial, since `tonic` does not report SSL errors to the same granularity 

Cherry-picked directly:
- #415
- 42800e7 Update VERSION to 2.18.1
- 7cd0a5a Add eclipsesource-minimal-json to maven dependencies (#423)
- 63614ec Update CODEOWNERS
- #424

Dropped entirely (Java-specific or obsolete):
- #422 
- 8c0caa1 Update VERSION and release notes


java/concept/ConceptImpl.java
```diff
@@ -52,7 +55,9 @@ protected ConceptImpl(com.vaticle.typedb.client.jni.Concept concept) {
     }
 
     protected static com.vaticle.typedb.client.jni.Transaction nativeTransaction(TypeDBTransaction transaction) {
-        return ((ConceptManagerImpl) transaction.concepts()).nativeTransaction;
+        com.vaticle.typedb.client.jni.Transaction nativeTransaction = ((ConceptManagerImpl) transaction.concepts()).nativeTransaction;
+        if (!nativeTransaction.isOwned()) throw new TypeDBClientException(TRANSACTION_CLOSED);
+        return nativeTransaction;
```
flyingsilverfin, line 60: Let's follow the same methodology as in `client-python`

java/concept/ConceptManagerImpl.java
```diff
@@ -83,78 +85,127 @@ public AttributeType getRootAttributeType() {
...
         if (label == null || label.isEmpty()) throw new TypeDBClientException(MISSING_LABEL);
-        com.vaticle.typedb.client.jni.Concept res = concepts_get_entity_type(nativeTransaction, label);
-        if (res != null) return new EntityTypeImpl(res);
-        else return null;
+        if (!nativeTransaction.isOwned()) throw new TypeDBClientException(TRANSACTION_CLOSED);
+        try {
+            com.vaticle.typedb.client.jni.Concept res = concepts_get_entity_type(nativeTransaction, label);
+            if (res != null) return new EntityTypeImpl(res);
+            else return null;
+        } catch (com.vaticle.typedb.client.jni.Error e) {
```
flyingsilverfin, line 93: Can we have our `TypeDBClientException` extend the JNI error so we don't have to catch and re-throw all the time?

java/logic/RuleImpl.java
```diff
@@ -69,17 +71,35 @@ public ThingVariable<?> getThen() {
...
+            return rule_is_deleted(nativeTransaction(transaction), nativeObject);
+        } catch (com.vaticle.typedb.client.jni.Error e) {
+            throw new TypeDBClientException(e);
+        }
+    }
+
+    private static com.vaticle.typedb.client.jni.Transaction nativeTransaction(TypeDBTransaction transaction) {
+        com.vaticle.typedb.client.jni.Transaction nativeTransaction = ((LogicManagerImpl) transaction.logic()).nativeTransaction;
+        if (!nativeTransaction.isOwned()) throw new TypeDBClientException(TRANSACTION_CLOSED);
+        return nativeTransaction;
```
flyingsilverfin, line 102: In the new model we decided in Python, we can keep this method but have it get the `this.nativeObject()` with all the projection that entails. This means we keep the `transaction` as the nativeObject, and 'rename' it via a simple getter that redirects.

rust/src/common/error.rs
```diff
@@ -28,34 +28,44 @@ use super::{address::Address, RequestID};
...
     ClusterTokenCredentialInvalid() =
-        16: "Invalid token credential.",
+        17: "Invalid token credential.",
     SessionCloseFailed() =
-        17: "Failed to close session. It may still be open on the server: or it may already have been closed previously.",
+        18: "Failed to close session. It may still be open on the server: or it may already have been closed previously.",
+    ClusterEndpointEncrypted() =
+        19: "Unable to connect to TypeDB Cluster: attempting an unencrypted connection to an encrypted endpoint.",
+    ClusterSSLCertificateNotValidated() =
+        20: "SSL handshake with TypeDB Cluster failed: the server's identity could not be verified.",
```
flyingsilverfin, line 64: Can we write something easier to action in this message? Does it mean that the root CA does not match, or something else?

java/concept/ConceptManagerImpl.java
dmitrii-ubskii: That would require the user to catch `com.vaticle.typedb.client.jni.Error` instead of `com.vaticle.typedb.client.common.exception.TypeDBClientException`, which leaks implementation details.

Unless you meant the other way around (`jni.Error extends TypeDBClientException`), which would be friendlier, but not possible at the moment as that would involve a circular dependency. We could do that at some point by extracting the client error into its own top-level common package (or into typedb-common).

rust/src/common/error.rs
dmitrii-ubskii: > Does it mean that the root CA does not match

I believe so. How about "SSL handshake with TypeDB Cluster failed: the server's identity could not be verified. Possible CA mismatch."?

flyingsilverfin: Sure, that sounds good

java/concept/ConceptManagerImpl.java
flyingsilverfin: Let's keep this for now.

flyingsilverfin (approved):

PR #436, opened 2023-08-10
Python-specific conversions for SWIG

dmikhalin:

We define python-specific conversions for SWIG for arrays, exceptions and `ValueType`.


java/test/behaviour/typeql/language/insert/BUILD
```diff
@@ -55,7 +55,7 @@ typedb_behaviour_java_test(
     runtime_deps = [
         "//java/test/behaviour/config:parameters",
     ],
-    size = "large",
+    size = "enormous",
```
flyingsilverfin, line 58: Why has this gotten larger? Are the tests slower?

rust/swig/typedb_client_python.swg
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+%pythoncode %{
+    TypeDBClientException = native_client_python.TypeDBClientException
+%}
+
+/* ValueType* needs special handling */
+%typemap(in) enum ValueType * {
+    if ($input != Py_None) {
```
flyingsilverfin, line 105: How does swig translate `none`? Is this the protection we built for the issue we had found where we passed a null and it got mapped to `object` implicitly?

java/test/behaviour/typeql/language/insert/BUILD
dmikhalin: There was a TL at Factory (900 s).

rust/swig/typedb_client_python.swg
dmikhalin: The main problem was with all values except `None`. Now we allocate memory for all such `ValueType`s (and free it later), and the value remains `null` if it was `None` in python.

java/test/behaviour/typeql/language/insert/BUILD
flyingsilverfin: How close was it to time out before we changed to the rust-wrapped client? We should always know the root cause of why we're increasing test timeouts

rust/swig/typedb_client_python.swg
flyingsilverfin: ok, got it. i hope we don't introduce any memory leaks!!

java/test/behaviour/typeql/language/insert/BUILD
dmikhalin: When the previous PR was tested on Factory, this test lasted for 830 s. The last attempt for this PR took 757 s. So it is close to 900 and sometimes occasionally take longer.

flyingsilverfin: Hmm ok, we can make it larger

PR #445, opened 2023-09-19
Finalize deployment to Cargo and NPM

dmitrii-ubskii:

We add deployment to Cargo and NPM, along with post-release cleanup from CircleCI. We also fix various NPM tests, which require encryption certificates to be set up.

## Changes in this PR

- Finish release pipeline for NPM, NPM Snapshot, and Cargo
- Fix enterprise tests that required an encrypted network connection (notably, Client NodeJS)


nodejs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "typedb-client",
-  "version": "2.11.1",
+  "version": "0.0.0",
```
dmitrii-ubskii, line 3: This version was outdated in the first place, and it gets overridden by `assemble_npm` anyway. Having any version other than a placeholder is misleading.

PR #451, opened 2023-09-26
Allow on_close callbacks to interact with the network

dmitrii-ubskii:

Previously, if the user were to submit a `transaction.on_close()` callback that attempted to perform a network request (e.g., open a new transaction on failure), the networking thread would deadlock: the transactiopn stream would block the async runtime waiting for the request to resolve by that same runtime.

As a solution, we introduce a worker thread that can handle the blocking callbacks, and await on them asynchronously. This way, `on_close()` will still not progress until the callbacks resolve, but will now be able to yield to the other tasks running on the async runtime. 

- callbacks are persisted s.t. they are not garbage collected before being executed;
- callbacks are executed in a non-blocking fashion so that they cannot block the networking thread;
- drive-by: getSupertype() now returns an optional since the root types have no supertype.

### DIFF
--- a/rust/src/connection/connection.rs
+++ b/rust/src/connection/connection.rs
@@ -47,7 +47,7 @@ use crate::{
         address::Address,
         error::{ConnectionError, Error},
         info::{DatabaseInfo, SessionInfo},
-        Result, SessionID, SessionType, TransactionType,
+        Callback, Result, SessionID, SessionType, TransactionType,
     },
     connection::message::{Request, Response, TransactionRequest},
     error::InternalError,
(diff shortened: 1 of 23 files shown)


rust/src/connection/connection.rs
```diff
@@ -55,37 +60,60 @@ use crate::{
 pub struct Connection {
     server_connections: HashMap<Address, ServerConnection>,
     background_runtime: Arc<BackgroundRuntime>,
+
+    callback_handler: Arc<JoinHandle<()>>,
```
dmitrii-ubskii, line 64: TODO

flyingsilverfin (approved):

PR #498, opened 2023-10-13
Wrap native exceptions with python driver exception

dmikhalin:

We had two exception classes: `TypeDBDriverException`, that could be raised by native calls, and `TypeDBDriverExceptionExt(TypeDBDriverException)`, that was raised by python driver methods. In order to catch all driver exceptions we had to use `TypeDBDriverException`, but this class couldn't be commented properly, because it is defined in the SWIG-generated file, that we wouldn't like to expose.

Now all raised exceptions are `TypeDBDriverException`s, and this class is properly documented.

We catch all native TypeDB exceptions, that could be thrown by native functions calls, and raise `TypeDBDriverException` (formerly known as `TypeDBDriverExceptionExt`).

### DIFF
--- a/python/typedb/common/exception.py
+++ b/python/typedb/common/exception.py
@@ -26,7 +26,7 @@ from typing import Union, Any
 from typedb.native_driver_wrapper import TypeDBDriverExceptionNative
 
 
-class TypeDBDriverException(TypeDBDriverExceptionNative):
+class TypeDBDriverException(RuntimeError):
     """
     Exceptions raised by the driver.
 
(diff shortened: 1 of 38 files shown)


python/typedb/common/exception.py
```diff
@@ -40,26 +38,25 @@ class TypeDBDriverExceptionExt(TypeDBDriverException):
         try:
             transaction.commit()
         except TypeDBDriverException as err:
-            println("Error:", err)
+            print("Error:", err)
     """
 
-    def __init__(self, msg: Union[ErrorMessage, str], cause: BaseException = None, params: Any = None):
+    def __init__(self, msg: Union[ErrorMessage, str], params: Any = None):
```
dmikhalin, line 44: We haven't use `cause` anywhere, so we removed it and replaced old usages of `of()` with `__init__()`. New `of()` is used for wrapping native exceptions.

flyingsilverfin (approved):

PR #500, opened 2023-10-16
Fix filenames with incorrect symbols for Windows

dmikhalin:

We had some filenames in generated AsciiDoc files that are prohibited in Windows. It affected only NodeJS Driver, but cloning the repo was impossible in Windows. We fixed it.

We replaced symbols `"<"` and `">"` with underscores.

We also fixed the script for docs testing: now we add all new generated files to git.


nodejs/docs/Stream<T>.adoc
krishnangovindraj, line 1: Since this file is being deleted, shouldn't a file named `Stream_T_.adoc` be added?

dmikhalin: Definitely. And it should be tested, but somehow it isn't...

krishnangovindraj (approved):
Looks good!

flyingsilverfin:
We should regenerate the docs after this change no?

krishnangovindraj:
@flyingsilverfin I think he did regenerate the docs. If he hadn't, the test would fail.

dmikhalin:
> We should regenerate the docs after this change no?

@flyingsilverfin We have done it. But only one file is affected.

PR #502, opened 2023-10-18
Move CircleCI jobs for MacOS x86_64 drivers to run on mac arm64 machines

krishnangovindraj:
We reconfigure our CircleCI pipeline to use MacOS arm64 machines for running build & test jobs for our TypeDB MacOS x86_64 drivers. This is in light of [CircleCI sunsetting MacOS (intel) x86_64 resources in January 2024](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718).

Installing rosetta & using an x86_64 version of bazel was enough for bazel to build & execute targets in x86_64 mode. Additionally, we:
* Make java processes spawned outside of bazel use the same java installation as bazel
* Bump `typedb-common` for a change setting TypeDB runners to use the same java as the test process.
* Replace jobs using the `mac-x86_64` executor to use the `mac-arm64` executor

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,7 @@ version: 2.1
 
 orbs:
   win: circleci/windows@2.0.0
+  macos: circleci/macos@2.4.0
 
 executors:
   linux-arm64:
@@ -40,11 +41,6 @@ executors:
     macos:
       xcode: "13.4.1"
     resource_class: macos.m1.medium.gen1
-    working_directory: ~/typedb-driver
-
-  mac-x86_64:
-    macos:
-      xcode: "13.4.1"
     working_directory: ~/typedb-driver
 
 
@@ -65,8 +61,14 @@ commands:
           chmod a+x /usr/local/bin/bazel
 
   install-bazel-mac:
+    parameters:
+      arch:
+        type: string
     steps:
-      - run: brew install bazelisk
+      - run: |
+          curl -OL "https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazelisk-darwin-<<parameters.arch>>"
+          sudo mv "bazelisk-darwin-<<parameters.arch>>" /usr/local/bin/bazel
+          chmod a+x /usr/local/bin/bazel
 
   ###########################
   # Python deployment steps #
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -23,6 +23,7 @@ version: 2.1
 
 orbs:
   win: circleci/windows@2.0.0
+  macos: circleci/macos@2.4.0
```
krishnangovindraj, line 26: Used for rosetta

tool/test/start-core-server.sh
```diff
@@ -25,4 +25,5 @@ set -e
 rm -rf typedb-all
 
 bazel run //tool/test:typedb-extractor -- typedb-all
+JAVA_HOME=$(bazel run //tool/test:echo-java-home)
```
krishnangovindraj, line 28: set `JAVA_HOME` for typedb process launched below.

This feels a little flimsy, since it points inside a bazel folder. e.g: 
`/private/var/tmp/_bazel_krishnangovindraj/5096250ab346cdc7499cfe043de9e352/external/remotejdk11_macos_aarch64/zulu-11.jdk/Contents/Home`

.circleci/config.yml
krishnangovindraj, line 1: Once common is merged, I'd like to try to replace the typedb start commands in here with the scripts in `/tool/test`

.circleci/config.yml
flyingsilverfin: Which xcode version will this contain? We want to make sure we can support a relatively old version of mac

.circleci/config.yml
```diff
@@ -434,12 +465,15 @@ jobs:
...
+          arch: arm64
 
   deploy-maven-jni-snapshot-mac-x86_64:
-    executor: mac-x86_64
+    executor: mac-arm64
     steps:
-      - deploy-maven-jni-snapshot-mac
+      - macos/install-rosetta
+      - deploy-maven-jni-snapshot-mac:
+          arch: amd64
```
flyingsilverfin, line 476: This is now horribly confusing to read but nothing we can do I guess!

Can we roll the `mac-x86_64` jobs into the same jobs as for `mac-arm64` and just call it `<>-mac-all`? This would save us a large set of jobs.

krishnangovindraj: We can rename the Mac executor to be `mac-all`.
I doubt we gain anything from trying to combine the jobs since they have to be built for their architecture all the way down.

.circleci/config.yml
krishnangovindraj: It's pinned to `13.4.1` in the executor declaration.

flyingsilverfin (approved):

PR #504, opened 2023-10-20
Introduce C++ driver

krishnangovindraj:
Introduce the C++ driver for TypeDB. It is built against the C++17 standard and distributed as an archive containing the headers (under `/include` & a shared library under `/lib`). 

**Usage:** As usual, add the headers paths to your include path in the compile step & the library to your link step. For windows, the 'import-lib' `typedb-driver-cpp-<platform>.if.lib` is included to link against.

**Architecture:** The C++ driver is a thin wrapper around the TypeDB rust driver, introducing classes for a more intuitive interface. Each C++ object holds a unique pointer to the corresponding native rust object and is the unique owner of that rust object. To ensure this, we enforce move-semantics on the C++ objects. The rust object is freed when the C++ object owning it is destructed. Any error encountered will throw a `TypeDB::DriverException`. Note that methods which return `Iterable` or `Future` which encounter a server-side error will only throw when they are evaluated (using `begin` or `get` respectively).

**Example:**
```cpp
// All files are included from typedb.hpp
#include <typedb.hpp>

int main() {
    TypeDB::Driver driver = TypeDB::Driver::coreDriver("127.0.0.1:1729");
    std::string dbName = "move-example";
    TypeDB::Driver driver = TypeDB::Driver::coreDriver("127.0.0.1:1729");

    driver.databases.create(dbName); 
    TypeDB::Database db = driver.databases.get(dbName);
    // Database db1 = db; // Copying is disabled: Produces a compiler error.
    TypeDB::Database db1 = std::move(db); // Moves ownership from db to db1
    try {
        std::cout << db.name() << std::endl; // db is no longer valid
    } catch (TypeDB::DriverException e) {
        // C++ Internal Error: The object does not have a valid native handle. It may have been:  uninitialised, moved or disposed
        std::cerr << "Caught exception: " << e.message() << std::endl;
    }
    std::cout << "db.name(): " << db1.name() << std::endl; // Ok: Prints 'move-objects'
    return 0;
}
```

**Driver:**
* Introduces the C++ driver, wrapping the C driver produced by rust.
* Relies on using `std::unique_ptr` and move semantics to reduce the risk of memory leaks & dangling pointers to native (rust) objects.

**Behaviour tests:**
* Introduces a minimal test framework which uses cucumber's gherkin as parser for feature files & gtest as test framework.
* Implements the steps for driver behaviour tests.


cpp/lib/connection/TypeDBDriver.cpp
```diff
@@ -0,0 +1,40 @@
...
+TypeDBDriver::TypeDBDriver(const std::string& coreAddress) : TypeDBDriver(TypeDBNative::connection_open_core(coreAddress.c_str())) { }
+
+TypeDBDriver::TypeDBDriver(TypeDBNative::Connection* conn) noexcept
+    : connectionNative(conn),
+      databaseManager(this->connectionNative),
+      userManager(this->connectionNative) {}
+
+TypeDBDriver::~TypeDBDriver() {
+    // TODO: connection_close is called before the databaseManager & UserManager are destructed. Should we wrap it in an object to enforce the order?
+    TypeDBNative::connection_close(connectionNative);
```
krishnangovindraj, line 37: Do you think this (the TODO) will be a problem? Closing the native connection and then the databaseManager & userManager?
Do you think it's worth introducing a really thin struct (possibly through a macro) just for this?

cpp/lib/common/TypeDBDriverException.cpp
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+#include "typedb/common/TypeDBDriverException.hpp"
+
+namespace TypeDB {
+
+TypeDBDriverException::TypeDBDriverException(const char* code, const char* message)
+    : std::runtime_error(""),
+      errCode(code),
+      errMsg(message) {}
```
krishnangovindraj, line 29: Copy over the native exception's char* so ownership is simplified.

krishnangovindraj (commented):
self-review comments

cpp/include/typedb/user/UserManager.hpp
```diff
@@ -0,0 +1,37 @@
...
+namespace TypeDB {
+
+class UserManager {
+    private:
+     TypeDBNative::UserManager* userManagerNative;
+
+    public:
+     UserManager(const TypeDBNative::Connection*);
+     UserManager(const UserManager&) = delete;
+     ~UserManager();
```
dmitrii-ubskii, line 34: Rule of five violation: https://en.cppreference.com/w/cpp/language/rule_of_three
```suggestion
     UserManager(const UserManager&) = delete;
     UserManager& operator=(const UserManager&) = delete;
     UserManager(UserManager&&) = default;
     UserManager& operator=(UserManager&&) = default;
     ~UserManager();
```

cpp/lib/common/TypeDBDriverException.cpp
dmitrii-ubskii: This class should be the one responsible for retrieving the exception contents from over FFI.

cpp/lib/common/utils.hpp
```diff
@@ -0,0 +1,39 @@
...
+namespace TypeDB {
+
+void check_and_throw() {
+    if (TypeDBNative::check_error()) {
+        TypeDBNative::Error* error = TypeDBNative::get_last_error();
+        char* errcode = TypeDBNative::error_code(error);
+        char* errmsg = TypeDBNative::error_message(error);
+        TypeDBDriverException exception(errcode, errmsg);
+        free(errmsg);
+        free(errcode);
```
dmitrii-ubskii, line 33: You shouldn't just `free()` memory that is managed by a different allocator. Use `string_free()` that's provided by the FFI layer.

dmitrii-ubskii: http://jakegoulding.com/rust-ffi-omnibus/string_return/

cpp/lib/connection/TypeDBDriver.cpp
dmitrii-ubskii: It won't be a problem, there's no lifetime dependency between those three.

cpp/include/typedb/connection/DatabaseManager.hpp
```diff
@@ -0,0 +1,41 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#pragma once
+
+#include <string>
+
+#include "typedb/common/native.hpp"
```
dmitrii-ubskii, line 25: The native header should _only_ be used in implementation files to avoid leaking it into the larger namespace. Where it is used as a field, such as here, you can use a pimpl pattern:
```cpp
// *.hpp
class DatabaseManager {
    // ...
    private:
        struct Impl;
        std::unique_ptr<Impl> pImpl;  // or Impl* pImpl;
};

// *.cpp
#include "typedb/common/native.hpp"

struct DatabaseManager::Impl {
    TypeDBNative::DatabaseManager* native;
    ~Impl() { database_manager_drop(native); } // if using `unique_ptr<Impl>`
}
```
More info: https://en.cppreference.com/w/cpp/language/pimpl

cpp/include/typedb/connection/TypeDBDriver.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include "typedb/common/native.hpp"
+#include "typedb/connection/DatabaseManager.hpp"
+#include "typedb/user/UserManager.hpp"
+
+#include <string>
+
+namespace TypeDB {
+
+class TypeDBDriver {
```
dmitrii-ubskii, line 31: `TypeDB::Driver`, not `TypeDB::TypeDBDriver`

cpp/include/typedb/connection/TypeDBDriver.hpp
```diff
@@ -0,0 +1,52 @@
...
+    const UserManager userManager;
+
+   private:
+    TypeDBDriver(TypeDBNative::Connection* conn) noexcept;
+
+   public:
+    TypeDBDriver(const std::string &coreAddress);
+    TypeDBDriver(const TypeDBDriver&) = delete;
+    ~TypeDBDriver();
+};
```
dmitrii-ubskii, line 51: Let's not interleave public and private blocks like this.

dmitrii-ubskii (changes requested):

cpp/include/typedb/connection/DatabaseManager.hpp
krishnangovindraj: They're being leaked, but not into the TypeDB namespace. Just into the TypeDBNative namespace.

This does make sense, but I guess the pImpl will always be on the heap then? Would we still want to do it for classes that we're instantiating plenty of times (e.g. Concept)?

krishnangovindraj: After a coffee table conversation, we decided we'd avoid PIMPL, keep things simple and import the native symbols into the nested namespace `TypeDB::_native` instead.

cpp/include/typedb/answer/conceptmap.hpp
```diff
@@ -0,0 +1,60 @@
...
+   public:
+    ConceptMap(_native::ConceptMap*);
+    ConceptMap(const ConceptMap&) = delete;
+    ConceptMap(ConceptMap&&);
+    
+    ConceptMap& operator=(const ConceptMap&) = delete;
+    ConceptMap& operator=(ConceptMap&&);
+
+    StringIterable variables();
+    ConceptIterable concepts(); // TODO: Are we likely to create multiple copies to the same native object if we expose this? Should we copy the whole map over?
```
krishnangovindraj, line 49: Specific question:
If I iterate through `concepts()` and get a variable through get, Are they the same rust object?
Are these ever the same object? (things like DatabaseManager? )

General question: 
If I call the same method twice, will I get a different native pointer each time?

E.g.: 
If I have an attribute and get its value ?

WORKSPACE
```diff
@@ -265,3 +265,9 @@ load("@vaticle_bazel_distribution//common:rules.bzl", "workspace_refs")
 workspace_refs(
     name = "vaticle_typedb_driver_workspace_refs"
 )
+
+new_local_repository(
+    name = "gherkin_cpp",
+    path = "/Users/krishnangovindraj/code/dependency_libraries/dist/gherkin",
+    build_file = "cpp/test/cucumber/gherkin.bzl"
+)
```
krishnangovindraj, line 273: I should already upload this somewhere so I don't lose it if my disk dies.

cpp/BUILD
```diff
@@ -0,0 +1,66 @@
...
+
+cc_library(
+    name = "typedb-driver-cpp",
+    srcs = glob([
+        "lib/**/*.cpp", 
+        "lib/inc/*.hpp"
+    ]),
+    hdrs = glob(["include/typedb/**/*"]), # exported public includes
+    strip_include_prefix = "include",
+    includes = ["lib/"], # private includes
```
krishnangovindraj, line 31: I could also use relative paths.

cpp/BUILD
```diff
@@ -0,0 +1,66 @@
...
+    ],
+    copts = ["-std=c++17"],
+    env = {"RUST_BACKTRACE": "full"},
+)
+
+cc_binary(
+    name = "test-behaviour-cucumber-server",
+    srcs = glob(["test/features/step_definitions/**/*.cpp"]),
+    deps = ["@cucumber_cpp//:cucumber-cpp"],
+)
```
krishnangovindraj, line 58: This is no longer used and has to be cleaned up.

cpp/include/typedb/answer/iterators.hpp
```diff
@@ -0,0 +1,46 @@
...
+    TypeDB::ConceptMap>;
+
+template <>
+std::function<void(_native::ConceptMapIterator*)> ConceptMapIterator::fn_nativeIterDrop;
+template <>
+std::function<_native::ConceptMap*(_native::ConceptMapIterator*)> ConceptMapIterator::fn_nativeIterNext;
+template <>
+std::function<void(_native::ConceptMap*)> ConceptMapIterator::fn_nativeElementDrop;
+
+}  // namespace TypeDB
```
krishnangovindraj, line 46: I've since merged this into `conceptmap.hpp`

cpp/include/typedb/common/exception.hpp
```diff
@@ -0,0 +1,45 @@
...
+   public:
+    TypeDBDriverException(const char* code, const char* message);
+    const std::string& code();
+    const std::string& message();
+    const char* what() const noexcept override;
+
+    static void check_and_throw();
+};
+
+typedef TypeDBDriverException TypeDBDriverCustomException;
```
krishnangovindraj, line 43: How should I handle errors which aren't due to TypeDB? Such as users accessing a moved object? TypeDBDriver but with errorCode INTERNAL?
I've currently used TypeDBDriverCustomException to keep them separate.

cpp/include/typedb/common/future.hpp
```diff
@@ -0,0 +1,75 @@
...
+using VoidFuture = TypeDBFuture<void, _native::VoidPromise>;
+using BoolFuture = TypeDBFuture<bool, _native::BoolPromise>;
+using StringFuture = TypeDBFuture<std::string, _native::StringPromise>;
+
+template <>
+std::function<void(_native::VoidPromise*)> VoidFuture::fn_nativePromiseResolve;
+template <>
+std::function<bool(_native::BoolPromise*)> BoolFuture::fn_nativePromiseResolve;
+template <>
+std::function<std::string(_native::StringPromise*)> StringFuture::fn_nativePromiseResolve;
```
krishnangovindraj, line 73: Primitive promises live here.
TypeDB class promises (e.g. ConceptPromise) live in `concept.hpp` (or that's where I intend to put them when I 'implement' them)

cpp/include/typedb/common/iterator.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+    // TODO: Make const & maybe public
+    static std::function<void(NATIVE_ITER*)> fn_nativeIterDrop;
+    static std::function<NATIVE_T*(NATIVE_ITER*)> fn_nativeIterNext;
+    static std::function<void(NATIVE_T*)> fn_nativeElementDrop; // TODO: Unused. Remove? obj handles this currently.
+
+   private:
+    
+    NativePointer<NATIVE_ITER> iteratorNative;
+    T obj; // TODO: Make std::optional
```
krishnangovindraj, line 56: Still TODO

cpp/include/typedb/common/native.hpp
```diff
@@ -0,0 +1,61 @@
...
+namespace Constants {
+namespace SessionType {
+    constexpr _native::SessionType DATA = _native::Data;
+    constexpr _native::SessionType SCHEMA = _native::Schema;
+}
+namespace TransactionType {
+    constexpr _native::TransactionType READ = _native::Read;
+    constexpr _native::TransactionType WRITE = _native::Write;
+}
+}
```
krishnangovindraj, line 59: Will change these to `enum class`. This was before I learned they were a thing.

cpp/include/typedb/concept/thing/attribute.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+class Attribute : public Thing {
+    friend class Concept;
+
+   private:
+    std::optional<Value> value;
+    Attribute(_native::Concept* conceptNative);
+   public:
+    Value* getValue(); // TODO: Review
+    // ?
```
krishnangovindraj, line 40: Should have written TODO instead. Will do a search and replace.

cpp/include/typedb/concept/type/attributetype.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "typedb/concept/type/thingtype.hpp"
+
+namespace TypeDB {
+
+class AttributeType : public ThingType {
+    friend class Concept;
+
+   private:
+    AttributeType(_native::Concept* conceptNative);
```
krishnangovindraj, line 32: Missing symbols have since been implemented. 

Is there a way to check for these missing symbols before they're called? Should I add `-z defs` (or whatever the right one is) to my linker options?

cpp/BUILD
```diff
@@ -0,0 +1,66 @@
...
+    ]),
+    hdrs = glob(["include/typedb/**/*"]), # exported public includes
+    strip_include_prefix = "include",
+    includes = ["lib/"], # private includes
+    deps = [
+        "//c:typedb_driver_clib",
+        "//c:typedb_driver_clib_headers"
+    ],
+    visibility = ["//visibility:public"],
+    copts = ["-std=c++17"]
```
krishnangovindraj, line 37: We should figure out the set of options to use.

cpp/include/typedb/concept/thing/attribute.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace TypeDB {
+
+class Concept;
+
+class Attribute : public Thing {
+    friend class Concept;
+
+   private:
+    std::optional<Value> value;
```
krishnangovindraj, line 36: Should I store it at all? Or just hit the native layer in the function?
It ties into the question of whether it's the same native pointer each time.

krishnangovindraj (commented):
questions

cpp/include/typedb/common/iterator.hpp
dmitrii-ubskii, line 1: I don't think `Iterable` serves a purpose here. It might be simpler to just put `begin()` and `end()` on the iterator, where `begin()` would return itself and `end()` would return a default-constructed iterator. That approach would get rid of a lot of code here.

cpp/include/typedb/concept/value/value.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <chrono>
+
+#include "typedb/concept/concept.hpp"
+
+namespace TypeDB {
+
+class Value;
+
+typedef std::chrono::time_point<std::chrono::system_clock, std::chrono::milliseconds> DateTime;  // TODO: Find the appropriate type
```
dmitrii-ubskii, line 33: `std::chrono::time_point` is probably the best we can get in the STL without defining our own.

cpp/include/typedb/query/querymanager.hpp
```diff
@@ -0,0 +1,72 @@
...
+    WARN_UNUSED ConceptMapIterable get(const std::string& query, const Options& options) const;
+    WARN_UNUSED JSONIterable fetch(const std::string& query, const Options &options) const;
+    WARN_UNUSED ConceptMapIterable insert(const std::string& query, const Options& options) const;
+    WARN_UNUSED VoidFuture matchDelete(const std::string& query, const Options& options) const;
+    WARN_UNUSED ConceptMapIterable update(const std::string& query, const Options& options) const;
+    WARN_UNUSED AggregateFuture getAggregate(const std::string& query, const Options& options) const;
+    WARN_UNUSED ConceptMapGroupIterable getGroup(const std::string& query, const Options& options) const;
+    WARN_UNUSED ValueGroupIterable getGroupAggregate(const std::string& query, const Options& options) const;
+    WARN_UNUSED ExplanationIterable explain(const Explainable& explainable, const Options& options) const;
+    #undef WARN_UNUSED
```
dmitrii-ubskii, line 62: Any reason we're not just using `[[nodiscard]]` here?

cpp/include/typedb/query/querymanager.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include "typedb/answer/valuegroup.hpp"
+#include "typedb/answer/valuefuture.hpp"
+
+#include "typedb/logic/explainable.hpp"
+#include "typedb/logic/explanation.hpp"
+
+namespace TypeDB {
+
+typedef std::string JSONString;
+using JSONIterable = TypeDB::StringIterable;
```
krishnangovindraj, line 39: TODO: Implement the minimal JSON

Do we want to return a custom "JSONObject" or just a string and give them a `JSONObject.from(std::string&)`  method in case they want to use ours?

cpp/BUILD
```diff
@@ -0,0 +1,66 @@
...
+    ]),
+    hdrs = glob(["include/typedb/**/*"]), # exported public includes
+    strip_include_prefix = "include",
+    includes = ["lib/"], # private includes
+    deps = [
+        "//c:typedb_driver_clib",
+        "//c:typedb_driver_clib_headers"
+    ],
+    visibility = ["//visibility:public"],
+    copts = ["-g -std=c++17"] # TODO:  Remove -g?
```
krishnangovindraj, line 37: TODO

krishnangovindraj (commented):
Flag some more discussion points.

cpp/include/typedb/query/querymanager.hpp
krishnangovindraj: Nope. I've changed it now.

cpp/include/typedb/common/iterator.hpp
krishnangovindraj: I'm wondering whether to add other methods like `count()`, `first()` and so on. Could those also go on the iterator?

We could indeed just put it on the Iterator, but the separation made it cleaner to think about when implementing.

cpp/include/typedb/__conceptfactory.hpp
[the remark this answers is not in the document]
krishnangovindraj: It's partially so I could have one friend class for all the concept classes.

cpp/include/typedb/common/future.hpp
```diff
@@ -0,0 +1,75 @@
...
```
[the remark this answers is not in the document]
krishnangovindraj:
```
 warning: instantiation of variable 'TypeDB::TypeDBFuture<void, TypeDB::_native::VoidPromise>::fn_nativePromiseResolve' required here, but no definition is available [-Wundefined-var-template]
            fn_nativePromiseResolve(promiseNative.release());
            ^
```

WORKSPACE
krishnangovindraj: Do we build from source (with the build file contents specified in the http_archive target)?

Gherkin has dependencies (and transitive dependencies, on the json library), which might make it annoying.
The easiest solution is to upload the binaries & headers (i.e. the contents of the folder here) for just the platforms we want and keep this approach.

cpp/lib/common/utils.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include "typedb/common/iterator.hpp"
+#include "typedb/common/native.hpp"
+
+#include "inc/macros.hpp"
+
+namespace TypeDB {
+namespace Utils {
+
+std::string stringAndFree(char* c) {
```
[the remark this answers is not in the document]
flyingsilverfin: Maybe the name was confusing - is the more like `nativeToStd(char* c)`?

cpp/include/typedb/concept/conceptmanager.hpp
```diff
@@ -0,0 +1,64 @@
...
+    ConceptPtrFuture<AttributeType> getAttributeType(const std::string& label) const;
+
+    ConceptPtrFuture<Entity> getEntity(const std::string& iid) const;
+    ConceptPtrFuture<Relation> getRelation(const std::string& iid) const;
+    ConceptPtrFuture<Attribute> getAttribute(const std::string& iid) const;
+
+    std::vector<TypeDBDriverException> getSchemaExceptions();
+
+   private:
+    TypeDB::Transaction* const parentTransaction;
```
flyingsilverfin, line 58: If you've adopted this naming convention, you can simplify this just to be `transaction` instead of `parentTransaction` everywhere

cpp/include/typedb/connection/driver.hpp
```diff
@@ -0,0 +1,71 @@
...
+class Driver;
+
+class Credential {
+   public:
+    Credential(const std::string& username, const std::string& password, bool withTLS, const std::string& customRootCAStr = "");
+
+   private:
+    NativePointer<_native::Credential> credentialNative;
+    _native::Credential* getNative() const;
+    friend class Driver;
```
flyingsilverfin, line 41: Conceptually, i'd not expect `Credential` to depend on having access into `Driver`, right?

cpp/include/typedb/user/usermanager.hpp
```diff
@@ -0,0 +1,59 @@
...
+ * under the License.
+ */
+#pragma once
+
+#include "typedb/common/native.hpp"
+#include "typedb/user/user.hpp"
+
+namespace TypeDB {
+
+class Driver;  // forward declaration for friendship
```
flyingsilverfin, line 28: Why do we need it here? None of the fields or constructors use a driver?

cpp/lib/concept/type/relationtype.cpp
```diff
@@ -0,0 +1,94 @@
...
+VoidFuture RelationType::unsetRelates(Transaction& transaction, const std::string& roleLabel) {
+    CONCEPTAPI_CALL(VoidFuture, _native::relation_type_unset_relates(ConceptFactory::getNative(transaction), conceptNative.get(), roleLabel.c_str()));
+}
+
+ConceptIterable<RelationType> RelationType::getSubtypes(Transaction& transaction, Transitivity transitivity) {
+    CONCEPTAPI_ITER(RelationType, _native::relation_type_get_subtypes(ConceptFactory::getNative(transaction), conceptNative.get(), (_native::Transitivity)transitivity));
+}
+
+VoidFuture RelationType::setSupertype(Transaction& transaction, RelationType* superRelationType) {
+    // CHECK_NATIVE(superRelationType);
```
flyingsilverfin, line 85: stray comment? Or required...

cpp/lib/concept/type/type.cpp
```diff
@@ -0,0 +1,174 @@
...
+    TypeDBDriverException::check_and_throw();
+
+    _native::concept_drop(rootEntity);
+    _native::concept_drop(rootAttribute);
+    _native::concept_drop(rootRelation);
+    return new ConceptIteratorWrapperChained({relationSubtypeIterator, entitySubtypeIterator, attributeSubtypeIterator});
+}
+
+ConceptIteratorWrapper* getSupertypesIteratorNativeForRootThingType(Transaction& transaction) {
+    // Actually does sync work rather than returning an async Iterator
```
flyingsilverfin, line 73: Is this expected?

cpp/test/integration/test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    {
+        auto sess = driver.session(dbName, TypeDB::SessionType::SCHEMA, options);
+        auto tx = sess.transaction(TypeDB::TransactionType::WRITE, options);
+        TypeDBDriverException::check_and_throw();
+        std::string schema =
+            " define attr sub attribute, value long, owns attr;"
+            "rule " +
+            ruleLabel + ": when { " + when + " } then { " + then + "; }; ";
+        tx.query.define(schema, options).wait();
+        TypeDBDriverException::check_and_throw();
```
flyingsilverfin, line 82: Is the user expected to manually check/throw like this? Lots of the methods in our implementation already do this inside right?

flyingsilverfin:
One more set of comments:
1. we can adopt a naming convention of `_` in file names
2. we can dissolve the `lib/inc` folder by moving the header files next to implementations for internal dependencies. Macros can go into `common`
3. **When you rebase, make sure you go back over the range of commits that you are rebasing over to make sure you include all features/bugfixes merged since you branched out**

cpp/test/integration/test.cpp
krishnangovindraj: No. I've removed those.

cpp/lib/inc/utils.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+template <class... Args>
+TypeDBDriverException exception(const ErrorMessage* errMsg, Args... args) {
+    size_t sizeRequired = snprintf(nullptr, 0, errMsg->formatString, args...);
+    char* buffer = new char[sizeRequired];
+    snprintf(buffer, sizeRequired, errMsg->formatString, args...);
+    TypeDBDriverException exception(errMsg->code, buffer);
+    delete[] buffer;
+    return exception;
+}
```
krishnangovindraj, line 39: @dmitrii-ubskii , Tried this for the TypeDBDriverException::of.
I got a "format-security" warning at some point, so I moved this here since it's an internal way of throwing anyway.

krishnangovindraj (commented):
Some replies.

krishnangovindraj (commented):
Some more comments.

cpp/include/typedb/concept/thing/thing.hpp
```diff
@@ -0,0 +1,58 @@
...
+    VoidFuture drop(Transaction& transaction);
+
+    [[nodiscard]] VoidFuture setHas(Transaction& transaction, Attribute* attribute);
+    [[nodiscard]] VoidFuture unsetHas(Transaction& transaction, Attribute* attribute);
+
+    ConceptIterable<Attribute> getHas(Transaction& transaction);
+    #pragma message("UX messages for variadic args when I use vectors instead")
+    ConceptIterable<Attribute> getHas(Transaction& transaction, const AttributeType* attribute);
+    ConceptIterable<Attribute> getHas(Transaction& transaction, const std::vector<const AttributeType*>& attributeTypes);
+    ConceptIterable<Attribute> getHas(Transaction& transaction, const std::vector<Annotation>& annotations);
```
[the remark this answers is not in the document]
krishnangovindraj: @dmitrii-ubskii 
I need design input / an opinion on my choice of using `const std::vector<ConceptType*>&` for those arguments.

Particularly in the light of how we do things with Annotations, I'm wondering whether they should be
`const std::vector<std::unique_ptr<ConceptType>>&`, to avoid ownership issues. I'll need to create a new vector for each native call, but that feels like an acceptable cost.

Please give thoughts.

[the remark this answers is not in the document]
dmitrii-ubskii: In general, when a reference to a Rust value is accepted as a reference, it is read-only, i.e. the memory will not be freed after the call resolves. Accepting `T const *` rather than `std::unique_ptr<T>` reflects that intent.
For the sake of the UX however, we can overload each place that accepts a collection of raw `T*` pointers to also accept a collection of `std::unique_ptr`s, though I'm not sure how useful that'd be.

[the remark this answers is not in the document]
krishnangovindraj: After a discord decision, we've decided to add a  `const std::vector<std::unique_ptr<ConceptType>>&` variant

cpp/lib/common/iterator.cpp
```diff
@@ -0,0 +1,37 @@
...
+namespace TypeDB {
+
+TYPEDB_ITERATOR_HELPER(
+    _native::StringIterator,
+    char,
+    std::string,
+    _native::string_iterator_drop,
+    _native::string_iterator_next,
+    _native::string_free);
+
```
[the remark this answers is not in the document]
krishnangovindraj: I like the extra after opening the namespace and before ending it.

cpp/include/typedb/common/future.hpp
```diff
@@ -0,0 +1,77 @@
...
+        } else {
+            auto t = fn_nativePromiseResolve(promiseNative.release());
+            TypeDBDriverException::check_and_throw();
+            return t;
+        }
+    }
+
+    void wait() {
+        get();
+    }
```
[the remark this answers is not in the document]
flyingsilverfin: Let's not bother atm

cpp/lib/common/utils.cpp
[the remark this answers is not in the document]
flyingsilverfin: Yeah that works

cpp/include/typedb/common/iterator.hpp
```diff
@@ -0,0 +1,171 @@
...
+    ITERATOR end() {
+        return ITERATOR();
+    }
+
+   private:
+    NativePointer<NATIVE_ITER> iteratorNative;
+};
+
+template <typename NATIVE_ITER, typename NATIVE_T, typename T>
+class TypeDBIteratorHelper {
```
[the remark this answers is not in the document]
flyingsilverfin: It's probably ok for now

cpp/include/typedb/common/native.hpp
```diff
@@ -0,0 +1,85 @@
...
+struct UserManager;
+struct ValueGroup;
+struct ValueGroupIterator;
+struct VoidPromise;
+struct StringPair;
+
+}  // namespace _native
+
+template <typename T>
+using NativePointer = std::unique_ptr<T, std::function<void(T*)> >;
```
flyingsilverfin, line 83: This is just any T right now right? Can we bound it at all?

cpp/include/typedb/connection/transaction.hpp
```diff
@@ -0,0 +1,81 @@
...
+    Transaction& operator=(const Transaction&) = delete;
+    Transaction& operator=(Transaction&&);
+
+    TypeDB::TransactionType type() const;
+
+    bool isOpen() const;
+
+    void close();
+
+    void forceClose();
```
flyingsilverfin, line 64: Why do we have this public? To the user will probably only use 'close' right?

flyingsilverfin (approved):
Generally approved, let's be sure to catch up all the APIs, address last couple comments, and merge 'dependencies'!

cpp/include/typedb/connection/session.hpp
```diff
@@ -0,0 +1,57 @@
...
+enum class SessionType {
+    DATA,
+    SCHEMA,
+};
+
+class DatabaseManager;  // forward declaration for friendship
+
+class Session {
+   public:
+    Session();
```
dmitrii-ubskii, line 37: What does it mean to default-construct a `Session`?

cpp/include/typedb/connection/transaction.hpp
dmitrii-ubskii: On the contrary, `close()` should be called automatically in `~Transaction()` and not be exposed, while `forceClose()` is a user-facing API with niche use cases. Normally transaction should be closed when it either is commited, or exits the scope.

cpp/include/typedb/answer/json.hpp
```diff
@@ -0,0 +1,138 @@
...
+const std::string ENTITY = "entity";
+const std::string RELATION = "relation";
+const std::string ATTRIBUTE = "attribute";
+}  // namespace Root
+
+}  // namespace JSONConstant
+
+enum class JSONType {
+    NONE,
+
```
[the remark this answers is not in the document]
krishnangovindraj: Remnants from when I couldn't wrap my head around how to initialise a union.

cpp/include/typedb/concept/concept.hpp
```diff
@@ -0,0 +1,146 @@
...
+    Concept(ConceptType conceptType, _native::Concept* conceptNative);
+    
+    static std::unique_ptr<Concept> ofNative(_native::Concept* conceptNative);
+
+   friend class ConceptFactory;
+};
+
+// ConceptFuture, ConceptIterator & RolePlayerIterator
+class ConceptFutureWrapper;
+class ConceptIteratorWrapper;
```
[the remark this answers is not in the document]
krishnangovindraj: That does work for Future, but less for the iterator.

It's related to your comment here: https://github.com/vaticle/typedb-driver/pull/504/files#r1419179672 .
Do we want the promise resolve (or iterator next/drop) methods to be static, or a pointer in each instance?

We could shift all of it to be instance members.

cpp/include/typedb/connection/session.hpp
krishnangovindraj: Removed all default constructors, including driver.
Do you think driver should have one?

cpp/include/typedb/concept/type/thing_type.hpp
```diff
@@ -0,0 +1,83 @@
...
+namespace TypeDB {
+
+class ThingType : public Type {
+   public:
+    virtual std::string getLabel() override;
+    [[nodiscard]] virtual VoidFuture setLabel(Transaction& transaction, const std::string& newLabel) override;
+    bool isRoot();
+    virtual bool isAbstract() override;
+    virtual BoolFuture isDeleted(Transaction& transaction) override;
+    [[nodiscard]] virtual VoidFuture drop(Transaction& transaction) override;
```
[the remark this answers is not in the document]
krishnangovindraj: Sorry, I should have flagged this earlier.
`delete` is a keyword, so I need a different function name. Currently it's `drop` everywhere.

cpp/include/typedb/connection/transaction.hpp
krishnangovindraj: Are we sure there's never going to be a case where the user wants to close the transaction in regular use? 
Should they use an `std::optional` in that case as well?

cpp/lib/answer/json.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+        case nlohmann::json::value_t::object:
+            return buildMap(from);
+        case nlohmann::json::value_t::array:
+            return buildArray(from);
+
+        default:
+            THROW_ILLEGAL_STATE;
+    }
+}
```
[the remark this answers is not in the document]
krishnangovindraj: I hit a wall when, because the recursive version of parse takes a parsed nlohmann::json object. I don't want to expose that type, so I'll still have an opaque forward declaration `ParsedJSON`, but can avoid the friendship.

I think this is a bit cleaner?

krishnangovindraj (commented):
some replies.

cpp/lib/common/exception.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+TypeDBDriverException TypeDBDriverException::of(const ErrorMessage* errMsg, ...) {
+    va_list args;
+    char buffer[2048];
+    va_start(args, errMsg);
+    vsnprintf(buffer, sizeof(buffer), errMsg->formatString, args);
+    va_end(args);
+    return TypeDBDriverException(errMsg->code, buffer);
+}
```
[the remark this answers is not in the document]
krishnangovindraj: This warning still exists. Since it's templated with Args..., does that mean a new specialisation is generated for each combination of arguments?
It's worried about a mismatch between the number of format string specifiers & the number of arguments, esp given the format string is a variable and not a literal.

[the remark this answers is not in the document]
dmitrii-ubskii: `format-security` is triggered when using a `printf`-family function with a non-literal format string, but no arguments. `va_list`, even when empty, is still a format argument, so silences that error. The issue was still there, just unreported.

You should've got `format-nonliteral` in both situations anyway. Either way, this is something we just kind of have to live with in this error message paradigm.

cpp/include/typedb/answer/conceptmap.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include "typedb/common/iterator.hpp"
+#include "typedb/common/native.hpp"
+#include "typedb/concept/all.hpp"
+#include "typedb/logic/explainable.hpp"
+
+namespace TypeDB {
+
+class ConceptMap {
+   public:
```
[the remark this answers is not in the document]
dmitrii-ubskii: I'm with Josh here, I'm used to seeing labels dedented 4 spaces. This looks misaligned.

dmitrii-ubskii (commented):
Left some comments, but it's in individual comments rather than a review.

cpp/include/typedb/concept/type/thing_type.hpp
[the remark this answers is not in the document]
krishnangovindraj: Bump

cpp/include/typedb/concept/thing/thing.hpp
```diff
@@ -0,0 +1,56 @@
...
+    BoolFuture isDeleted(Transaction& transaction);
+    VoidFuture drop(Transaction& transaction);
+
+    [[nodiscard]] VoidFuture setHas(Transaction& transaction, Attribute* attribute);
+    [[nodiscard]] VoidFuture unsetHas(Transaction& transaction, Attribute* attribute);
+
+    ConceptIterable<Attribute> getHas(Transaction& transaction);
+    ConceptIterable<Attribute> getHas(Transaction& transaction, const AttributeType* attribute);
+    ConceptIterable<Attribute> getHas(Transaction& transaction, const std::vector<const AttributeType*>& attributeTypes);
+    ConceptIterable<Attribute> getHas(Transaction& transaction, const std::vector<Annotation>& annotations);
```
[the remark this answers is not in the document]
krishnangovindraj: Still todo

cpp/lib/query/query_manager.cpp
```diff
@@ -0,0 +1,109 @@
...
+namespace TypeDB {
+
+QueryManager::QueryManager(TypeDB::Transaction* transaction)
+    : transaction(transaction) {}
+
+VoidFuture QueryManager::define(const std::string& query, const Options& options) const {
+    CHECK_NATIVE(transaction);
+    CHECK_QUERY(query);
+    WRAPPED_NATIVE_CALL(VoidFuture, _native::query_define(transaction->getNative(), query.c_str(), options.getNative()));
+}
```
krishnangovindraj, line 47: The UX here to pass options every time isn't great. Should we introduce a variant that takes the transaction options by default?

PR #515, opened 2023-11-03
UX improvements

dmitrii-ubskii:
**For the title of this PR:** please follow the grammatical rules of a usual publication title, without capitalisation (except for the first letter). Thus, the title should NOT CONTAIN CODE: no dots, no parentheses, no backticks, no brackets, etc. It needs to be distinctive (not detailed) and succinct (not lengthy). Details of this PR will go in the description. **For the description of this PR:** please replace every line in curly brackets ( { like this } ) with an appropriate description following the guidance. Finally, **please remove this paragraph**.

{ In the form of a paragraph (only use bullet points if strictly necessary), please describe the goal of this PR, why they are valuable to achieve, and reference the related GitHub issues. This section will be automatically compiled into the release notes, so please:
  - describe the impact of the change in this PR to the _user_ of this repository (e.g. end user, contributor, developer).
  - describe the new product behaviour in _present tense_, and the old behaviour and how it's been changed in _past tense_.
  - Use the _Royal We_: _"We"_ made changes, not _"I"_ made changes. }

{ Please explain what you implemented, why your changes are the best way to achieve the goal(s) above. Please describe every method, class and package, by explaining:
  - its responsibility, 
  - how it's expected to behave, and 
  - how it relates to the adjacent methods/classes/packages it interacts with. 

This would allow the reviewer to understand your intentions in the code much better. If you're adding new classes, make sure these explanations are also included in the class header comments. Last but not least, please reference the GitHub issues to be automatically closed, such like 'closes #number'. }


rust/tests/integration/queries.rs
```diff
@@ -180,6 +180,7 @@ test_for_each_arg! {
             let age = unwrap_long(age?.map.remove("age").unwrap());
             assert_eq!(age, 1);
         }
+        drop(ages);
```
dmitrii-ubskii, line 183: `ages` (a `Stream`) now borrows the transaction

rust/tests/behaviour/query/language/steps.rs
```diff
@@ -105,14 +105,16 @@ generic_step_impl! {
     #[step(expr = "get answers of typeql get")]
     pub async fn get_answers_typeql_get(context: &mut Context, step: &Step) -> TypeDBResult {
         let parsed = parse_query(step.docstring().unwrap())?;
-        context.answer = context.transaction().query().get(&parsed.to_string())?.try_collect::<Vec<_>>().await?;
+        let answer = context.transaction().query().get(&parsed.to_string())?.try_collect::<Vec<_>>().await?;
+        context.answer = answer;
```
dmitrii-ubskii, line 109: We should be able to write this on one line, but the borrow checker gets confused by the `.await`, and thinks that we attempt to borrow `context` twice at the same time.

rust/src/connection/network/transmitter/transaction.rs
```diff
@@ -259,7 +259,9 @@ impl TransactionTransmitter {
             match grpc_source.next().await {
                 Some(Ok(message)) => collector.collect(message).await,
                 Some(Err(err)) => {
-                    break collector.close(ConnectionError::TransactionIsClosedWithErrors(err.to_string())).await
+                    break collector
+                        .close(ConnectionError::TransactionIsClosedWithErrors(err.message().to_owned()))
+                        .await
```
dmitrii-ubskii, line 264: Use the correct error message rather than `Status::to_string()`

java/common/exception/TypeDBDriverException.java
```diff
@@ -28,9 +28,6 @@
  */
 public class TypeDBDriverException extends RuntimeException {
 
-    @Nullable
-    private final com.vaticle.typedb.driver.jni.Error nativeError;
-
```
dmitrii-ubskii, line 33: We no longer store the native error and instead just take its message.

c/src/error.rs
```diff
@@ -30,6 +31,14 @@ thread_local! {
     static LAST_ERROR: RefCell<Option<Error>> = RefCell::new(None);
 }
 
+#[no_mangle]
+pub extern "C" fn init_logging() {
+    const ENV_VAR: &str = "TYPEDB_DRIVER_LOG_LEVEL";
+    if let Err(err) = env_logger::try_init_from_env(Env::new().filter(ENV_VAR)) {
+        warn!("{err}");
+    }
+}
```
dmitrii-ubskii, line 40: This function allows the FFI drivers to initialize the Rust logger. (Rust driver users can do this directly)

flyingsilverfin (approved):

PR #530, opened 2023-11-21
Session callbacks: on reopen, persistent on close; FFI bug fixes

dmitrii-ubskii:
## Release notes: usage and product changes

**Session callbacks**
- All drivers:
  - implement session reopen callbacks, executed when a session closed on the server side successfully reconnects;
  - core session now attempts to reconnect if it is closed on the remote server, in line with enterprise behaviour;
  - `Session::on_close()` callbacks are now executed each time the session closes (rather than just once);
- NodeJS:
  - implement session and transaction callbacks (`onClose()`, `Session::onReopen()`);

**Miscellaneous fixes**
- Java, Python:
  - prevent exceptions in callbacks from crashing the native layer;
  - fix the issue where static root types could not be used with Concept APi;
  - reintroduce `ConceptMap.map()` to retrieve the full mapping;
- Rust:
  - convert error messages from tuple enum variants to struct, allowing the fields to be named;
- All drivers:
  -  fix the issue where session closed on remote server would not register automatically on the client side until a transaction open attempt.

## Implementation

- `Session::on_close()` callbacks are persisted when a session is closed on the server;
- fix: session pulse reports when the connnection is closed on server;
- fix: root thing types are now marked as owned in SWIG, allowing them to be used and properly cleaned up.

### DIFF
--- a/c/src/session.rs
+++ b/c/src/session.rs
@@ -66,3 +66,37 @@ pub extern "C" fn session_force_close(session: *mut Session) {
 pub extern "C" fn session_on_close(session: *const Session, callback_id: usize, callback: extern "C" fn(usize)) {
     borrow(session).on_close(move || callback(callback_id));
 }
+
+mod private {
+    use std::mem::ManuallyDrop;
+
+    pub(super) struct ForeignOnReopenCallback<Callback, Destroy: FnOnce()> {
+        pub(super) callback: Callback,
+        destroy: ManuallyDrop<Destroy>,
+    }
+
+    impl<Callback, Destroy: FnOnce()> ForeignOnReopenCallback<Callback, Destroy> {
+        pub(super) fn new(callback: Callback, destroy: Destroy) -> Self {
+            Self { callback, destroy: ManuallyDrop::new(destroy) }
+        }
+    }
+
+    impl<Callback, Destroy: FnOnce()> Drop for ForeignOnReopenCallback<Callback, Destroy> {
+        fn drop(&mut self) {
+            // SAFETY: `destroy` is inaccessible outside of `new()`, where it is initialized, and
+            // `drop()`, where it is consumed.
+            unsafe { (ManuallyDrop::take(&mut self.destroy))() }
+        }
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn session_on_reopen(
+    session: *const Session,
+    callback_id: usize,
+    callback: extern "C" fn(usize),
+    destroy: extern "C" fn(usize),
+) {
+    let callback = private::ForeignOnReopenCallback::new(move || callback(callback_id), move || destroy(callback_id));
+    borrow(session).on_reopen(move || (callback.callback)())
+}
--- a/c/src/session.rs
+++ b/c/src/session.rs
@@ -98,5 +98,8 @@ pub extern "C" fn session_on_reopen(
     destroy: extern "C" fn(usize),
 ) {
     let callback = private::ForeignOnReopenCallback::new(move || callback(callback_id), move || destroy(callback_id));
-    borrow(session).on_reopen(move || (callback.callback)())
+    borrow(session).on_reopen(move || {
+        let callback = &callback;
+        (callback.callback)()
+    })
 }
--- a/c/src/session.rs
+++ b/c/src/session.rs
@@ -19,7 +19,7 @@
  * under the License.
  */
 
-use std::ffi::c_char;
+use std::ffi::{c_char, c_void};
 
 use typedb_driver::{DatabaseManager, Options, Session, SessionType};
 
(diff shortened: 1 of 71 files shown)


java/api/TypeDBSession.java
```diff
@@ -97,6 +97,18 @@ public interface TypeDBSession extends AutoCloseable {
      */
     void onClose(Runnable function);
 
+    /**
+     * Registers a callback function which will be executed when this session is reopened.
```
flyingsilverfin, line 101: Let's add a bit more here: when the session is reopened, such as during an automatic failover.

rust/src/database/session.rs
```diff
@@ -142,6 +161,21 @@ impl Session {
         session_info.on_close_register_sink.send(Box::new(callback)).ok();
     }
 
+    /// Registers a callback function which will be executed when this session is reopened.
```
flyingsilverfin, line 164: Extend comment so users can self-drive more easily (eg. obvious next question: when is a session repoened)

c/src/session.rs
```diff
@@ -66,3 +66,40 @@ pub extern "C" fn session_force_close(session: *mut Session) {
...
+        }
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn session_on_reopen(
+    session: *const Session,
+    callback_id: usize,
+    callback: extern "C" fn(usize),
+    destroy: extern "C" fn(usize),
```
flyingsilverfin, line 98: Discussed in person:
`callback_id` -> `data` (generic payload)
`destroy` -> `finished`

flyingsilverfin:
Lastly, need to propagate same API to Node and Python as well :)

c/src/session.rs
```diff
@@ -62,7 +62,59 @@ pub extern "C" fn session_force_close(session: *mut Session) {
...
+    use std::{ffi::c_void, mem::ManuallyDrop};
+
+    struct SendPtr(*mut c_void);
+    unsafe impl Send for SendPtr {}
+
+    pub(super) struct ForeignCallback<Callback, Finished: FnOnce(*mut c_void)> {
+        data: SendPtr,
+        callback: Callback,
+        finished: ManuallyDrop<Finished>,
+    }
```
dmitrii-ubskii, line 75: `*mut c_void` is not `Send`, and neither is a closure that captures one. `SendPtr` provides a way to "smuggle" the pointer between threads. It is up to the user to ensure that it is legal to dereference the pointer in another thread.

c/typedb_driver.i
```diff
@@ -121,25 +121,51 @@ struct SessionCallbackDirector {
...
-    sessionOnCloseCallbacks.insert({ID, handler});
-    session_on_close(session, ID, &session_callback_execute);
+    std::uintptr_t ID = session_callback_register(handler);
+    session_on_close(session, reinterpret_cast<void*>(ID), &session_callback_execute, &session_callback_erase);
+}
+
+void session_on_reopen_register(const Session* session, SessionCallbackDirector* handler) {
+    std::uintptr_t ID = session_callback_register(handler);
+    session_on_reopen(session, reinterpret_cast<void*>(ID), &session_callback_execute, &session_callback_erase);
 }
```
dmitrii-ubskii, line 169: A little trick: `uintptr_t` is an integer the same size as a pointer. Callbacks expect a pointer to user data as an argument, but since our data is just an integer ID, we reinterpret the integer as a `void*` and pass that in, avoiding unnecessary allocations.

rust/src/common/options.rs
```diff
@@ -30,7 +30,7 @@ use std::time::Duration;
 /// ```rust
 /// let options = Options::new().infer(true).explain(true);
 /// ```
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Copy, Debug, Default)]
 pub struct Options {
```
dmitrii-ubskii, line 34: `Options` is 88 bytes of data which doesn't manage any memory. Marking it as trivially copyable simplifies its usage.

rust/src/connection/network/transmitter/transaction.rs
```diff
@@ -130,14 +131,13 @@ impl TransactionTransmitter {
...
-            assert!(error.is_some());
-            return box_promise(async move { Err(error.unwrap().into()) });
+            let error = self.error();
+            return box_promise(async move { Err(error.into()) });
         }
         let (res_sink, recv) = oneshot();
         let send_result = self.request_sink.send((req, Some(ResponseSink::AsyncOneShot(res_sink))));
         box_promise(async move {
-            send_result?;
+            send_result.map_err(|_| ConnectionError::TransactionIsClosed())?;
```
dmitrii-ubskii, line 140: Transforms an uninformative `SendError` into a more specific one.

rust/src/database/database.rs
```diff
@@ -179,43 +179,35 @@ impl Database {
     #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
     pub(super) async fn run_on_any_replica<F, P, R>(&self, task: F) -> Result<R>
     where
-        F: Fn(ServerDatabase, ServerConnection, bool) -> P,
+        F: Fn(ServerDatabase, ServerConnection) -> P,
         P: Future<Output = Result<R>>,
     {
-        let mut is_first_run = true;
```
dmitrii-ubskii, line 185: `Session::transaction()` was the only place that used this flag. It is no longer necessary.

java/api/TypeDBSession.java
```diff
@@ -97,6 +97,19 @@ public interface TypeDBSession extends AutoCloseable {
...
+     * A session is reopened automatically when opening a new transaction.
```
flyingsilverfin, line 102: Please update

flyingsilverfin, line 102: Add a line before when a session is closed (timeout & failover)

nodejs/stream/BidirectionalStream.ts
```diff
@@ -91,6 +93,13 @@ export class BidirectionalStream {
         this._responseCollector.close(error);
         this._responsePartCollector.close(error);
         this._dispatcher.close();
+        for (const callback of this._onClose) {
+            callback(error);
```
flyingsilverfin, line 97: I think these should be `async` - and therefore `await` here and take in async callbacks

python/typedb/api/connection/session.py
```diff
@@ -133,6 +133,23 @@ def on_close(self, function: callable) -> None:
         """
         pass
 
+    @abstractmethod
+    def on_reopen(self, function: callable) -> None:
+        """
+        Registers a callback function which will be executed when this session is reopened.
+        A session is reopened automatically when opening a new transaction.
```
flyingsilverfin, line 140: These need updating with the same text you pick for Rust

rust/src/database/session.rs
```diff
@@ -29,12 +30,27 @@ use crate::{
...
-#[derive(Debug)]
+type Callback = Box<dyn FnMut() + Send>;
+
 pub struct Session {
     database: Database,
     server_session_info: RwLock<SessionInfo>,
     session_type: SessionType,
-    is_open: AtomicCell<bool>,
+    is_open: Arc<AtomicCell<bool>>,
+
```
flyingsilverfin, line 40: Newline

flyingsilverfin (changes requested):

PR #531, opened 2023-11-29
Fetch value type

dmitrii-ubskii:


.factory/automation.yml
```diff
@@ -87,23 +87,14 @@ build:
...
+        bazel test //rust:typedb_driver_unit_tests --test_output=streamed || exit 1
+        export CORE_FAILED=
         tool/test/start-core-server.sh &&
           bazel test //rust/tests --test_output=streamed --test_arg=-- \
-            --test_arg=integration::queries::core &&
-          export TEST_SUCCESS=0 || export TEST_SUCCESS=1
+            --test_arg=integration::queries::core || export CORE_FAILED=1
         tool/test/stop-core-server.sh
-        exit $TEST_SUCCESS
+        if [[ -n "$CORE_FAILED" ]]; then exit 1; fi
```
flyingsilverfin, line 96: Can we use the same style of `CORE_FAILED` for the enterprise tests below please? Then we're establishing a pattern of `export X_FAILED=... || export X_FAILED=1` and switching or exiting on the variable?
Also we should call `TEST_SUCCESS` below `ENTERPRISE_` instead of `TEST_`

dependencies/vaticle/repositories.bzl
```diff
@@ -54,6 +54,6 @@ def vaticle_typedb_protocol():
 def vaticle_typedb_behaviour():
     git_repository(
         name = "vaticle_typedb_behaviour",
-        remote = "https://github.com/vaticle/typedb-behaviour",
-        commit = "b6a1336260a4d11b77750f06abaccba6c9a21f13",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
+        remote = "https://github.com/dmitrii-ubskii/typedb-behaviour",
+        commit = "24c4637be57a3c70ecedd671eae962cfa5737fde",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
```
flyingsilverfin, line 58: Needs update

rust/src/answer/json.rs
```diff
@@ -57,10 +57,73 @@ impl fmt::Display for JSON {
...
-            JSON::String(string) => write!(f, "{:?}", string.as_ref())?,
+            JSON::String(string) => write_escaped_string(string, f)?,
             JSON::Number(number) => write!(f, "{number}")?,
             JSON::Boolean(boolean) => write!(f, "{boolean}")?,
         }
         Ok(())
     }
 }
+
+fn write_escaped_string(string: &str, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
flyingsilverfin, line 68: Let's put a comment here where we took inspiration from so that we can go back there in the future easily if something breaks/changes? Also for creditation (check their license is complete open too please!)

rust/src/answer/readable_concept.rs
```diff
@@ -81,41 +85,52 @@ impl Node {
     }
 }
 
+const TYPE: Cow<'static, str> = Cow::Borrowed("type");
```
flyingsilverfin, line 88: Nice I like this COW approach, we will leverage it more in the future!

rust/src/answer/json.rs
dmitrii-ubskii: Hmm this code has very little in common with serde-json (Apache/MIT, not that this algorithm is their creation anyway). I would've thought testing against their implementation would be enough to give a place to start looking if smth breaks?

rust/src/answer/readable_concept.rs
dmitrii-ubskii: Note, `Cow` is a good solution for when the repeated string is static, but we should look to do proper string interning to help with type labels and other runtime strings.

.factory/automation.yml
dmitrii-ubskii: It reads a little awkward compared to `TEST_SUCCESS`, so not propagating to other jobs.

rust/src/answer/json.rs
flyingsilverfin: Ok if it's not copied out from them let's leave it as is

flyingsilverfin (approved):

PR #535, opened 2023-12-07
Java JNI library loading: fallback when platform not specified

dmitrii-ubskii:
## Usage and product changes

Previously, the JNI library would be selected based on if the containing JAR contains the expected platform string. When TypeDB Driver Java is repackaged by the end user, the JNI library is relocated and is likely missing the platform specification in its path. Now, if only one native library candidate is found in classpath, we attempt to use that rather than fail.

### DIFF
--- a/java/common/Loader.java
+++ b/java/common/Loader.java
@@ -32,7 +32,6 @@ import java.nio.file.FileSystem;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


java/common/Loader.java
```diff
@@ -75,15 +74,24 @@ private static Path getNativeResourceURI() throws IOException {
         }
         String platformString = DRIVER_JNI_JAR_NAME.get(platform);
         ClassLoader loader = Loader.class.getClassLoader();
-        URL jniURL = null;
         Iterator<URL> resourceIterator = loader.getResources(DRIVER_JNI_LIBRARY_NAME).asIterator();
+
+        URL jniURL = null;
+        boolean isValid = true;
```
flyingsilverfin, line 80: I think this will be easier to read/understand quickly if we were verbose like this:
```
URL firstJNI = null;
URL validatedJNI = null;

while (...) {
  resource = ;
  if (firstJNI == null) firstJNI = resource;

  if (...)
    validatedJNI = resource;

 }
 // switch on the final choice of URL here
 ```

flyingsilverfin (approved):

PR #540, opened 2023-12-14
Add driver fields to the connection open request

dmitrii-ubskii:
## Usage and product changes

Added and populated driver name and version fields to the Connection.Open request payload.

Note: despite proto3 being able to handle added fields, due to a library packaging issue, older versions of TypeDB cannot handle new fields. This means that this change breaks compatibility with TypeDB server 2.25.x and earlier.


.factory/automation.yml
```diff
@@ -1290,6 +1290,10 @@ release:
       command: |
         export NOTES_VALIDATE_TOKEN=$REPO_GITHUB_TOKEN
         bazel run @vaticle_dependencies//tool/release/notes:validate --test_output=streamed -- $FACTORY_OWNER $FACTORY_REPO RELEASE_NOTES_LATEST.md
+    validate-versions:
```
flyingsilverfin, line 1293: `validate-version-files`

cpp/lib/connection/driver.cpp
```diff
@@ -41,7 +42,7 @@ _native::Credential* Credential::getNative() const {
 
 
 Driver Driver::coreDriver(const std::string& coreAddress) {
-    auto p = _native::connection_open_core(coreAddress.c_str());
+    auto p = _native::connection_open_core_with_id(coreAddress.c_str(), "C++", TYPEDB_DRIVER_VERSION);
```
flyingsilverfin, line 45: We should variabilise the driver name as `C++` here so in the future they're easy to find/rename if we want

nodejs/common/rpc/RequestBuilder.ts
```diff
@@ -261,7 +263,14 @@ export namespace RequestBuilder {
...
         export function openReq() {
-            return new ConnectionOpenReq({version: Version.VERSION})
+            const VERSION_FILE_PATH = path.resolve(__dirname, "../../VERSION");
+            let DRIVER_VERSION: string;
+            try {
+                DRIVER_VERSION = fs.readFileSync(VERSION_FILE_PATH).toString().trim();
+            } catch {
+                DRIVER_VERSION = "UNKNOWN";
+            }
+            return new ConnectionOpenReq({version: Version.VERSION, driver_name: "NodeJS", driver_version: DRIVER_VERSION})
```
flyingsilverfin, line 273: variablise driver_name elsewhere so we're consistent in all the drivers

rust/src/connection/connection.rs
```diff
@@ -64,6 +64,9 @@ pub struct Connection {
     is_cloud: bool,
 }
 
+static DEFAULT_NAME: &str = "Rust";
```
flyingsilverfin, line 67: DRIVER_NAME

flyingsilverfin:
Another thing to roll into the backwards compatibility break is refactoring `ReadableConceptTree`'s empty leaf node to represent empty sub-aggregates

farost:
We preserve this PR until 3.0.0 as it breaks backward compatibility, and we don't want to break it until 3.0.0.

farost:
It is probably now implemented in 3.0, and the PR can be closed. @flyingsilverfin let's have a brief check a bit later

farost:
@flyingsilverfin We already implement the version and the driver name specification in 3.0 without this PR. The only difference is that we use a single root `VERSION` file, while we populate multiple version files here (just to validate it with a new CI job...). I do believe we can close this PR.

PR #549, opened 2023-12-20
Fix fetch sub-query aggregation null pointer

flyingsilverfin:
## Usage and product changes

We fix a bug where a Fetch query with a Get-Aggregate subquery that returned an empty (ie. undefined) answer throw a null pointer exception.

For example this used to incorrectly throw an exception if 'Alice' doesn't have any salaries in the database, since a 'sum' is undefined for 0 entries.
```
match
$x isa person, has name $n; $n == "Alice";
fetch
$n as "name";
total-salary: {
  match $x has salary $s;
  get $s; 
  sum $s;
};
```

We now correctly return the following JSON structure
```
[
  {
    "name": {"value": "Alice",  "type":  {"label": "name", "root": "attribute", "value_type": "string"}},
    "total-salary": null
  }
]
```

## Implementation

* We introduce a 'null' variant internally in each language's JSON representation.
* We update behaviour test runners and evaluators to handle nulls in the JSON answers to be checked against.

### DIFF
--- a/cpp/lib/answer/json.cpp
+++ b/cpp/lib/answer/json.cpp
@@ -32,7 +32,7 @@ help *
 namespace TypeDB {
 
 const char* jsonTypeNames[] = {
-    "JSONType::NONE",
+    "JSONType::INVALID",
 
     "JSONType::MAP",
     "JSONType::ARRAY",
@@ -41,6 +41,7 @@ const char* jsonTypeNames[] = {
     "JSONType::LONG",
     "JSONType::DOUBLE",
     "JSONType::STRING",
+    "JSONType::NULL_VALUE",
 };
 
 #define NAME(X) jsonTypeNames[(int)(X)]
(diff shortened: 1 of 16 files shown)


cpp/lib/answer/json.cpp
```diff
@@ -41,6 +41,7 @@ const char* jsonTypeNames[] = {
     "JSONType::LONG",
     "JSONType::DOUBLE",
     "JSONType::STRING",
+    "JSONType::NULL_VALUE",
```
flyingsilverfin, line 44: We can't use NULL since this is a macro name that already exists!

nodejs/concept/answer/ReadableConceptTreeImpl.ts
```diff
@@ -142,6 +144,25 @@ export namespace ReadableConceptTreeImpl {
...
+        export class Leaf {
+            private readonly readableConcept: ReadableConcept | null;
+
+            constructor(readableConcept: ReadableConcept | null) {
+                this.readableConcept = readableConcept;
+            }
+
+            asJSON(): JSON {
+                if (this.readableConcept == null) return null;
+                else return this.readableConcept.asObject();
```
flyingsilverfin, line 156: A 'Leaf' captures the nullability of the concept returned

PR #569, opened 2024-01-12
Fix cci

dmitrii-ubskii:


flyingsilverfin:
Implemented directly in 'development'

PR #571, opened 2024-01-18
Sync dependencies in CI

dmitrii-ubskii:
## Usage and product changes

We add a sync-dependencies job to be run in CI after successful snapshot and release deployments. The job sends a request to vaticle-bot to update all downstream dependencies.

The snapshot dependencies sync job is run in Factory CI after all behaviour and integration tests pass. The CircleCI tests only verify deployment, which is not extensive enough verification to trigger downstream propagation, and not strictly necessary for sync as that is only affected by git dependencies.

The release dependencies sync job, in contrast, runs in CircleCI and only runs after all deployments have succeeded and been verified, so that all downstream deployments can be safely updated.

Importantly, due to the way sync-dependencies is implemented, we revert `dependencies/vaticle/repositories.bzl` from having a dedicated typedb-protocol version line for the purposes of workspace-status, back to inlined version with the sync-marker. This means that the sync-marker is performing double duty as a workspace status marker.

Note: this PR does _not_ update the `dependencies` repo dependency. It will be updated automatically by the bot during its first pass.

### DIFF
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -1218,6 +1218,141 @@ build:
         tool/test/stop-core-server.sh
         exit $TEST_SUCCESS
 
+    deploy-crate-snapshot:
+      filter:
+        owner: vaticle
+        branch: [master, development]
+      image: vaticle-ubuntu-22.04
+      dependencies:
+        - build
+        - test-rust-unit-integration
+        - test-rust-behaviour-concept
+        - test-rust-behaviour-connection
+        - test-rust-behaviour-query-read
+        - test-rust-behaviour-query-write
+      command: |
+        export DEPLOY_CRATE_TOKEN=$REPO_VATICLE_CRATES_TOKEN
+        bazel run @vaticle_dependencies//tool/bazelinstall:remote_cache_setup.sh
+        bazel run --define version=$(git rev-parse HEAD) //rust:deploy_crate -- snapshot
+
+    deploy-npm-snapshot:
+      filter:
+        owner: vaticle
+        branch: [master, development]
+      image: vaticle-ubuntu-22.04
+      dependencies:
+        - build
+        - test-nodejs-integration
+        - test-nodejs-cloud-failover
+        - test-nodejs-behaviour-connection-core
+        - test-nodejs-behaviour-connection-cloud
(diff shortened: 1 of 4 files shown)


.factory/automation.yml
```diff
@@ -1218,6 +1218,141 @@ build:
         tool/test/stop-core-server.sh
         exit $TEST_SUCCESS
 
+    deploy-crate-snapshot:
```
flyingsilverfin, line 1221: We will descope this for now, let's push it to later until we know CloudSmith can support what we need.

.factory/automation.yml
```diff
@@ -1218,6 +1218,141 @@ build:
...
+        - test-rust-behaviour-concept
+        - test-rust-behaviour-connection
+        - test-rust-behaviour-query-read
+        - test-rust-behaviour-query-write
+      command: |
+        export DEPLOY_CRATE_TOKEN=$REPO_VATICLE_CRATES_TOKEN
+        bazel run @vaticle_dependencies//tool/bazelinstall:remote_cache_setup.sh
+        bazel run --define version=$(git rev-parse HEAD) //rust:deploy_crate -- snapshot
+
+    deploy-npm-snapshot:
```
flyingsilverfin, line 1238: Why did we bring this back in 'factory'? It's in CircleCI right?

.factory/automation.yml
```diff
@@ -1218,6 +1218,141 @@ build:
...
+        tool/test/start-core-server.sh
+        cd nodejs/test/deployment/
+        npm install https://repo.vaticle.com/repository/npm-snapshot-group/typedb-driver/-/typedb-driver-0.0.0-$FACTORY_COMMIT.tgz
+        sudo -H npm install jest --global
+        jest --detectOpenHandles application.test.js && export TEST_SUCCESS=0 || export TEST_SUCCESS=1
+        cd -
+        tool/test/stop-core-server.sh
+        exit $TEST_SUCCESS
+
+    sync-dependencies:
```
flyingsilverfin, line 1279: Let's say for commits we trigger Sync dependencies from 'Factory', but for Releases we trigger from CircleCI.
Please put this decision with a justification in the PR comment.

PR #577, opened 2024-01-23
Set default Python version to 3.11

dmitrii-ubskii:
## Usage and product changes

Set default python version for python tool binaries to 3.11 as 3.9 (the previous default) was incompatible with the sync-dependencies tool.

### DIFF
--- a/python/docs/schema/ValueType.adoc
+++ b/python/docs/schema/ValueType.adoc
@@ -19,20 +19,3 @@ a| `STRING` a| `_ValueType(is_writable=True, is_keyable=True, 4)`
 |===
 // end::enum_constants[]
 
-// tag::methods[]
-[#_ValueType_of__]
-==== of
-
-[source,python]
-----
-static of(value_type: 0 | 1 | 2 | 3 | 4 | 5) -> ValueType
-----
-
-
-
-[caption=""]
-.Returns
-`ValueType`
-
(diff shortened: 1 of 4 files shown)


python/docs/schema/ValueType.adoc
```diff
@@ -25,7 +25,7 @@ a| `STRING` a| `_ValueType(is_writable=True, is_keyable=True, 4)`
 
 [source,python]
 ----
-static of(value_type: Object | Boolean | Long | Double | String | DateTime) -> ValueType
+static of(value_type: 0 | 1 | 2 | 3 | 4 | 5) -> ValueType
```
dmitrii-ubskii, line 28: Sphinx with Py3.11 for whatever reason uses the integer representation for the constants.

flyingsilverfin: That's not great - can we look into it to try to get the names back?

dmitrii-ubskii: These really are just integers. I have not found any issues about that change, and anything more in depth is out of scope of this PR.

dmitrii-ubskii: Since this method is meant for internal use only, we've decided to remove it from the doc generation altogether.

flyingsilverfin (approved):

PR #579, opened 2024-01-25
Setup python3.11 CFFI in CI for sync-dependencies

dmitrii-ubskii:
## Usage and product changes

CI executors based on Ubuntu 22.04 default to python 3.10. Due to us having to be explicit about which version of Python we are using for native addon compilation for FFI driver, we explicitly install python3.11, pip, and the cffi package before starting a sync-dependencies job. This eliminates the potential for mismatch in the future when defaults change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -713,15 +713,16 @@ jobs:
               -c ${CIRCLE_SHA1} -delete $(cat VERSION) ~/dist/
 
   sync-dependencies:
-    executor: linux-x86_64
+    executor: linux-x86_64-ubuntu-2204
     steps:
       - checkout
       - install-bazel-apt:
           bazel-arch: amd64
-      - run:
-          command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -713,15 +713,16 @@ jobs:
               -c ${CIRCLE_SHA1} -delete $(cat VERSION) ~/dist/
 
   sync-dependencies:
-    executor: linux-x86_64
+    executor: linux-x86_64-ubuntu-2204
```
dmitrii-ubskii, line 716: Another merge resolution artifact

.circleci/config.yml
```diff
@@ -713,15 +713,16 @@ jobs:
...
     steps:
       - checkout
       - install-bazel-apt:
           bazel-arch: amd64
-      - run:
-          command: |
```
dmitrii-ubskii, line 722: Not technically an error, but not necessary either

PR #580, opened 2024-01-29
Implement a method to convert a JSON object to a JSON string

izmalk:
## Usage and product changes

The `toString()` method is callable for JSON class objects. It returns a string.

## Implementation

Add the `toString` method to the JSON class for the Fetch response in the answer section.


krishnangovindraj (changes requested):
Could you add a few integration tests?
```cpp
std::string jsonString = "{...}";
ASSERT_EQ(jsonString, JSON::parse(jsonString).toString());
``` 

I imagine we'd want newlines and indentation will not be great.

cpp/test/integration/test.cpp
```diff
@@ -160,6 +160,55 @@ TEST(TestConnection, TestMissingPort) {
...
+            result.append(json.toString());
+        }
+        ASSERT_EQ(fetchResult, result);
+    }
+
+    {
+        ASSERT_EQ(fetchResult, JSON::parse(fetchResult).toString());
+    }
+
+}
```
krishnangovindraj, line 210: This is a great test. You've fit in MAP and ARRAY, but we're still missing a bit of coverage for long, double, datetime,  and boolean. Do you think you could expand the example to include them? 
Even an unrealistic one where you're using a value variable to force it would be fine.

cpp/test/integration/BUILD
```diff
@@ -34,7 +34,8 @@ cc_test(
     copts = cxxopts,
     includes = ["include"],
     env = {"RUST_BACKTRACE": "full"},
-    linkstatic = True
+    linkstatic = True,
+    linkopts = ["-framework CoreFoundation"]
```
krishnangovindraj, line 38:
```
    linkopts = select({
        "@vaticle_bazel_distribution//platform:is_mac": ["-framework CoreFoundation"],
        "//conditions:default": [],
    }),
```

cpp/test/integration/test.cpp
izmalk: Fixed in [Add tests for Long, Double, and Boolean values](https://github.com/vaticle/typedb-driver/pull/580/commits/9c86c96bfee6bbb09b0a5056e30b58bda68e97da)

cpp/test/assembly/CMakeLists.txt
```diff
@@ -35,3 +35,5 @@ IF (WIN32)
 ELSE()
     target_link_libraries(test_assembly typedb-driver-cpp)
 ENDIF()
+
+target_link_libraries(test_assembly "-framework CoreFoundation")
```
krishnangovindraj, line 39: Missing newline

cpp/lib/answer/json.cpp
```diff
@@ -28,6 +28,7 @@ help *
 #include "../common/macros.hpp"
 #include "../common/native.hpp"
 #include "../common/utils.hpp"
+#include <sstream>
```
krishnangovindraj, line 31: I've followed a convention where I've put all inbuilt standard headers above my local headers. Can you move this one to the top as well?

krishnangovindraj (changes requested):
Last little changes.

cpp/BUILD
```diff
@@ -54,6 +54,7 @@ cc_binary(
     copts = cxxopts,
     linkopts = select({
         "@vaticle_bazel_distribution//platform:is_windows": ["ntdll.lib", "secur32.lib", "crypt32.lib", "ncrypt.lib"],
+        "@vaticle_bazel_distribution//platform:is_mac": ["-framework CoreFoundation", "-std=c++17"],
```
krishnangovindraj, line 57: `-std=c++17` isn't a link options
```suggestion
        "@vaticle_bazel_distribution//platform:is_mac": ["-framework CoreFoundation"],
```

cpp/test/integration/test.cpp
```diff
@@ -160,6 +160,84 @@ TEST(TestConnection, TestMissingPort) {
...
+    {
+        auto sess = driver.session(dbName, TypeDB::SessionType::DATA, options);
+        auto tx = sess.transaction(TypeDB::TransactionType::READ, options);
+        std::string fetchQuery = "match $u isa user, has name 'Bob'; fetch $u: name, email;";
+        TypeDB::JSONIterable response = tx.query.fetch(fetchQuery, options);
+        std::string result;
+        for (TypeDB::JSON json : response) {
+            result.append(json.toString());
+        }
+        ASSERT_EQ(fetchResult, result);
```
krishnangovindraj, line 203: It's a little confusing when they're both called results.
Maybe the naming split of actual v/s expected is better.

cpp/test/integration/test.cpp
```diff
@@ -160,6 +160,84 @@ TEST(TestConnection, TestMissingPort) {
...
+
+    {
+        auto sess = driver.session(dbName, TypeDB::SessionType::DATA, options);
+        auto tx = sess.transaction(TypeDB::TransactionType::READ, options);
+        std::string fetchQuery = "match $u isa user, has name 'Bob'; fetch $u: name, email;";
+        TypeDB::JSONIterable response = tx.query.fetch(fetchQuery, options);
+        std::string result;
+        for (TypeDB::JSON json : response) {
+            result.append(json.toString());
+        }
```
krishnangovindraj, line 202: This is a little unintuitive since we're only expecting a single answer.
We should try to make it a bit more intuitive. Maybe collect answers in a vector, assert the vector size to be one, and then compare the first elements of the vector with the expected string?

krishnangovindraj (commented):
More in depth review

cpp/test/integration/test.cpp
```diff
@@ -160,6 +160,84 @@ TEST(TestConnection, TestMissingPort) {
     }
 }
 
+TEST(TestJSON, TestJSON) {
+    std::string dbName = "test_json";
+    TypeDB::Driver driver = TypeDB::Driver::coreDriver("127.0.0.1:1729");
+    delete_if_exists(driver, dbName);
+    driver.databases.create(dbName);
+    TypeDB::Options options;
+    std::string fetchResult = R"({"u": {"email": [{"type": {"label": "email", "root": "attribute", "value_type": "string"}, "value": "bob@vaticle.com"}], "name": [{"type": {"label": "name", "root": "attribute", "value_type": "string"}, "value": "Bob"}], "type": {"label": "user", "root": "entity"}}})";
```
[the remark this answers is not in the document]
izmalk: That would be less elegant, but I can use string concatenation.

cpp/test/integration/test.cpp
izmalk: BTW, Originally I was going to use multiple results for this test, but I figured it would be hard to sort the answers in Fetch. Also, our parse method works with one JSON at a time, so I switched to a single answer by adjusting the query.

cpp/test/integration/test.cpp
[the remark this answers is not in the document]
izmalk: Fixed in https://github.com/vaticle/typedb-driver/pull/580/commits/9984d31c8e4ad3e2f3712c427a61b8387a4c4630

cpp/test/integration/test.cpp
```diff
@@ -160,6 +160,84 @@ TEST(TestConnection, TestMissingPort) {
...
+        std::string result;
+        for (TypeDB::JSON json : response) {
+            result.append(json.toString());
+        }
+        ASSERT_EQ(fetchResult, result);
+    }
+
+    {
+        ASSERT_EQ(fetchResult, JSON::parse(fetchResult).toString());
+    }
```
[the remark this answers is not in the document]
izmalk: Fixed in https://github.com/vaticle/typedb-driver/pull/580/commits/9984d31c8e4ad3e2f3712c427a61b8387a4c4630

cpp/test/integration/test.cpp
izmalk: Fixed in https://github.com/vaticle/typedb-driver/pull/580/commits/9984d31c8e4ad3e2f3712c427a61b8387a4c4630

cpp/test/integration/test.cpp
krishnangovindraj: Looks good.

krishnangovindraj (approved):

cpp/test/assembly/CMakeLists.txt
krishnangovindraj, line 39: This should be for Mac only.

cpp/include/typedb/answer/json.hpp
```diff
@@ -172,6 +172,11 @@ class JSON {
      * if this JSON object holds a string value, returns the value. Else throws an exception
      */
     const JSONString& asString() const;
+
+    /**
+     * Convert a JSON object to a JSON string.
+     */
+    const JSONString toString() const;
```
krishnangovindraj, line 179: This should return std::string.

krishnangovindraj (commented):
I missed two

cpp/lib/answer/json.cpp
```diff
@@ -321,6 +327,56 @@ JSON JSONBuilder::buildArray(const nlohmann::json& from) {
...
+        case JSONType::DOUBLE: {
+            ss << json.asDouble();
+            break;
+        }
+        case JSONType::NULL_VALUE: {
+            ss << "null";
+            break;
+        }
+        case JSONType::INVALID: {
+            throw Utils::exception(DriverError::INVALID_JSON_CAST, NAME(_type), NAME(JSONType::STRING));
```
dmitrii-ubskii, line 374: This is misleading. We're not casting to a JSON string here, but serialising.

Also, this references `this->_type` rather than `json.type()`.

cpp/lib/answer/json.cpp
```diff
@@ -321,6 +327,56 @@ JSON JSONBuilder::buildArray(const nlohmann::json& from) {
     return JSON(std::move(a));
 }
 
+void JSON::appendToString(std::stringstream& ss, const JSON& json) const {
```
dmitrii-ubskii, line 330: What's the reasoning behind not using `this` here?

cpp/lib/answer/json.cpp
```diff
@@ -256,6 +255,13 @@ const JSONString& JSON::asString() const {
     if (!isString()) throw Utils::exception(DriverError::INVALID_JSON_CAST, NAME(_type), NAME(JSONType::STRING));
     return stringValue;
 }
+
+const JSONString JSON::toString() const {
+    std::stringstream ss;
+    appendToString(ss, *this);
+    return JSONString(ss.str());
+}
```
dmitrii-ubskii, line 263: Since we are using `sstream` anyway, an implementation of `operator<<(std::ostream&, JSON&)` would be better instead. It's going to be pretty much identical to the `appendToString()`. `toString()` then could just be:
```suggestion
const std::string JSON::toString() const {
    std::stringstream ss;
    ss << *this;
    return ss.str();
}
```

dmitrii-ubskii (changes requested):

cpp/lib/answer/json.cpp
dmitrii-ubskii: `INVALID_NATIVE_HANDLE`

cpp/lib/answer/json.cpp
dmitrii-ubskii: Alternatively:
```suggestion
void appendToString(std::stringstream& ss, const JSON& json) {
```
And remove from .hpp

cpp/test/assembly/CMakeLists.txt
izmalk: Fixed in https://github.com/vaticle/typedb-driver/pull/580/commits/363cd36ed0ef83c4aeb814bec69f6828dcab5e22

izmalk:
Rebased, fixed comments (except the `this` implementation), and successfully tested (`bazel test //cpp/test/integration:test-cpp-driver-core`).

dmitrii-ubskii (approved):
LGTM

PR #593, opened 2024-02-09
Make options optional

izmalk:
## Usage and product changes

Sessions and Transactions now do not require passing an instance of options.

## Implementation

Using a new set of options by default if no options are provided when creating a session or a transaction.


krishnangovindraj:
Moved to #594 for quicker CI

PR #609, opened 2024-03-15
Introduce C# driver without deployment

farost:
## Usage and product changes

We introduce the C# driver for TypeDB. It is built using the cross-platform [.NET 6 framework](https://dotnet.microsoft.com/en-us/download/dotnet/6.0).

**Usage**: Deployment and usage examples will be provided in a separate pull request. Current state of the code lets you compiling the driver + writing and running behaviour and integration tests for it. 
The driver is expected to be a Nuget package, which could be added as a dependency to a project and referenced via "using" statements inside the users' code for all platforms.

**Architecture**: The C# driver is a thin wrapper around the TypeDB Rust driver, introducing classes for a more intuitive interface. Mostly each C# object holds a reference to the corresponding native Rust object, using an FFI ([SWIG for C#](https://www.swig.org/Doc4.2/SWIGDocumentation.html#CSharp)) for the native object wrappers generation and resource management.

Any error encountered will throw a `TypeDBDriverException`. Note that methods which return an `IEnumerable` or a Promise and encounter a server-side error will only throw when the return objects are evaluated (e.g. iterate over or call a `Linq` method for an `IEnumerable` and call `Resolve()` for a `Promise`).

A simple usage example (more to come after the integration tests are finished in another PR):
```
// Inside a try-catch block
using (ITypeDBDriver driver = TypeDB.CoreDriver(TypeDB.DEFAULT_ADDRESS))
{
    string dbName = "mydb";
    driver.Databases.Create(dbName);
    IDatabase mydb = driver.Databases.Get(dbName);
    System.Console.WriteLine(mydb.Name);

    using (ITypeDBSession schemaSession = driver.Session(dbName, SessionType.SCHEMA))
    {
        using (ITypeDBTransaction writeTransaction = schemaSession.Transaction(TransactionType.WRITE))
        {
            string defineQuery = "...some define query...";
            writeTransaction.Query.Define(defineQuery).Resolve();
            writeTransaction.Commit();
        }
    }

    mydb.Delete();
}
```

## Implementation

**Driver:**
Introduces the C# driver, wrapping the C-native classes produced by Rust driver using SWIG. The driver aims to provide a familiar interface for C# developers, hiding data transformations and native calls under the hood.
The code's architecture and build flow follow the one from Java with small C#-specific alterations.

**Tests:**
Integration tests use NUnit framework as a built-in framework for C# bazel test rules (more to come in another PR, right now these tests are messy).
Behaviour tests use Xunit.Gherkin.Quick framework and are implemented using [partial classes](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods) for flexible construction of tests based on different sets of step declaration files.

**Build:**
To fetch nuget packages, we use [Paket](https://www.nuget.org/packages/Paket) and reference this dependencies in a regular Bazel way ([example from the official Bazel repo for C# rules](https://github.com/bazelbuild/rules_dotnet/blob/master/paket.dependencies)).
[The corresponding dependencies PR for C# rules is here](https://github.com/vaticle/dependencies/pull/515).


farost:
Self review incoming in several hours.

farost:
### A quick comment about Naming

**A typical class's signature is:**
```
class ClassName
{
    private int _privateField;
    public int PublicField;

    void MethodName(int argumentName)
    {
        ...
    }
}
```

[C# identifier names coding conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names)
[C# common code conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions)

**Exception:**  we decided to mark public constants LIKE_THIS, because the rules from the Microsoft docs were not clear... Maybe we'll come up with something within this review.

**Namespaces and directories:** 
I try to follow what Microsoft says, considering Haikal's recent comment: "Please don't ever spell it Typedb":
[C# Names of Namespaces](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces)

After viewing some of the open-source C# projects, we came up with naming everything starting with an Uppercase char and declaring one namespace per library (Api, Common, Concept, Connection, ...). The last rule is broken in Behaviour/Utils, but I'll additionally comment it inside the code.

dependencies/maven/artifacts.bzl
farost, line 1: I was asked to remove it!

dependencies/maven/artifacts.snapshot
farost, line 1: I was asked to remove it!

java/api/TypeDBSession.java
```diff
@@ -55,7 +51,7 @@ public interface TypeDBSession extends AutoCloseable {
      * </pre>
      */
     @CheckReturnValue
-    String database_name();
+    String databaseName();
```
farost, line 54: This could potentially make somebody sad, so you can stop me.

java/api/TypeDBSession.java
```diff
@@ -144,15 +140,6 @@ enum Type {
             this.isSchema = nativeObject == com.vaticle.typedb.driver.jni.SessionType.Schema;
         }
 
-        public static Type of(com.vaticle.typedb.driver.jni.SessionType sessionType) {
```
farost, line 147: We decided that it's not needed for our code and users actually don't need (or even have not to) to use native values to get Java's objects, so we removed it.

java/test/behaviour/query/QuerySteps.java
```diff
@@ -84,9 +84,6 @@ public static List<ConceptMap> answers() {
 
     @Given("the integrity is validated")
     public void integrity_is_validated() {
-
-        // TODO
```
farost, line 88: "We never gonna make it, just remove the todo"

java/test/behaviour/query/QuerySteps.java
```diff
@@ -122,23 +119,23 @@ public void typeql_undefine_throws_exception(String exception, String undefineQu
     }
 
     @Given("typeql insert")
-    public Stream<ConceptMap> typeql_insert(String insertQueryStatements) {
+    public List<ConceptMap> typeql_insert(String insertQueryStatements) {
         TypeQLInsert typeQLQuery = TypeQL.parseQuery(String.join("\n", insertQueryStatements));
-        return tx().query().insert(String.join("\n", insertQueryStatements));
+        return tx().query().insert(String.join("\n", insertQueryStatements)).collect(Collectors.toList());
```
farost, line 124: It's more correct to try collecting it even if we don't check the value in the end, because we could skip some of the fails. However, it's more like a task for BDD to call for "check", so I'm not really sure if we need to collect on the "Given" step... Ready to roll back and make the same move for C# if you say so.

java/test/behaviour/query/language/rulevalidation/BUILD
farost, line 1: This test was missing in Java!!!

java/test/behaviour/query/language/rulevalidation/RuleValidationTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+@RunWith(Cucumber.class)
+@CucumberOptions(
+        strict = true,
+        plugin = "pretty",
+        glue = "com.vaticle.typedb.driver.test.behaviour",
+        features = "external/vaticle_typedb_behaviour/query/language/rule-validation.feature",
+        tags = "not @ignore and not @ignore-driver and not @ignore-typedb-driver-java"
+)
+public class RuleValidationTest extends BehaviourTest { // TODO: How to add to automation.yml?
```
farost, line 37: The TODO says it all. I didn't really understand how to correctly include it to automation.yml, because there are different sets of query/language tests over there... Would be glad to get a direct guidance.

rust/tests/behaviour/query/reasoner/steps.rs
```diff
@@ -103,6 +103,6 @@ generic_step_impl! {
 
     #[step(expr = "verify answers are consistent across {int} executions")]
     async fn verify_answers_are_consistent_across_executions(_context: &mut Context, _executions: usize) {
-        //     We can't execute previous query again because don't remember the query
+        //     We can't execute previous query again because don't remember the query // TODO: start saving the query and implement
```
farost, line 106: Maybe we want to create a task AND reference this task from the code so these TODOs could be found more easily?

farost (commented):
Tried to clean everything up, it's ready for your review.

csharp/Connection/TypeDBSession.cs
```diff
@@ -0,0 +1,157 @@
...
+namespace TypeDB.Driver.Connection
+{
+    public class TypeDBSession : NativeObjectWrapper<Pinvoke.Session>, ITypeDBSession
+    {
+        private readonly SessionType _type;
+        private readonly TypeDBOptions _options;
+
+        private readonly List<SessionCallback> _callbacks;
+
+        public TypeDBSession(IDatabaseManager databaseManager, string database, SessionType type, TypeDBOptions options)
```
farost, line 38: I just understood that it is too public, needs to be available only for our executable so (internal).  Gonna review all the constructors on Monday.

csharp/Api/BUILD
```diff
@@ -0,0 +1,50 @@
...
+package(default_visibility = ["//visibility:public"])
+
+
+load("//csharp:build_opts.bzl", "nullable_context", "target_frameworks", "targeting_packs")
+load("@rules_dotnet//dotnet:defs.bzl", "csharp_library")
+load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+
+
+csharp_library(
+    name = "api",
```
dmitrii-ubskii, line 31: Should the target names follow the package names and also be named in `UpperCamelCase`?

csharp/Api/BUILD
```diff
@@ -0,0 +1,50 @@
...
+    srcs = glob(["*.cs", "*/*.cs", "*/*/*.cs"]),
```
dmitrii-ubskii, line 32: This can be `**/*.cs`, which is recursive.

dmitrii-ubskii (changes requested):
Partial review

csharp/Api/Answer/IConceptMap.cs
```diff
@@ -0,0 +1,186 @@
...
+    {
+        /**
+         * Returns a collection of all variables in this <code>IConceptMap</code>.
+         *
+         * <h3>Examples</h3>
+         * <pre>
+         * conceptMap.Variables;
+         * </pre>
+         */
+        IEnumerable<string> Variables { get; }
```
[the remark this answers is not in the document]
farost: @dmitrii-ubskii I was thinking about it, but I wasn't really sure. 

Is `Pinvoke.typedb_driver.concept_map_get_variables(NativeObject)``` a network call as well? If so, it looks like we barely have any property in our project (not telling if it's a bad thing).

csharp/Api/BUILD
farost: It is actually a good point! 

However, it looks like (if we want to follow it...) it's even worse: following the [guideline](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-assemblies-and-dlls), our targets (which are dlls) should be like `Typedb.Driver.***` as well

csharp/Api/BUILD
farost: Thanks! I copied what Java does, and this style is more explicit for the architecture's understanding, I guess. I'd leave it like this, but I'm ready to refactor it if you want me to.

csharp/Api/BUILD
farost: Could be especially confusing in terms of tests: `Typedb.Driver.Behaviour.Connection.Database.Steps`, `Typedb.Driver.Behaviour.Connection.Database.TestCore`... 

Not sure if we want, although it would be better to make user-facing libs following the convensions.

csharp/Api/Concept/IConcept.cs
```diff
@@ -0,0 +1,354 @@
...
+         *
+         * <h3>Examples</h3>
+         * <pre>
+         * concept.AsType();
+         * </pre>
+         */
+        IType AsType()
+        {
+            throw new TypeDBDriverException(
+                ConceptError.INVALID_CONCEPT_CASTING, this.GetType().Name, typeof(IType).Name);
```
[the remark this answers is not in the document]
farost: My comment supposed to be an addition to my [other comment](https://github.com/vaticle/typedb-driver/pull/609/files#r1526406135), but it's really confusing without this context.

It means for us that I call for a C# standard library's method here, not taking a Concept's type. 

Here is an example of what's the outcome of this call:
` [CSCO01] Concept Error: Invalid concept conversion from Entity to IAttribute.`

csharp/Api/Concept/Value/IValue.cs
```diff
@@ -0,0 +1,278 @@
...
+         * <h3>Examples</h3>
+         * <pre>
+         * value.AsDateTime();
+         * </pre>
+         */
+        System.DateTime AsDateTime();
+
+        public class ValueType : NativeObjectWrapper<Pinvoke.ValueType>
+        {
+            public static readonly ValueType OBJECT = new ValueType(typeof(object), false, false, Pinvoke.ValueType.Object);
```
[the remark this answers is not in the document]
farost: @dmitrii-ubskii I don't really follow your idea, what would you like to achieve by doing it?

csharp/Api/Concept/Value/IValue.cs
```diff
@@ -0,0 +1,278 @@
...
+
+            public override string ToString()
+            {
+                if (this == OBJECT) return "object";
+                if (this == BOOL) return "bool";
+                if (this == LONG) return "long";
+                if (this == DOUBLE) return "double";
+                if (this == STRING) return "string";
+                if (this == DATETIME) return "DateTime";
+                return "Unknown";
```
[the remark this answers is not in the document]
farost: Done!

csharp/Api/TypeDBCredential.cs
```diff
@@ -0,0 +1,82 @@
...
+        {}
+
+        private TypeDBCredential(string username, string password, string? tlsRootCAPath, bool tlsEnabled)
+            : base(NativeCredential(username, password, tlsRootCAPath, tlsEnabled))
+        {}
+
+        private static Pinvoke.Credential NativeCredential(
+            string username, string password, string? tlsRootCAPath, bool tlsEnabled)
+        {
+            Debug.Assert(tlsEnabled || tlsRootCAPath == null);
```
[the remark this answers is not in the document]
farost: I'm not confused by the logic itself. It's just strange that we just have one assert in a single place for the whole proejct.

csharp/Api/Database/IDatabaseManager.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /**
+         * Retrieves all databases present on the TypeDB server.
+         *
+         * <h3>Examples</h3>
+         * <pre>
+         * driver.Databases.All;
+         * </pre>
+         */
+        IList<IDatabase> All { get; }
```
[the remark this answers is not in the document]
farost: Sure, it's just a bit confusing that `Databases` (or `getDatabases`) return a `DatabaseManager`, and I was thinking of how further methods align with both `DatabaseManager` and `Databases` names.

But it's not a huge deal, especially if we decide that we want it to be a `GetAll()` in the end of [this discussion](https://github.com/vaticle/typedb-driver/pull/609/files#r1528444354).

java/test/behaviour/query/language/rulevalidation/RuleValidationTest.java
farost, line 1: My previous comment got outdated for some reason, but I'll repeat it: this test was missing in Java!

csharp/Api/Concept/Value/IValue.cs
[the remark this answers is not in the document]
dmitrii-ubskii: Can we make `ValueType` a C# enum, for which the values are equal to the `Pinvoke.ValueType` values, with methods that return the appropriate native type, and whether attributes of this type are writable or keyable?

csharp/Api/Answer/IConceptMap.cs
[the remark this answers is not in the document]
dmitrii-ubskii: `concept_map_get_variables` is not a network call, it's a getter. Anything that returns a promise is a network call, and iterators _usually_ are network streams, but not always (like, indeed, `concept_map_get_variables`).
The rule of thumb could be: if it makes sense to cache the response on the C# side, it can be a property.

c/swig/typedb_driver_csharp.swg
```diff
@@ -0,0 +1,660 @@
...
+
+%typemap(csdisposing, methodname="Dispose", methodmodifiers="protected", parameters="bool disposing") Prefix ## Promise %{{
+    lock(this)
+    {
+        if (swigCPtr.Handle != global::System.IntPtr.Zero)
+        {
+            if (swigCMemOwn)
+            {
+                swigCMemOwn = false;
+                Resolver(); // An equivalent for the standard "delete_xxxPromise".
```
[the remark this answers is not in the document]
dmitrii-ubskii: Discussed in-person: keep `Resolve`, ditch `Resolver`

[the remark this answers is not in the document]
farost: Done!

csharp/Api/Concept/Value/IValue.cs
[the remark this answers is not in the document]
farost: > Can we make `ValueType` a C# enum, for which the values are equal to the `Pinvoke.ValueType` values

Only if we hardcode the numeric values for compile time. But if somebody decides to change the order of `Pinvoke.ValueType`s, we are dead.

csharp/Api/Answer/IConceptMap.cs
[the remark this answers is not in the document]
farost: Well, if we decide to rename it (with @alexjpwalker signal as well, I guess), I'd like to collect a list of all such cases and approve this list before making changes with you. Would be a good training as well.

csharp/Test/BUILD
dmitrii-ubskii, line 1: Can't help but notice there aren't any changes in `.factory/automation.yml`, that's something that should be sorted out soonish.

rust/tests/behaviour/query/reasoner/steps.rs
dmitrii-ubskii: #612

java/test/behaviour/query/language/rulevalidation/RuleValidationTest.java
dmitrii-ubskii: Rust includes `rule_validation` with other `query-write` tests, such as `define` and `insert`.

java/test/behaviour/query/QuerySteps.java
dmitrii-ubskii: `Given` doesn't really mean anything in this case and may as well be `Step`. The actual step keywords are ignored.
`.collect()` _could_ be replaced with `.next()` if it weren't for the fact that this "step" is reused as a helper function by other steps. As things stand though, this seems to be a reasonable compromise.

csharp/nuget/BUILD
dmitrii-ubskii, line 1: Should this package be moved under "dependencies/nuget", next to "dependencies/maven"?

dmitrii-ubskii (changes requested):

farost:
Going to add the build steps for C# to automation.yml and fix the build tomorrow.

java/test/behaviour/query/language/rulevalidation/RuleValidationTest.java
farost: Yep, I'm not really sure why I was confused back then, but it's not a problem anymore. Thanks!

rust/tests/behaviour/query/reasoner/steps.rs
farost: Cool, thank you! What do you think about starting referencing github issues inside the code TODOs? For example:
` // TODO #612: start saving the query and implement`

It would help us cleaning outdated todos and remembering reasons behind each of them on a long distance.

csharp/Api/Concept/Type/IType.cs
```diff
@@ -0,0 +1,162 @@
...
+         * <h3>Examples</h3>
+         * <pre>
+         * type.IsDeleted(transaction).Resolve();
+         * </pre>
+         *
+         * @param transaction The current transaction
+         */
+        Promise<bool> IsDeleted(ITypeDBTransaction transaction);
+    }
+}
```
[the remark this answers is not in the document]
farost: It is fixed and the comment is outdated! @dmitrii-ubskii maybe you've been looking at an old commit yesterday? Is it good now?

csharp/nuget/BUILD
farost: I dunno, maybe! I put it inside csharp as it's a csharp-only thing, but if you want to store such dependencies in a specific place, I'm not against moving it.

csharp/Test/BUILD
farost: Done!

csharp/Api/Concept/Value/IValue.cs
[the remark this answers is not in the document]
farost: You're right, I have everything on the compile time. Fixed!

csharp/Api/Answer/IConceptMap.cs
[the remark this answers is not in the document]
farost: Decided and renamed. Updated the original comment.

csharp/Common/Exception/ErrorMessage.cs
```diff
@@ -0,0 +1,147 @@
...
+ * under the License.
+ */
+
+using System;
+
+namespace TypeDB.Driver.Common
+{
+    public abstract class ErrorMessage
+    {
+        private readonly string _codePrefix;
```
farost, line 28: (duplicate old comment as it was outdated within the refactoring):

readonly vs { get; }:

*If I expect to always get the same result for each call, it's a readonly field.
*If I expect that the result can be different, it is a property with a getter (with no setters).
*If we call a native function under the hood, it's a property with a getter (with no setters).

csharp/nuget/update.sh
```diff
@@ -0,0 +1,29 @@
...
+# under the License.
+#
+
+# To update the required dependencies, follow these steps:
+# 1. Add the desired nuget packages into the 'paket.dependencies' file.
+# 2. Run this shell script !from this directory!: ./update.sh.
+# 3. Check the fetched packages in the regenerated 'paket.csharp_deps.bzl' file.
+# 4. Reference this packages in your BUILD files via '@paket.csharp_deps//{package.name.in.lowercase}'.
+
+paket install
```
flyingsilverfin, line 28: So this means that CSharp build is _not_ hermitic right? We should outline this in the PR description and note the build requirements on the local system.

java/api/concept/value/Value.java
```diff
@@ -208,16 +208,42 @@ public static Type of(com.vaticle.typedb.driver.jni.ValueType valueType) {
             throw new TypeDBDriverException(UNEXPECTED_NATIVE_VALUE);
         }
 
+        /**
```
flyingsilverfin, line 211: Thanks for all these fixes!

csharp/Test/Integration/Memory/MemoryTest.cs
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    [TestFixture]
+    public class MemoryTestFixture
+    {
+        [Test]
+        public void SendArraysToNativeSide()
+        {
+            string dbName = "mydb";
+            ITypeDBDriver driver = Utils.OpenConnection();
```
flyingsilverfin, line 103: Note: in your PR description, you should also use `driver` as the name for a variable holding a driver instance.

csharp/Test/Integration/Memory/MemoryTest.cs
```diff
@@ -0,0 +1,311 @@
...
+
+            foreach (var db in allDbs)
+            {
+                Assert.True(expectedDbNames.Contains(db.Name));
+            }
+        }
+    }
+
+    [TestFixture]
+    public class MemoryTestFixture
```
flyingsilverfin, line 97: Sorry, i'm a bit confused - these are called 'memory' tests but i don't see anything to do with assertions around memory usage/layout/existence exactly? What is the intent of the test?

flyingsilverfin (changes requested):
Very good already, almost there!
Note: your integration tests are very complete, hopefully they didn't take you too long  and were useful during development.

c/swig/typedb_driver_csharp.swg
```diff
@@ -0,0 +1,683 @@
...
+      {
+          return new global::System.Runtime.InteropServices.HandleRef(null, global::System.IntPtr.Zero);
+      }
+
+      if (!obj.swigCMemOwn)
+      {
+        throw new global::System.ApplicationException("Cannot release ownership as memory is not owned");
+      }
+
+      global::System.Runtime.InteropServices.HandleRef ptr = obj.swigCPtr;
```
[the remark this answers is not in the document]
farost: It's kinda complicated to import something inside the SWIG file as it's full of layered namespaces. SWIG files tend to be verbose because of it, it is true. I can add `using ...`s locally for methods, but I don't really see a need in it for SWIG files...

c/swig/typedb_driver_csharp.swg
```diff
@@ -0,0 +1,683 @@
...
+ * under the License.
+ */
+
+%csconst(1);
+
+%typemap(cscode) SWIGTYPE %{
+    public $csclassname Released()
+    {
+        var cPtr = swigCPtr.Handle;
+        if (swigCMemOwn)
```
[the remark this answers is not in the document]
farost: Didn't think about it! I can wrap the body of this method in `lock(this) {}` just as the SWIG-generated `Dispose` does it just not to worry about it.

c/swig/typedb_driver_csharp.swg
```diff
@@ -0,0 +1,683 @@
...
+
+    public void Resolve()
+    {
+        swigCMemOwn = false;
+        typedb_driver.void_promise_resolve(this);
+    }
+%}
+
+%typemap(csdisposing, methodname="Dispose", methodmodifiers="protected", parameters="bool disposing") VoidPromise %{{
+    lock(this)
```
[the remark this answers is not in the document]
farost: `Dispose` is auto generated by SWIG and is a method of freeing unmanaged resources when leaving `using` blocks with allocated memory for [`IDisposable` interface](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-8.0). You see it here as I use `%typemap(csbody)` typemap to rewrite the body of the generated class, thus erasing this automatically generated method, so I just return it here. 

But yes, they want to prevent race conditions here (with a bit dirty way, tbh). 

On the other hand, `Released` is a method for manual release of our resources in specific situations, similar to Java Driver.

csharp/BUILD
```diff
@@ -0,0 +1,73 @@
...
+
+
+swig_native_csharp_library(
+    name = "typedb_driver_pinvoke",
+    lib = "//c:typedb_driver_clib_headers",
+    namespace = "TypeDB.Driver.Pinvoke",
+    interface = "//c:typedb_driver.i",
+    includes = ["//c:swig/typedb_driver_csharp.swg"],
+    enable_cxx = True,
+    nullable_context = nullable_context,
```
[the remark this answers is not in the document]
farost: It doesn't help us for SWIG-generated files as they lack some code inside (and I can't insert it in the needed places, only within namespaces), but I thought that ignoring this flag for this specific build would push somebody to waste time and try throwing it all the way down this build again without success. 

Don't think that we need to worry about it

csharp/Api/BUILD
farost: It looks like I can't specify a separate name for the library file using [the rule](https://github.com/bazelbuild/rules_dotnet/blob/master/docs/csharp_library.md) (`[impl](https://github.com/bazelbuild/rules_dotnet/blob/862c7f5b8bc9d13de064ef0183e30da11034f3a4/dotnet/private/rules/csharp/library.bzl#L4)`). Moreover, these files are generated without namespaces/packages in their names. so its just `driver-csharp.dll`, `api.dll`, `common.dll` for now. So I'm not sure what to follow and if we want to try to rename these files with an additional rule to have `TypeDB.Driver.Api`and... Uhm... `TypeDB.Driver`?

csharp/nuget/update.sh
farost: Well, it depends on what we consider as hermetic. You're right that this file is not hermetic (I forgot about it), but it is needed only if you want to run `paket2bazel` (read as "run `update.sh`"), which is needed only for developers when you update dependencies. The driver and its build afterwards is hermetic and I've already "tested" it while setting up my Linux env: I just needed to install Bazel and fetch the repo.

csharp/Test/Integration/Memory/MemoryTest.cs
farost: I do, I just didn't show the initialization of this `driver` variable. Didn't think that it would bother you, my bad! The examples from the next PR are going to show more.

csharp/Test/Integration/Memory/MemoryTest.cs
farost: All the memory manipulations are hidden within the logic and I don't really know what to check here regarding memory directly. I've named these tests `Memory` to check how marshalling of arrays and other complicated objects works via my FFI, and the tests' names had to highlight it: `SendArraysToNativeSide`, `SendOptionalEnumToNativeSide`. If something is wrong, it will not work / the data checks will fail / the tests will crash while running or stopping. 

I'm open to add comments / rename it.

cpp/lib/common/error_message.cpp
```diff
@@ -32,7 +32,7 @@ namespace TypeDB {
 
 namespace DriverError {
 
-#define ERR_DRIVER(ID, MSG) ERRMSG("CCL", "Driver Error", ID, MSG)
+#define ERR_DRIVER(ID, MSG) ERRMSG("CCCL", "Driver Error", ID, MSG)
```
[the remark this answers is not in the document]
farost: Yeah, I forgot about `DR` back then! Good find.

However, there are also `.ts` and `.py` drivers with codes like "DRI" (driver), "CON" (concept), etc. Do we want to change it as well?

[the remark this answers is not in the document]
farost: Well, you're right. I think that we can change all the drivers' errors to just one form of "DRI" instead of "{langpref}DR", it would be easier. If you agree with it, I'm going to do it.

Btw, by "conflicting" I meant logical conflicts, it was your argument when you initially asked me to refactor it... =)

csharp/Test/Integration/Memory/MemoryTest.cs
flyingsilverfin: Maybe we can call it `MarshalTest` then? Just to use the vocabulary that you use yourself

csharp/nuget/update.sh
flyingsilverfin: I see - build is hermitic, but updating dependencies is not. Correct?

flyingsilverfin (approved):

flyingsilverfin (approved):

csharp/Test/Integration/Memory/MemoryTest.cs
farost: Okay!

csharp/nuget/update.sh
farost: Correct.

cpp/lib/common/error_message.cpp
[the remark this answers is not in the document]
dmitrii-ubskii: @flyingsilverfin @farost  The reason we introduced language prefixes in codes is so that we don't run into collisions between the Rust and the host language's error codes, mainly for internal errors. `INT02` coming from Java's ""Illegal state has been reached!" is distinct from Rust `INT02` "Unable to send response over callback channel (receiver dropped)."
We then reasoned that there's no guarantee that Rust driver won't be emitting its own Concept errors, e.g., so the prefix got added to all Java error codes. That should have also been the case in Python. Node is the only place where collisions are irrelevant as it's not using FFI.

c/swig/typedb_driver_csharp.swg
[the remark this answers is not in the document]
farost: Well, not every place, just `Dispose`. `swigRelease` does not, that's why I am unsure.

.factory/automation.yml
```diff
@@ -551,12 +542,15 @@ build:
...
+        - test-csharp-integration-core
+        - test-csharp-behaviour-core
+        - test-csharp-behaviour-cloud
       command: |
-          sudo add-apt-repository ppa:deadsnakes/ppa
-          sudo apt update && sudo apt install -y python3.9 python3.9-distutils python3-pip
-          python3.9 -m pip install -U cffi
-          export SYNC_DEPENDENCIES_TOKEN=$REPO_GITHUB_TOKEN
-          bazel run @vaticle_dependencies//tool/sync:dependencies -- --source ${FACTORY_REPO}@${FACTORY_COMMIT}
+        sudo add-apt-repository ppa:deadsnakes/ppa
```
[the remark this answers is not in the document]
farost: You're right, it has been lost within rebases. Sorry for that

.factory/automation.yml
```diff
@@ -551,12 +542,15 @@ build:
...
+        sudo apt update && sudo apt install -y python3.9 python3.9-distutils python3-pip
+        python3.9 -m pip install -U cffi
+        export SYNC_DEPENDENCIES_TOKEN=$REPO_GITHUB_TOKEN
+        bazel run @vaticle_dependencies//tool/sync:dependencies -- --source ${FACTORY_REPO}@${FACTORY_COMMIT}
```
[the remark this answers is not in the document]
farost: Other commands have only 2 spaces as tabs even without `-`. Why do we need 4 here?

cpp/lib/common/error_message.cpp
```diff
@@ -32,7 +32,7 @@ namespace TypeDB {
 
 namespace DriverError {
 
-#define ERR_DRIVER(ID, MSG) ERRMSG("CCL", "Driver Error", ID, MSG)
```
farost, line 35: Git went crazy, so I duplicate the conversation here.

The latest comment from @dmitrii-ubskii:

> The reason we introduced language prefixes in codes is so that we don't run into collisions between the Rust and the host language's error codes, mainly for internal errors. INT02 coming from Java's ""Illegal state has been reached!" is distinct from Rust INT02 "Unable to send response over callback channel (receiver dropped)."
We then reasoned that there's no guarantee that Rust driver won't be emitting its own Concept errors, e.g., so the prefix got added to all Java error codes. That should have also been the case in Python. Node is the only place where collisions are irrelevant as it's not using FFI.

My reply:
I had a talk about it with @flyingsilverfin. We thought that we are okay with the collisions of the error codes as far as they are just ideological, not real collision problems within the code (each driver runs separately). And for each error code we have context of the driver if somebody wants to check what to do in each specific situation.

I'm okay with two ways:
* Having language-based prefix for **all** error codes we have in drivers
* Not having one. 

I don't really like exceptions like Nodejs, because the need of differentiating FFI/non-FFI drivers only complicates things here and a user receiving such errors is not interested if a driver they are using is based on FFI or not.

Are there technical problems with having same error codes with different content for different drivers?

c/swig/typedb_driver_csharp.swg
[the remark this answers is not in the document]
dmitrii-ubskii: In that case `Release()` twice without a lock doesn't introduce extra races, because those were already present for all native calls.
Illustration for why the lock wouldn't help here: https://dotnetfiddle.net/kO2zJf

cpp/lib/common/error_message.cpp
dmitrii-ubskii: * We have checks in cloud tests that compare error codes.
* UX and error reporting suffers when users search for the error code and find something entirely different.

.factory/automation.yml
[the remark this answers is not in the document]
dmitrii-ubskii: I just thought it may have been an artifact of uncommenting, leave as is then.

cpp/lib/common/error_message.cpp
farost: * We do not have checks that compare error codes, they would fail on factory otherwise. Only steps with "exception contains substring"
* If we care about UX, then we still have to change python and nodejs codes as well (they were outdated).

@flyingsilverfin has @dmitrii-ubskii been convincing enough to accept change to "CCDR" (cpp), "CSDR" (csharp), "JDR" (java), "PDR" (python), "NDR" (node?..) (or "TDR" as typescript?..)?

flyingsilverfin: I mean yeah i don't mind using a unique prefix for each! We can do that :)
C++ - CCDR
C# - CSDR
Java - JDR
Rust - RDR
Node - NDR
Python - PDR

If C has its own then we can use CDR, but i think it doesn't?

c/swig/typedb_driver_csharp.swg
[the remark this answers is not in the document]
farost: Well, yeah, you're right about the races and locks, looks like it already works this way, I guess...

cpp/lib/common/error_message.cpp
farost: Yeah, it does not. Okay, I proceed with this plan then, thanks!

c/swig/typedb_driver_csharp.swg
```diff
@@ -618,17 +618,41 @@
...
+        {
+            global::System.IntPtr* arrayPtr = (global::System.IntPtr*)unmanaged$csinput.ToPointer();
+
+            int i = 0;
+            for (; i < arraySize; i++)
+            {
+                arrayPtr[i] = global::System.Runtime.InteropServices.Marshal.StringToCoTaskMemAnsi($csinput[i]);
+            }
+
+            arrayPtr[i] = global::System.IntPtr.Zero;
```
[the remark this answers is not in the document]
farost: Done

cpp/lib/common/error_message.cpp
dmitrii-ubskii: > * We do not have checks that compare error codes, they would fail on factory otherwise. Only steps with "exception contains substring"

Sorry, I meant in the typedb-cloud repo.

dmitrii-ubskii (approved):

PR #616, opened 2024-03-25
Introduce C# driver without documentation and deployment

farost:
## Usage and product changes

We introduce the C# driver for TypeDB. It is built using the cross-platform [.NET 6 framework](https://dotnet.microsoft.com/en-us/download/dotnet/6.0).

**Usage**: Deployment and usage examples will be provided in a separate pull request. Current state of the code lets you compiling the driver + writing and running behaviour and integration tests for it. 
The driver is expected to be a Nuget package, which could be added as a dependency to a project and referenced via "using" statements inside the users' code for all platforms.

**Architecture**: The C# driver is a thin wrapper around the TypeDB Rust driver, introducing classes for a more intuitive interface. Mostly each C# object holds a reference to the corresponding native Rust object, using an FFI ([SWIG for C#](https://www.swig.org/Doc4.2/SWIGDocumentation.html#CSharp)) for the native object wrappers generation and resource management.

Any error encountered will throw a `TypeDBDriverException`. Note that methods which return an `IEnumerable` or a Promise and encounter a server-side error will only throw when the return objects are evaluated (e.g. iterate over or call a `Linq` method for an `IEnumerable` and call `Resolve()` for a `Promise`).

A simple usage example:
```
// Inside a try-catch block
using (ITypeDBDriver driver = TypeDB.CoreDriver(TypeDB.DEFAULT_ADDRESS))
{
    string dbName = "mydb";
    driver.Databases.Create(dbName);
    IDatabase mydb = driver.Databases.Get(dbName);
    System.Console.WriteLine(mydb.Name);

    using (ITypeDBSession schemaSession = driver.Session(dbName, SessionType.SCHEMA))
    {
        using (ITypeDBTransaction writeTransaction = schemaSession.Transaction(TransactionType.WRITE))
        {
            string defineQuery = "...some define query...";
            writeTransaction.Query.Define(defineQuery).Resolve();
            writeTransaction.Commit();
        }
    }

    mydb.Delete();
}
```

## Implementation

**Driver:**
Introduces the C# driver, wrapping the C-native classes produced by Rust driver using SWIG. The driver aims to provide a familiar interface for C# developers, hiding data transformations and native calls under the hood.
The code's architecture and build flow follow the one from Java with small C#-specific alterations.

**Tests:**
Integration tests use NUnit framework as a built-in framework for C# bazel test rules (more to come in another PR).
Behaviour tests use Xunit.Gherkin.Quick framework and are implemented using [partial classes](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods) for flexible construction of tests based on different sets of step declaration files.

**Build:**
To fetch nuget packages, we use [Paket](https://www.nuget.org/packages/Paket) and reference this dependencies in a regular Bazel way ([example from the official Bazel repo for C# rules](https://github.com/bazelbuild/rules_dotnet/blob/master/paket.dependencies)).


farost:
The original PR is [here](https://github.com/vaticle/typedb-driver/pull/609). The current PR tries to remerge the changes by squashing.

flyingsilverfin:
Re-merged after previous merge didn't do a squash-merge :)

PR #628, opened 2024-04-05
Add Bazel remote cache setup for CircleCI unix commands

farost:
## Usage and product changes
CircleCI has random fails because of some process races (when a setup process takes too much time to initialize) or network errors working with Bazel (`Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/root/.cache/bazel/_bazel_root/37e9b89f8be6f62402523a372e67422a/server/jvm.out')`). 

To address the last problem, we connect CircleCI builds to our Bazel remote cache for each Bazel operation on Unix-based systems. This way, we won't send too many requests to the remote Bazel servers, using already existing build results from our cache when it's possible.

## Implementation
We add a remote cache setup call before each Bazel command for Unix CircleCI builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,6 +110,7 @@ commands:
       - run: |
           export DEPLOY_PIP_USERNAME=$REPO_TYPEDB_USERNAME
           export DEPLOY_PIP_PASSWORD=$REPO_TYPEDB_PASSWORD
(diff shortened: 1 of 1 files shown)


farost:
This PR can cause some of the machines to have errors because of the too big amount of open files (it had already been tested before), so we don't go this way.

PR #638, opened 2024-04-22
Fix C# README examples and add issues refs to recent TODOs

farost:
## Usage and product changes
C# driver's examples should be able to be built in any namespace now.

Some of the recent `TODO`s left in the codebase have been cleaned or marked with a specific GitHub issue number.  

## Implementation
We suggest leaving refs to GitHub's issues for each newly merged `TODO`s not to lose track of these problems and have more context behind them. It can be done:
- `#issuenumber` if the issue is a part of the target repo;
- full URL if the issue is a part of another repo.

### DIFF
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -293,7 +293,7 @@ build:
 
         source tool/test/start-cloud-servers.sh 3 && # use source to receive export vars
           bazel test //python/tests/integration:test_connection --test_env=ROOT_CA=$ROOT_CA --test_output=streamed --jobs=1 &&
-          # TODO currently broken test
+          # TODO #635: currently broken test
           # bazel test //python/tests/integration:test_cloud_failover --test_env=ROOT_CA=$ROOT_CA --test_output=streamed --jobs=1 &&
           export CLOUD_FAILED= || export CLOUD_FAILED=1
         tool/test/stop-cloud-servers.sh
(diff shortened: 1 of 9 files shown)


cpp/test/cucumber/include/cucumber_bdd/runner.hpp
```diff
@@ -57,7 +57,7 @@ class TestRunner : public TestRunnerBase {
         steps.reserve(totalSteps);
         for (const StepCollection<CTX> stepVec : stepLists) {
             for (const cucumber_bdd::StepDefinition<CTX> stepDef : stepVec) {
-                steps.push_back(StepDefinition<CTX>{stepDef.regex, stepDef.impl});  // TODO: Avoid deep copy?
+                steps.push_back(StepDefinition<CTX>{stepDef.regex, stepDef.impl});
```
farost, line 60: Doesn't look like we will be interested in fixing it in the future. @krishnangovindraj approved this TODO's removal.

.factory/automation.yml
```diff
@@ -293,7 +293,7 @@ build:
 
         source tool/test/start-cloud-servers.sh 3 && # use source to receive export vars
           bazel test //python/tests/integration:test_connection --test_env=ROOT_CA=$ROOT_CA --test_output=streamed --jobs=1 &&
-          # TODO currently broken test
+          # TODO #635: currently broken test
```
farost, line 296: This format comes from [the short links format](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls), but we can switch to full URLs if it's not that comfortable. I'm open to discussions.

flyingsilverfin (approved):

PR #658, opened 2024-06-12
Fix aspect_bazel_lib dependency, update typedb and nodejs dependencies

farost:
## Usage and product changes
1. We update `typedb` dependencies to `2.28.3` before the release to verify that the builds are still correct before the `driver` release.
2. We update other dependencies that have previously broken the build because of the updated `ts` (`nodejs`) rules:
**dependencies** ref with `nodejs` and `csharp` rules becoming more mature based on [aspect_bazel_lib version globalization](https://github.com/vaticle/dependencies/pull/547);
**typedb-protocol** ref to [propagate](https://github.com/vaticle/typedb-protocol/pull/202) the fixed `ts` rules and fix its build based on this propagation. 

## Implementation
We use the updated rules fixed in [dependencies](https://github.com/vaticle/dependencies/pull/547) and also update the code based on the new requirements of the updates `ts` rules. More details in the comments.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -33,3 +33,7 @@ build:macos --stamp --workspace_status_command=$PWD/workspace-status.sh
 
 # TODO
 # build:windows --stamp --workspace_status_command=workspace-status.bat
+
+build --@aspect_rules_ts//ts:skipLibCheck=always
+fetch --@aspect_rules_ts//ts:skipLibCheck=always
+query --@aspect_rules_ts//ts:skipLibCheck=always
--- a/dependencies/vaticle/repositories.bzl
+++ b/dependencies/vaticle/repositories.bzl
@@ -20,8 +20,8 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
     git_repository(
         name = "vaticle_dependencies",
-        remote = "https://github.com/vaticle/dependencies",
-        commit = "729d960a92e145e03794395bbe59e02f122f1aee", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
+        remote = "https://github.com/farost/dependencies", # TODO: Change back to vaticle when the deps PR is merged!
+        commit = "a22e2662b126498acb258d0a788fcfe0e790a226", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
     )
 
 def vaticle_typeql():
--- a/dependencies/vaticle/repositories.bzl
+++ b/dependencies/vaticle/repositories.bzl
@@ -20,8 +20,8 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
     git_repository(
         name = "vaticle_dependencies",
-        remote = "https://github.com/farost/dependencies", # TODO: Change back to vaticle when the deps PR is merged!
-        commit = "a22e2662b126498acb258d0a788fcfe0e790a226", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
+        remote = "https://github.com/vaticle/dependencies",
+        commit = "b89c25993c1e4f11b367435b83ac19bb658bb712", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
     )
 
 def vaticle_typeql():
--- a/nodejs/BUILD
+++ b/nodejs/BUILD
@@ -82,6 +82,7 @@ ts_project(
         ":node_modules/uuid",
         ":node_modules/@types/uuid",
     ],
+    transpiler = "tsc",
     visibility = ["//visibility:public"],
     out_dir = "dist",
 )
@@ -133,6 +134,7 @@ ts_project(
         ":node_modules/typedoc",
         ":node_modules/@types/node",
     ],
+    transpiler = "tsc",
     visibility = ["//visibility:public"],
     out_dir = ".",
 )
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -63,12 +63,14 @@ cpp_deps()
 # Load //builder/csharp
 load("@vaticle_dependencies//builder/csharp:deps.bzl", dotnet_deps = "deps")
 dotnet_deps()
+
 load(
     "@rules_dotnet//dotnet:repositories.bzl",
     "dotnet_register_toolchains",
     "rules_dotnet_dependencies",
 )
 rules_dotnet_dependencies()
+
 dotnet_register_toolchains("dotnet", "6.0.413")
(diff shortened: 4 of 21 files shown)


.bazelrc
```diff
@@ -33,3 +33,7 @@ build:macos --stamp --workspace_status_command=$PWD/workspace-status.sh
 
 # TODO
 # build:windows --stamp --workspace_status_command=workspace-status.bat
+
+build --@aspect_rules_ts//ts:skipLibCheck=always
+fetch --@aspect_rules_ts//ts:skipLibCheck=always
+query --@aspect_rules_ts//ts:skipLibCheck=always
```
farost, line 39: We started receiving these errors with the newer version of the `ts` rule:
```
ERROR: /private/var/tmp/_bazel_spare/d7e7537e56195a31a080f84b389fd0bd/external/aspect_rules_ts/ts/BUILD.bazel:121:8: configurable attribute "skip_lib_check" in @aspect_rules_ts//ts:options doesn't match this configuration: 

######## Required Typecheck Performance Selection ########

TypeScript's type-checking exposes a flag `--skipLibCheck`:
https://www.typescriptlang.org/tsconfig#skipLibCheck

Using this flag saves substantial time during type-checking.
Rather than doing a full check of all d.ts files, TypeScript will only type-check the code you
specifically refer to in your app's source code.
We recommend this for most rules_ts users.

HOWEVER this performance improvement comes at the expense of type-system accuracy. 
For example, two packages could define two copies of the same type in an inconsistent way.
If you publish a library from your repository, your incorrect types may result in errors for your users.

You must choose exactly one of the following flags:

1. To choose the faster performance put this in /.bazelrc:

    # passes an argument `--skipLibCheck` to *every* spawn of tsc
    # Bazel 6.4 or greater: 'common' means 'any command that supports this flag'
    common --@aspect_rules_ts//ts:skipLibCheck=always

    # Between Bazel 6.0 and 6.3, you need all of this, to avoid discarding the analysis cache:
    build --@aspect_rules_ts//ts:skipLibCheck=always
    fetch --@aspect_rules_ts//ts:skipLibCheck=always
    query --@aspect_rules_ts//ts:skipLibCheck=always

    # Before Bazel 6.0, only the 'build' and 'fetch' lines work.

2. To choose more correct typechecks, put this in /.bazelrc:

    # honor the setting of `skipLibCheck` in the tsconfig.json file
    # Bazel 6.4 or greater: 'common' means 'any command that supports this flag'
    common --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig

    # Between Bazel 6.0 and 6.3, you need all of this, to avoid discarding the analysis cache:
    build --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig
    fetch --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig
    query --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig

    # Before Bazel 6.0, only the 'build' and 'fetch' lines work.

##########################################################
```

Here I use the fix number 2, and these lines were added by @alexjpwalker in `platform`.

nodejs/BUILD
```diff
@@ -82,6 +82,7 @@ ts_project(
         ":node_modules/uuid",
         ":node_modules/@types/uuid",
     ],
+    transpiler = "tsc",
```
farost, line 85:
```
Traceback (most recent call last):
        File "/private/var/tmp/_bazel_spare/afea01e34d46fe314768544eee28004b/external/aspect_rules_ts/ts/private/ts_project.bzl", line 246, column 21, in _ts_project_impl
                fail(transpiler_selection_required)
Error in fail: 
######## Required Transpiler Selection ########

You must select a transpiler for ts_project rules, which produces the .js outputs.

Please read https://docs.aspect.build/rules/aspect_rules_ts/docs/transpiler

##########################################################
```

Here I use the transpiler that @alexjpwalker used for `platform`, not sure if anything more special is needed. As I understand, it might affect the runtime speed of the `ts` application if the transpiler is incorrect for this specific app, and I don't think that we want to compare them for our rules. Maybe @alexjpwalker could help us verify it. 

> [tsc](https://docs.aspect.build/rules/aspect_rules_ts/docs/transpiler): This is the simplest configuration without additional dependencies. However, it's also the slowest.

+ I didn't want to put "standard" values in `.bazelrc`, I think it's much better to be explicit here.

dependencies/vaticle/repositories.bzl
```diff
@@ -20,8 +20,8 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 def vaticle_dependencies():
     git_repository(
         name = "vaticle_dependencies",
-        remote = "https://github.com/vaticle/dependencies",
-        commit = "729d960a92e145e03794395bbe59e02f122f1aee", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_dependencies
+        remote = "https://github.com/farost/dependencies", # TODO: Change back to vaticle when the deps PR is merged!
```
farost, line 23: I won't forget =)

dependencies/vaticle/repositories.bzl
```diff
@@ -32,11 +32,16 @@ def vaticle_typeql():
...
+#        remote = "https://github.com/vaticle/typedb-protocol",
+#        # NOTE: the sync-marker is also used for workspace status by Bazel!
+#        tag = "2.28.0",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
+#    )
     git_repository(
         name = "vaticle_typedb_protocol",
-        remote = "https://github.com/vaticle/typedb-protocol",
-        # NOTE: the sync-marker is also used for workspace status by Bazel!
-        tag = "2.28.0",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
+        remote = "https://github.com/farost/typedb-protocol", # TODO: Change back to vaticle when the deps PR is merged!
```
farost, line 43: Can we reference a commit from `vaticle/typedb-protocol` here or do we have to release it? I don't feel like [this PR](https://github.com/vaticle/typedb-protocol/pull/202) is a new feature to be released...

nodejs/BUILD
farost: Approved by @alexjpwalker in https://github.com/vaticle/typedb-protocol/pull/202

dependencies/vaticle/repositories.bzl
dmitrii-ubskii: This can be a commit, yes. Keep the `NOTE` in the same place and let this be a commit dependency for now. Next release will have to start from protocol, but that's not a change that needs a release on its own.
Release validation should catch this not being a tag dependency.

dmitrii-ubskii (approved):

dependencies/vaticle/repositories.bzl
farost: sync dependencies updated the commit by itself, so I'll leave it as is. thanks!

PR #659, opened 2024-06-13
Remove excessive diagnostics flags, turn on development mode for tests

farost:
## Usage and product changes
We clean the usage of `--diagnostics.***` flags used in the test servers because of the recent propagation of a newer version of `TypeDB`, now supporting the `--development-mode.enable` flag. As the development mode includes disabling any outer reporting, it is a perfect substitution for this logic. 

## Implementation
We remove the unnecessary flags everywhere, aiming to cover it by the activated development mode.

* This mode is activated by default for all the `TypeDB***Runner`s [relevant for the `Java` driver]
* This mode is **not** activated for `vaticle_typedb_artifact` fetched by `vaticle_typedb_artifact()` or built by `bazel build @vaticle_typedb_artifact_*platform*//file`, so we need to manually add `--development-mode.enable=true` for all the invocations of these artifacts.

### DIFF
--- a/csharp/Test/Behaviour/Connection/ConnectionStepsBase.cs
+++ b/csharp/Test/Behaviour/Connection/ConnectionStepsBase.cs
@@ -52,11 +52,6 @@ namespace TypeDB.Driver.Test.Behaviour
                 {"transaction-timeout-millis", (option, val) => option.TransactionTimeoutMillis(Int32.Parse(val))}
         };
 
-        public static readonly Dictionary<string, string> ServerOptions =
-            new Dictionary<string, string>(){
-                {"--diagnostics.reporting.enable", "false"}
-        };
-
         // TODO: implement configuration and remove skips when @ignore-typedb-driver is removed from .feature.
         protected bool _requiredConfiguration = false;
 
--- a/dependencies/vaticle/repositories.bzl
+++ b/dependencies/vaticle/repositories.bzl
@@ -36,7 +36,7 @@ def vaticle_typedb_protocol():
         name = "vaticle_typedb_protocol",
         remote = "https://github.com/vaticle/typedb-protocol",
         # NOTE: the sync-marker is also used for workspace status by Bazel!
-        commit = "aea498d852d763c20d9ecc7d11423d7506964b95",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
+        tag = "2.28.4",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
     )
 
 def vaticle_typedb_behaviour():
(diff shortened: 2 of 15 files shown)


csharp/Test/Behaviour/Connection/ConnectionStepsBase.cs
```diff
@@ -52,11 +52,6 @@ public abstract class ConnectionStepsBase : Feature, IDisposable
                 {"transaction-timeout-millis", (option, val) => option.TransactionTimeoutMillis(Int32.Parse(val))}
         };
 
-        public static readonly Dictionary<string, string> ServerOptions =
```
farost, line 55: Was not used, is not needed.

dependencies/vaticle/repositories.bzl
```diff
@@ -36,7 +36,7 @@ def vaticle_typedb_protocol():
         name = "vaticle_typedb_protocol",
         remote = "https://github.com/vaticle/typedb-protocol",
         # NOTE: the sync-marker is also used for workspace status by Bazel!
-        commit = "aea498d852d763c20d9ecc7d11423d7506964b95",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
+        tag = "2.28.4",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
```
farost, line 39: I decided to highlight this "downgrade" of the protocol version here before the 2.28.4 release of the driver. Krishnan updated dependencies too early and it has been propagated everywhere now (causing the propagation of the updated by these dependencies protocol as well). I'll put this protocol commit here after the release.

java/test/behaviour/connection/ConnectionStepsBase.java
```diff
@@ -60,11 +61,7 @@ public abstract class ConnectionStepsBase {
             pair("transaction-timeout-millis", (option, val) -> option.transactionTimeoutMillis(Integer.parseInt(val)))
     );
 
-    public static final Map<String, String> serverOptions = map(
-            pair("--diagnostics.reporting.errors", "false"),
-            pair("--diagnostics.reporting.statistics", "false"),
-            pair("--diagnostics.monitoring.enable", "false")
-    );
+    public static final Map<String, String> serverOptions = Collections.emptyMap();
```
farost, line 64: I decided to leave it for the `Java` driver, it would be comfortable to have a single place for these additions if we will still need to update it in the future.

tool/test/start-cloud-servers.sh
```diff
@@ -45,9 +45,7 @@ function server_start() {
     --server.encryption.file.internal-grpc.root-ca=`realpath tool/test/resources/encryption/int-grpc-root-ca.pem` \
     --server.encryption.file.internal-zmq.private-key=`realpath tool/test/resources/encryption/int-zmq-private-key` \
     --server.encryption.file.internal-zmq.public-key=`realpath tool/test/resources/encryption/int-zmq-public-key` \
-    --diagnostics.reporting.errors=false \
-    --diagnostics.reporting.statistics=false \
-    --diagnostics.monitoring.enable=false
+    --development-mode.enable=true
```
farost, line 48: It uses artifacts, so we need to explicitly set this flag. Tested.

dmitrii-ubskii (approved):
Looks good!

PR #670, opened 2024-07-05
Python. Fix driver.close(). Add integration tests for connection closing for both core and cloud

farost:
## Usage and product changes
We fix the issue https://github.com/vaticle/typedb-driver/issues/669, where the Python Driver didn't close the connection when calling `TypeDBDriver.close()`.

## Implementation
We needed to reverse the condition of `is_open()`.

The main issue is the absence of tests for our drivers' connections, so we:
* add new integration tests for both core and cloud;
* create an issue for behaviour tests to cover this hole for a brighter future: https://github.com/vaticle/typedb-behaviour/issues/289.

### DIFF
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -287,12 +287,13 @@ build:
 
         tool/test/start-core-server.sh &&
           bazel test //python/tests/integration:test_stream --test_output=streamed --jobs=1 &&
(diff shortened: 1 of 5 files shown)


dmitrii-ubskii (approved):
Looks good!

PR #684, opened 2024-09-24
3.0 Java driver

farost:
## Usage and product changes
We introduce the updated Java driver for the upcoming 3.0 release. To do so, we update the C APIs to align with the updated Rust driver, add the necessary changes to the SWIG layer and reimplement the Java driver itself.

It is already used in integration tests (not in the CI) and in [the updated console](https://github.com/typedb/typedb-console/pull/258). 

## Implementation
Non-existing features are deleted. Non-implemented features are commented out (cloud, user management, options).

The release pipeline should be the same. However, we may need to update the dependencies on typedb artifacts later in the following smaller prs.

No readme updates for now.


c/src/answer.rs
```diff
@@ -20,230 +20,117 @@
...
-pub extern "C" fn explainables_drop(explainables: *mut Explainables) {
-    free(explainables);
-}
-
-/// Checks whether the provided <code>Explainables</code> objects are equal
-#[no_mangle]
-pub extern "C" fn explainables_equals(lhs: *const Explainables, rhs: *const Explainables) -> bool {
-    borrow(lhs) == borrow(rhs)
+pub extern "C" fn query_answer_is_concept_trees_stream(query_answer: *const QueryAnswer) -> bool {
+    // TODO: Rename?
```
farost, line 70: We'll decide later, it will be trees stream for now.

dependencies/vaticle/repositories.bzl
```diff
@@ -39,15 +39,23 @@ def vaticle_typedb_common():
...
     )
+#    git_repository( # TODO: Update ref to the merged protocol!
+#        name = "vaticle_typedb_protocol",
+#        remote = "https://github.com/typedb/typedb-protocol",
+#        tag = "3.0.0-alpha-2",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
+#    )
 
 def vaticle_typedb_behaviour():
-    git_repository(
+    native.local_repository(
```
farost, line 53: Will return the old one before the merge for now.

java/TypeDB.java
farost, line 1: I tried to update the docs as I went through the code, but I might have missed something. Will have another go while you are reviewing.

c/tests/integration/driver_test.c
```diff
@@ -160,7 +160,7 @@ bool test_query_data() {
     bool success = false;
 
     // Set up connection & database
-    connection = connection_open_core(TYPEDB_CORE_ADDRESS);
+    connection = connection_open_core(TYPEDB_CORE_ADDRESS, DRIVER_LANG);
```
flyingsilverfin, line 163: Can we have two C entry points by any chance? Do we have a separate set of C functions that aren't just generated from Rust?

Eg. it would be nice to have `connection_open_core(address)` always include `"c"`, but then have another one `connection_open_core_with_description(address, lang)` which use from SWIG for java, python, etc.?

Then the users of the C driver don't have to say.... I'm using C! every time they connect :)

Also, these `connection` should  be `driver`, and the call should be `driver_open_core` or just `driver_core`

c/typedb_driver.i
```diff
@@ -47,60 +47,50 @@ struct Type {};
...
-%dropproxy(SchemaExceptionIterator, schema_exception_iterator)
 
-%dropproxy(Credential, credential)
-%dropproxy(Options, options)
+//%dropproxy(Credential, credential)
+//%dropproxy(Options, options)
 
-#define connection_drop connection_close
-#define session_drop session_close
+#define typedb_driver_drop connection_close
```
flyingsilverfin, line 60: eg these should be renamed away from `connection`, which is a term we don't use anymore.

farost:
I'll also leave a TODO for datetimes to send them in nanos. Right now we still send them in milliseconds. The `timestamp_in_nanos` is deprecated and they suggest to use `_in_nanos_opt`, which I'm not sure how to deal with at the moment.

c/tests/integration/driver_test.c
farost: I'll rename `connection`
I'll add a todo to change add a separate c entry point so we'll revisit it when we will be fixing the tests

farost:
The newlines after "committed" messages were inconsistent before, and I just removed all the newlines from there. Can insert them everywhere instead, but it's harder to track!

PR #686, opened 2024-09-24
Fix native object ownership checks in python driver

krishnangovindraj:
## Usage and product changes
Fixes native object ownership checks in python driver

## Implementation
Replaces usages of `self.native_object.is_own` (which invokes  a `@property` getter, throwing an exception) with `self._native_object.is_own`


farost (approved):
Lgtm

PR #687, opened 2024-09-25
Improve Decimal formatting

dmitrii-ubskii:
## Usage and product changes

When fractional part of a decimal value was `0`, the format routine would go into an infinite loop as it tried to count trailing zeros. This PR fixes that behaviour.

### DIFF
--- a/rust/src/concept/value.rs
+++ b/rust/src/concept/value.rs
@@ -17,9 +17,7 @@
  * under the License.
  */
 
-use std::collections::HashMap;
-use std::fmt;
-use std::fmt::{Debug, Display, Formatter};
+use std::{collections::HashMap, fmt};
 
 use chrono::{DateTime, NaiveDate, NaiveDateTime};
 use chrono_tz::Tz;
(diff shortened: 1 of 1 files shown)


rust/src/concept/value.rs
```diff
@@ -349,14 +356,14 @@ pub struct Struct {
...
 }
 
-impl Display for Struct {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        Display::fmt(self, f)
+impl fmt::Display for Struct {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(self, f)
     }
 }
```
dmitrii-ubskii, line 363: Infinite recursion

rust/src/concept/value.rs
```diff
@@ -266,26 +264,31 @@ impl Decimal {
...
+                tail_0s += 1;
+                fractional /= 10;
+            }
 
-        // count number of leading 0's that have to be represented
-        let digits = (fractional as f64).log10().floor() as u64;
-        let leading_0s = Self::FRACTIONAL_PART_DENOMINATOR - digits - tail_0s;
-        write!(f, "{}.{:0width$}{}", self.integer_part(), "", digits, width = leading_0s as usize)
+            // count number of leading 0's that have to be represented
+            let digits = fractional.ilog10() as u64; // `fractional` cannot be a power of 10 here, so ilog10 correctly gives the number of digits
```
dmitrii-ubskii, line 287: `ilog10` rounds up, which is equivalent to counting decimal digits for positive non-power-of-10 integers.

flyingsilverfin (approved):

PR #689, opened 2024-09-27
Enable Java driver release for 3.0

farost:
## Usage and product changes
Uncommented JNI and Java Driver jobs for the `circleci` config.
Updated release notes for the 3.0 release.


.circleci/config.yml
```diff
@@ -700,39 +700,39 @@ jobs:
       - checkout
       - install-bazel-yum:
           bazel-arch: arm64
-      - deploy-pip-release-unix
+#      - deploy-pip-release-unix
       - deploy-maven-jni-release-unix
-      - deploy-clib-release-unix
```
farost, line 705: It shouldn't be needed for the JNI work as everything is a part of the JNI library, right? Is it just for the standalone C library?

farost: (the C driver is not ready, I only touched the path to the outer drivers)

flyingsilverfin: Yes it doesn't have to be released for JNI to work, we just need the JNI to bundle the c library

flyingsilverfin (approved):

RELEASE_NOTES_LATEST.md
```diff
@@ -29,66 +29,28 @@ Documentation: https://typedb.com/docs/drivers/java/overview
     <dependency>
         <groupid>com.vaticle.typedb</groupid>
```
krishnangovindraj, line 30: This is out of scope for today, but we should decide if we want to shrink this from
`com.vaticle.typedb` -> `com.typedb`

krishnangovindraj (approved):
looks good to me

RELEASE_NOTES_LATEST.md
farost: You're right, I'll at least add a task for it!

PR #690, opened 2024-09-27
Rename Java package com.vaticle.typedb to com.typedb and remove typeql dependencies.

farost:
## Usage and product changes
We rename the Java driver's package from `com.vaticle.typedb` to `com.typedb`.
We remove dependencies on `typeql` as it's no more needed. 

## Implementation
Some of the `@typeql/java/common` files were moved to `./java/common` package for easier usage.


flyingsilverfin (approved):
Looks good!!

java/common/exception/ErrorMessage.java
```diff
@@ -17,12 +17,51 @@
...
+    private static Map<String, Map<Integer, ErrorMessage>> errors = new HashMap<>();
+    private static int maxCodeNumber = 0;
+    private static int maxCodeDigits = 0;
+
+    private final String codePrefix;
+    private final int codeNumber;
+    private final String message;
+    private String code = null;
+
+    protected ErrorMessage(String codePrefix, int codeNumber, String messagePrefix, String messageBody) {
```
farost, line 35: It's `protected` so we could extend it in `Console`. Doesn't sound like a very bad thing exposed to other users. But might make it `private` and copypaste the class into `Console`!

PR #693, opened 2024-10-04
Fix CircleCI jobs for Maven installation and python builds

farost:
## Usage and product changes
We fix two CircleCI issues:
* `brew install maven` [has been failing](https://app.circleci.com/pipelines/github/typedb/typedb-driver/911/workflows/9571978a-e2c7-40cc-afd6-372356f267f7/jobs/7072) for Rosetta configuration;
* python installation for Linux machines required a different process after we updated from `python3.8` to `python3.9` as a minimal version.

## Implementation
* for the first issue, we update `xcode` version to match the first suitable macOS 13 version from [this table](https://circleci.com/docs/using-macos/#supported-xcode-versions), which is [required by brew](https://docs.brew.sh/Installation);
* to install python3.9 on an amazonlinux machine, we build it with `make` as it's absent through `amazon-linux-extras` or `yum`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,9 +64,9 @@ commands:
         type: string
     steps:
       - run: |
-          amazon-linux-extras install python3.9 java-openjdk11 -y
+          amazon-linux-extras install python3.8 java-openjdk11 -y
           yum install -y git tar gcc gcc-c++ file lsof which procps
-          ln -s /usr/bin/python3.9 /usr/bin/python3
+          ln -s /usr/bin/python3.8 /usr/bin/python3
           curl -OL "https://github.com/bazelbuild/bazelisk/releases/download/v1.17.0/bazelisk-linux-<<parameters.bazel-arch>>"
           mv "bazelisk-linux-<<parameters.bazel-arch>>" /usr/local/bin/bazel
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -64,9 +64,19 @@ commands:
...
-          yum install -y git tar gcc gcc-c++ file lsof which procps
-          ln -s /usr/bin/python3.9 /usr/bin/python3
+          amazon-linux-extras install java-openjdk11 -y
+          yum install wget make gcc gcc-c++ openssl-devel bzip2-devel libffi-devel zlib-devel file lsof which procps tar git -y
+          
+          cd /tmp
+          wget https://www.python.org/ftp/python/3.9.6/Python-3.9.6.tgz
+          tar -xvf Python-3.9.6.tgz
+          cd Python-3.9.6
+          ./configure --enable-optimizations --enable-shared
```
farost, line 74: We install a single python version for both deployment and testing steps, so:
* `shared` is needed for the dynamic library usage on the testing stage
* `ldconfig` sets the path to the dynamic library globally for all the sessions (so it could be found for both steps)

dmitrii-ubskii (approved):

PR #696, opened 2024-10-08
Introduce 3.0 Python driver docs and tests

farost:
## Usage and product changes
We introduce the second version of the Python driver, fixing a number of minor bugs from the first version, enhancing existing APIs, and presenting the updated documentation and automated tests.

Major changes:
* Introduce [3.0 driver BDD tests](https://github.com/typedb/typedb-behaviour/pull/303);
* Fully return core Python driver to the CI; 
* Refactor and enrich integration tests to cover custom value types and prepare examples for `README`;
* Prepare semi-automated generation of usage examples for `README` based on integration tests running in CI to make sure that it's up to date and in a needed format in the future;
* Update API references and documentation.

## Implementation
Minor changes and details:
* Remove `Label` classes from Java and Python (3.0) drivers to simply represent labels as strings;
* Fix Rust's driver's database management's deadlock;
* Python BDDs are reimplemented in a similar to the TypeDB server fashion (using `MayError`, more typed parsing (had been abandoned in one of the iterations, judging the state of the code), etc.);
* Semi-automated examples generation could be fully automated, but it would require much more Bazel work to generate the intermediate code blocks file and pass it as a file to the second rule. It can be done later if the task receives enough priority. However, it can already be used as a one-liner (with two `bazel run` commands), and can already be used in CIs. With just a bit more dirt than it would ideally generate.


python/tests/integration/core/test_core_example.py
```diff
@@ -0,0 +1,197 @@
...
+from typedb.driver import *
+
+# EXAMPLE END MARKER
+
+
+class TestExample(TestCase):
+
+    def setUp(self):
+        with TypeDB.core_driver(TypeDB.DEFAULT_ADDRESS) as driver:
+            if driver.databases.contains("typedb"):
```
flyingsilverfin, line 32: "typedb" should be a constant somewhere, since we're reusing it throughout the example IMO

python/tests/integration/core/test_core_stream.py
```diff
@@ -26,6 +26,8 @@
 SCHEMA = TransactionType.SCHEMA
 
 
+# TODO: Understand if we need this test. If yes, rewrite it in BDDs for all the drivers.
```
flyingsilverfin, line 29: I'd delete it for now, since this test isn't actually doing anything??
Although, I think we should have a BDD to test the behaviour of queries:

1. read queries can be opened and iterated concurrently & interleaved freely
2. write and schema queries cancel previous read and write queries from responding their answers

So as soon as you run a write of schema query, previous iterators will eventually be interrupted (didn't manage to make it "eager", youll either be able to finish the answers in the buffer and it's OK, or you eventually get an error in those streams.

Commit, rollback, and close actually do the same things!

flyingsilverfin (changes requested):

flyingsilverfin (approved):

farost (commented):
Forgot to publish my comments

python/tests/integration/core/test_core_stream.py
farost: Added todos, removed the test

PR #697, opened 2024-10-16
Introduce 3.0 datetime-tz offsets and Rust driver documentation and tests

farost:
## Usage and product changes
We introduce the second version of the Rust driver, adding TimeZone offsets for `datetime-tz` value types, fixing minor bugs and presenting the updated documentation and automated tests.

Rust driver changes:
* Add `datetime-tz` offsets;
* Refactor test structure to separate integration and behaviour tests using Bazel, not Cargo flags;
* Introduce example integration test for Rust and update README with the formatted code sample;
* Introduce updated 3.0 bdds for `connection` and `driver`;
* Introduce flags to run Rust bdds in `core` or `cloud` modes when TypeDB Cloud 3.x is implemented;
* Fix Rust driver docs parser and update generated docs.

Java and Python drivers changes:
* Add `datetime-tz` offsets;
* Rename `Thing` to `Instance`;
* Remove `ThingType`,
* Added `getLabel`/`get_label` for all `Concept` classes. Previously, it was only available for `Type`s.

Python driver changes:
* Add bdd steps to match the updated declarations.


c/src/concept/concept.rs
```diff
@@ -56,36 +56,59 @@ impl DatetimeInNanos {
...
     datetime_in_nanos: DatetimeInNanos,
-    zone_id: *mut c_char,
-}
-
-impl DatetimeAndZoneId {
-    pub fn new(datetime_in_nanos: DatetimeInNanos, zone_id: String) -> Self {
-        Self { datetime_in_nanos, zone_id: release_string(zone_id) }
+    zone_name: *mut c_char,
+    local_minus_utc_offset: i32,
+    is_fixed_offset: bool,
```
farost, line 67: I could work with optional integers instead of this bool flag (`null` if not offset), but it's an additional memory and SWIG headache.

c/src/database_manager.rs
```diff
@@ -71,5 +71,5 @@ pub extern "C" fn databases_contains(driver: *mut TypeDBDriver, name: *const c_c
 /// Retrieve the database with the given name.
 #[no_mangle]
 pub extern "C" fn databases_get(driver: *mut TypeDBDriver, name: *const c_char) -> *const Database {
-    borrow_mut(driver).databases().get(string_view(name)).map(|db| arc_into_raw(db)).unwrap_or_else(|_| null())
+    try_unwrap_arc(borrow_mut(driver).databases().get(string_view(name)))
```
farost, line 74: Some cleaning

dependencies/vaticle/repositories.bzl
```diff
@@ -32,8 +32,9 @@ def vaticle_typedb_protocol():
     )
 
 def vaticle_typedb_behaviour():
+    # TODO: Update to typedb repo
```
farost, line 35: Will update, pls see https://github.com/typedb/typedb-behaviour/pull/304

rust/tests/behaviour/config/BUILD
```diff
@@ -0,0 +1,61 @@
...
+# under the License.
+
+package(default_visibility = ["//visibility:public"])
+
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
+load("@rules_rust//rust:defs.bzl", "rust_library")
+load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+load("//rust/tests/behaviour:defs.bzl", "crate_features_common")
+
+string_flag(
```
farost, line 25: We can run `bazel test //rust/tests/behaviour/... --//rust/tests/behaviour/config:mode=cloud` to run cloud tests in Rust.

rust/tests/behaviour/config/BUILD
```diff
@@ -0,0 +1,61 @@
...
+    name = "cloud",
+    flag_values = {
+        ":mode": "cloud",
+    },
+)
+
+rust_library(
+    name = "config",
+    srcs = [":lib.rs"],
+    crate_features = select({
```
farost, line 47: That's how this flag is used. By setting the `cloud` feature, we modify the compiled `config/lib.rs`.

rust/tests/behaviour/config/lib.rs
```diff
@@ -17,7 +17,14 @@
  * under the License.
  */
 
-pub(crate) mod database;
-pub(crate) mod steps;
-pub(crate) mod transaction;
-mod user;
+pub fn is_cloud() -> bool {
```
farost, line 20: The feature in action

c/src/memory.rs
```diff
@@ -47,21 +47,21 @@ pub(super) fn release_string(str: String) -> *mut c_char {
...
+    trace!("Wrap raw into arc <{}> @ {:?}", std::any::type_name::<T>(), raw);
     unsafe { Arc::from_raw(raw) }
 }
 
 pub(super) fn release_arc<T>(raw: *const T) {
     trace!("Releasing ownership of arced <{}> @ {:?}", std::any::type_name::<T>(), raw);
     assert!(!raw.is_null());
-    drop(arc_from_raw(raw))
+    drop(wrap_into_arc(raw))
 }
```
dmitrii-ubskii, line 65: Naming here is messed up. We use `release_X` to mean "release ownership of `X` to SWIG"

It should be:
* `release_arc(Arc<T>) -> *const T` for giving up ownership (without decrementing),
* `take_arc(*const T) -> Arc<T>` to take ownership of the pointer back,
* and `decrement_arc(*const T) /* -> () */` (rather than `free_arc()` since it only decrements the counters and may not free) which is basically `drop(take_arc(t))`

c/src/error.rs
```diff
@@ -68,6 +73,14 @@ pub(super) fn try_release_optional_string(result: Option<Result<String>>) -> *mu
     ok_record_flatten(result).map(release_string).unwrap_or_else(null_mut)
 }
 
+pub(super) fn try_unwrap_arc<T>(result: Result<Arc<T>>) -> *const T {
+    try_unwrap_arc_optional(ok_record(result))
+}
+
+pub(super) fn try_unwrap_arc_optional<T>(result: Option<Arc<T>>) -> *const T {
+    result.map(unwrap_arc).unwrap_or_else(null)
+}
```
dmitrii-ubskii, line 82: These should be `try_release_arc` and `try_release_optional_arc` (see my earlier comment)

python/BUILD
```diff
@@ -64,7 +64,7 @@ sphinx_to_adoc(
 
 test_to_example(
     name = "example_core",
-    input = "//python/tests/integration/core:test_core_example.py",
+    input = "//python/tests/integration/core:test_example.py",
     output = "python/core_example",
```
dmitrii-ubskii, line 68: Why no `.py`?

rust/src/common/error.rs
```diff
@@ -168,8 +169,8 @@ error_messages! { ConnectionError
         24: "Address translation map does not match the server's advertised address list. User-provided servers not in the advertised list: {unknown:?}. Advertised servers not mapped by user: {unmapped:?}.",
     ValueTimeZoneNameNotRecognised { time_zone: String } =
         25: "Time zone provided by the server has name '{time_zone}', which is not an officially recognized timezone.",
-    ValueTimeZoneOffsetNotImplemented { offset: i32 } =
-        26: "Time zone provided by the server has numerical offset '{offset}', which is not yet supported by the driver.",
+    ValueTimeZoneOffsetNotRecognised { offset: i32 } =
+        26: "Time zone provided by the server has numerical offset '{offset}', which is recognised as a valid value for offset in seconds.",
```
dmitrii-ubskii, line 173: You mean "_not_ recognised"?
```suggestion
    ValueTimeZoneOffsetNotRecognised { offset: i32 } =
        26: "Time zone provided by the server has numerical offset '{offset}', which is not recognised as a valid value for offset in seconds.",
```

dmitrii-ubskii (changes requested):

c/src/memory.rs
farost: You are totally right. If I remember correctly, I didn't want to use `release` for `Arc`, but for some reason used it to free the memory. Thanks!

c/src/error.rs
farost: Done.

python/BUILD
farost: Ideally, this file should not be exposed in the repo. Instead, it should be hidden in the sandbox, and the example is expected to be presented in the `README`. So I didn't want to make an impression like it's a real file to look at, more like a temporary configuration stuff.

We can the approach and make it a feature, though.

rust/src/common/error.rs
farost: Exactly! Thanks

dmitrii-ubskii (approved):
LGTM!

PR #698, opened 2024-10-17
Add concept documents for fetch queries

farost:
## Usage and product changes
We add concept documents to support the results of the reintroduced `fetch` queries.

In Rust, results of `fetch` are streams of `ConceptDocument`s. It is possible to work with the structured document as a Rust `struct`, but it's also possible to convert it to a `JSON` document and/or its `String` representation.

In Java, these results are presented as iterators over custom `JSON` class instances. A respective `toString` method is available.

In Python, these results are presented as iterators over standard `dict` instances (able to be printed).

Additionally, we add `QueryType` getters for general `QueryAnswer`s, so it's possible to check its type without collection.

Usage examples are shown in `README` for the Rust and the Python drivers. Example integration tests are also available for all 3 drivers. 

## Implementation
For testing purposes, example integration tests (with README examples) for the Rust and the Python drivers have been updated.
As the example tests for Java are not yet ready, an additional equivalent fetch step has been added.


dependencies/vaticle/repositories.bzl
```diff
@@ -28,7 +28,7 @@ def vaticle_typedb_protocol():
     git_repository(
         name = "vaticle_typedb_protocol",
         remote = "https://github.com/typedb/typedb-protocol",
-        tag = "3.0.0-alpha-3",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
+        commit = "94b79eb376e853a0a1c2b1ac838cb9d61f385100",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
```
farost, line 31: Will need to update it once the protocol is released, can be done after the merge.

java/api/answer/JSON.java
```diff
@@ -59,6 +59,10 @@ private static JSON of(JsonValue value) {
         }
     }
 
+    private static java.lang.String indent(java.lang.String string) {
```
farost, line 62: It was autoformatter lol

I haven't changed the class as it suits my current needs. I could remove some potentially redundant stuff, but we'll be working on different fetch options later, so we'll return here anyway, and there is no point in cleaning it right now. To my mind.

java/concept/answer/ConceptRowIteratorImpl.java
```diff
@@ -43,20 +42,12 @@ public ConceptRowIterator asConceptRows() {
 
     @Override
     public boolean hasNext() {
-        try {
```
farost, line 46: Redundant, the `try` block is already inside the `NativeIterator`

java/concept/answer/QueryAnswerImpl.java
```diff
@@ -19,24 +19,31 @@
...
+import static com.typedb.driver.jni.typedb_driver.query_answer_is_concept_row_stream;
 import static com.typedb.driver.jni.typedb_driver.query_answer_is_ok;
 
-public abstract class QueryAnswerImpl extends NativeObject<com.typedb.driver.jni.QueryAnswer> implements QueryAnswer {
+public abstract class QueryAnswerImpl implements QueryAnswer {
+    public final QueryType queryType;
+
     protected QueryAnswerImpl(com.typedb.driver.jni.QueryAnswer answer) {
-        super(answer);
+        queryType = QueryType.of(query_answer_get_query_type(answer));
```
farost, line 36: There was a bug when I initially implemented `QueryAnswer` as `NativeObject` and then proceeded to call `intoRows` inside `ConceptRowIteratorImpl`, which freed the `QueryAnswer`. 
Now, we don't store the native `QueryAnswer` anywhere on the Java side (same for Python), it's automatically unwrapped to something else, and that's it.

To access `QueryType` from non-consumed streams, I copy the small `enum` on the init, and we can continue doing it for any small metadata we want to expose. However, we might need to stop unwrapping the `QueryAnswer` to a specific ok/row/document on the init if we want to be able to call for these methods if more data appear (which would make the work with the native object even more complicated, so we leave it like this).

rust/src/answer/mod.rs
```diff
@@ -17,56 +17,115 @@
...
+    /// Retrieve the executed query's type (shared by all elements in this stream).
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// query_answer.get_query_type()
+    /// ```
+    pub fn get_query_type(&self) -> QueryType {
+        match &self {
+            QueryAnswer::Ok() => QueryType::SchemaQuery, // TODO: Get it from protocol
```
farost, line 51: I'll update it when we update the protocol (can be after this PR).

rust/src/concept/value.rs
```diff
@@ -539,9 +569,53 @@ impl FromStr for Duration {
     }
 }
 
+/// ISO-8601 compliant representation of a duration
 impl fmt::Display for Duration {
```
farost, line 573: The more I copy from the server for Durations, the more I want `typedb_common` =( (sarcasm (or not))

dmitrii-ubskii (approved):
LGTM!

PR #700, opened 2024-10-23
Introduce 3.0 Python driver docs and tests. Implement new fetch BDD steps in 3.0 drivers

farost:
## Usage and product changes
We introduce updated documentation, usage examples, and automated tests for the Java driver, to cover all the existing driver's APIs by sustainable validations.

Additionally, we implement additional BDD steps to check concept documents in BDDs for other 3.0 drivers: Rust and Python. 

As a small bonus, the protocol has been updated, and `Ok` `QueryAnswer`s receive correct `QueryType`s from the server instead of the client-side hardcode usage. 

## Implementation
For behaviour tests, the already existing architecture is preserved, and only the steps are rewritten to match the new definitions.
For integration tests, the new approach of the Python driver is chosen, and all the excessive tests are removed. Now, we only write `example` tests and tests to cover language-specific values (especially our implementations and their interfaces). 
For docs, the old instruments are used without changes.
For `README`, the `example` integration test parsing instrument is used as in Rust and Python.


docs/modules/ROOT/partials/java/answer/ConceptDocumentIterator.adoc
farost, line 1: I don't worry about the correctness of the generations as we don't publish it, but I hope that the content is fine and expect revalidation before the official publication.

java/test/integration/AddressTranslationTest.java
farost, line 1: I expected it to be implemented in BDDs when the cloud is released.

java/test/integration/concept/.gitkeep
```diff
@@ -1 +0,0 @@
-Placeholder for upcoming integration tests - delete this file once the directory has other files in it
```
farost, line 1: It didn't age well

java/test/integration/core/ExampleTest.java
```diff
@@ -0,0 +1,278 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.typedb.driver.test.integration.core;
+
+// EXAMPLE START MARKER
```
farost, line 22: The main issue here is that auto-formatter hates these comments between imports and tries to remove them. Don't see it as a huge issue, as we have PRs to update the docs (so we shouldn't miss it), but it's a bit irritating in the current implementation of my idea.

java/test/behaviour/query/QuerySteps.java
```diff
@@ -19,308 +19,980 @@
...
+            case ATTRIBUTE:
+                return concept.isAttribute();
+            case VALUE:
+                return concept.isValue();
+            default:
+                throw new AssertionError("Not covered ConceptKind: " + checkedKind);
+        }
+    }
+
+    // We want to call the checks on specific interfaces, not "Concept"s, so the boilerplate is needed
```
farost, line 166: I hate this and could write it much smaller, but it would require passing `Concept` interfaces everywhere, which is not transparent enough in terms of testing from my perspective. So, when the test says `get attribute type(p) get label`, I expect the test to literally get an instance of `AttributeType` and call its API (if it exists). 

In the end, I want to be straightforward (even if it means being really verbose) in these API checks for the drivers, otherwise, we lack coverage.

python/tests/integration/core/test_example.py
```diff
@@ -74,13 +74,14 @@ def test_example(self):
                 assert_that(answer.is_ok(), is_(True))
                 assert_that(answer.query_type, is_(QueryType.SCHEMA))
 
-                # Commit automatically closes the transaction. You can still safely call for it inside "with" blocks
+                # Commit automatically closes the transaction. It can still be safely called for inside "with" blocks
                 tx.commit()
```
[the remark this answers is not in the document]
dmitrii-ubskii: I've seen this in a few places now, in Java and python examples both.

dmitrii-ubskii (approved):

python/tests/integration/core/test_example.py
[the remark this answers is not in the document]
farost: My mistake! Thanks, fixed.

Yeah, I try to use the same comments if the behavior matches between drivers, that's why it's both in Java and Python

PR #705, opened 2024-11-01
Support CPython 3.13

sanmai-NL:
## Usage and product changes

Support [CPython 3.13](https://docs.python.org/3/whatsnew/3.13.html), which has been released on 7 October.

## Implementation

- Change the Bazel file.
- ...?

### DIFF
--- a/python/python_versions.bzl
+++ b/python/python_versions.bzl
@@ -48,6 +48,13 @@ python_versions = [
         "libpython": "@python312//:libpython",
         "suffix": "312",
     },
+    {
+        "name": "python313",
+        "python_version": "3.13",
+        "python_headers": "@python313//:python_headers",
+        "libpython": "@python313//:libpython",
+        "suffix": "313",
+    },
 ]
 
 def register_all_toolchains():


farost:
@sanmai-NL Hi, thank you for your contribution!

We need to update our core Bazel dependencies first, but we're already in the process of its verification and come back a bit later.

Just in case, do you need the driver for Python 3.13 to try 3.0-alpha or to work on production stuff? The PR is directed into `3.0` branch, but we can consider a special release of a stable 2.x version from `development` if you really need it.

sanmai-NL:
> @sanmai-NL Hi, thank you for your contribution!
> 
> We need to update our core Bazel dependencies first, but we're already in the process of its verification and come back a bit later.
> 
> Just in case, do you need the driver for Python 3.13 to try 3.0-alpha or to work on production stuff? The PR is directed into `3.0` branch, but we can consider a special release of a stable 2.x version from `development` if you really need it.

We actually use TypeDB 2.x in for a product where upgrading to Python 3.13 is desired and held back by this dependency, plus one other small dependency.

farost:
> > @sanmai-NL Hi, thank you for your contribution!
> We actually use TypeDB 2.x for a product where upgrading to Python 3.13 is desired 

That's nice! Unfortunately, the timing is a bit off for us at the moment. We are in the middle of finishing up version 3.0, and making the dependencies update required for the TypeDB 2.x Python 3.13 build is quite unsafe. 

What features of the driver do you use and how critical is the update for you this month? I suggest waiting until we release the stable version of 3.0 (in the coming months) with all the needed driver versions and API updates. You'd need to migrate your database using our tools and explore the updated syntax and constraints. Another noticeable thing that could require some refactoring from you is the elimination of Concept API (you'll have a single entry point of retrieving and updating your concepts through `transaction.query("...")`).

farost (approved):
Looking good, the dependencies are already prepared.

PR #709, opened 2024-11-04
Unify transaction cleanup in driver BDDs when opening new transactions

farost:
# Usage and product changes
We add unconditional transactions cleanup when opening new transactions in driver BDDs to allow writing BDD descriptions in a [direct usage fashion](https://github.com/typedb/typedb-behaviour/pull/310).


farost:
The CI won't work until we update artifacts with the new server.

flyingsilverfin (approved):

PR #710, opened 2024-11-08
Add timeout to channel builder

### LINKED ISSUE #672: GRPC calls may hang indefinitely in the event of a server fault
## Problem to Solve

Suppose there is some (any) kind of issue affecting the server, or the connection to it. In a recent incident we had a TypeDB Cloud cluster node that was not responding to the `user_token` GRPC request. This meant that `TypeDB.cloudDriver` (using, in our case, either the Java driver or Rust driver) would hang indefinitely, rather than throwing an error.

## Proposed Solution

The most obvious solution would be to add a timeout to GRPC calls in the Rust driver. This would need to be done with care, as long-running queries are legitimate.

## Additional Information

We ran the following test...
```rs
let connection = Connection::new_cloud_with_translation(
        [
            ("address1", "localhost:1729"),
            ("address2", "localhost:1730"),
            ("address3", "localhost:1731"),
        ]
        .into(),
        Credential::without_tls("username", "password")
    )
    .unwrap();
```

... with the following modifications to our source code (`println` statements)
```rs
/* connection/network/transmitter/rpc.rs */

    pub(in crate::connection) fn start_cloud(
        address: Address,
        credential: Credential,
        runtime: &BackgroundRuntime,
    ) -> Result<Self> {
        println!("{}", address.clone().to_string());
        let (request_sink, request_source) = unbounded_async();
        let (shutdown_sink, shutdown_source) = unbounded_async();
        runtime.run_blocking(async move {
            println!("a");
            let (channel, call_credentials) = open_callcred_channel(address, credential)?;
            println!("b");
            let rpc = RPCStub::new(channel, Some(call_credentials)).await;
            println!("c");
            tokio::spawn(Self::dispatcher_loop(rpc, request_source, shutdown_source));
            Ok::<(), Error>(())
        })?;
        Ok(Self { request_sink, shutdown_sink })
    }
```
```rs
/* connection/network/stub.rs */
    pub(super) async fn new(channel: Channel, call_credentials: Option<Arc<CallCredentials>>) -> Self {
        println!("d");
        let mut this = Self { grpc: GRPC::new(channel), call_credentials };
        println!("e");
        if let Err(err) = this.renew_token().await {
            warn!("{err:?}");
        }
        println!("f");
        this
    }

    async fn renew_token(&mut self) -> Result {
        if let Some(call_credentials) = &self.call_credentials {
            trace!("renewing token...");
            println!("g");
            call_credentials.reset_token();
            let req = user::token::Req { username: call_credentials.username().to_owned() };
            trace!("sending token request...");
            println!("h");
            let token = self.grpc.user_token(req).await?.into_inner().token;
            println!("i");
            call_credentials.set_token(token);
            trace!("renewed token");
            println!("j");
        }
        Ok(())
    }
```

This produced the following output ...
```
running 1 test
localhost:1730
a
b
d
e
g
h
test integration::network::address_translation has been running for over 60 seconds
```

This indicates that it hung at `self.grpc.user_token(req).await` in `renew_token`.

Naturally, you'd need a broken server to actually reproduce the issue; we hypothesise that it stops responding when there are too many concurrent connections to it.

Putting the server on a breakpoint might also work.

dmitrii-ubskii:
## Usage and product changes

We introduce a timeout in the gRPC channel so that the driver correctly fails when the server is available but unresponsive.

Resolves #672.

### DIFF
--- a/rust/src/connection/network/channel.rs
+++ b/rust/src/connection/network/channel.rs
@@ -17,7 +17,10 @@
  * under the License.
  */
 
-use std::sync::{Arc, RwLock};
+use std::{
+    sync::{Arc, RwLock},
+    time::Duration,
+};
 
 use tonic::{
     body::BoxBody,
@@ -49,8 +52,13 @@ impl GRPCChannel for PlainTextChannel {}
 
 impl GRPCChannel for CallCredChannel {}
 
+const TIMEOUT: Duration = Duration::from_secs(60);
+
 pub(super) fn open_plaintext_channel(address: Address) -> PlainTextChannel {
-    PlainTextChannel::new(Channel::builder(address.into_uri()).connect_lazy(), PlainTextFacade)
+    PlainTextChannel::new(
+        Channel::builder(address.into_uri()).timeout(TIMEOUT).connect_lazy(),
+        PlainTextFacade,
+    )
 }
 
 #[derive(Clone, Debug)]
@@ -66,7 +74,7 @@ pub(super) fn open_callcred_channel(
     address: Address,
     credential: Credential,
 ) -> Result<(CallCredChannel, Arc<CallCredentials>)> {
-    let mut builder = Channel::builder(address.into_uri());
+    let mut builder = Channel::builder(address.into_uri()).timeout(TIMEOUT);
     if credential.is_tls_enabled() {
         builder = builder.tls_config(credential.tls_config().clone().unwrap())?;
     }
--- a/rust/src/connection/network/channel.rs
+++ b/rust/src/connection/network/channel.rs
@@ -55,10 +55,7 @@ impl GRPCChannel for CallCredChannel {}
 const TIMEOUT: Duration = Duration::from_secs(60);
 
 pub(super) fn open_plaintext_channel(address: Address) -> PlainTextChannel {
-    PlainTextChannel::new(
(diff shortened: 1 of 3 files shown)


rust/src/connection/network/channel.rs
```diff
@@ -49,8 +52,13 @@ impl GRPCChannel for PlainTextChannel {}
 
 impl GRPCChannel for CallCredChannel {}
 
+const TIMEOUT: Duration = Duration::from_secs(60);
```
dmitrii-ubskii, line 55: Chosen arbitrarily. Anything less than a second or so is too strict for slow connections, but going significantly over a minute would probably be too long.

farost (commented):
Let's test queries

rust/src/connection/network/channel.rs
dmitrii-ubskii: The query does time out if the server does not respond within this time (tested by adding a `sleep(60000)` in `TransactionService::respond()` on the server side).

The default transaction timeout is 5 minutes. I'd extend this timeout to 10 minutes or even an hour.

Ideally we'd have a connection builder that would let the user specify the desired request timeout, but that would take a lot longer.

farost (changes requested):
Waiting for the timeout value update

PR #715, opened 2024-11-22
Speed up transaction opening and fix parallel 

farost:
## Usage and product changes
* We fix transaction opening for all the supported drivers, speeding up the operation by 2x.
* We eliminate database-related errors happening during concurrent database management (create and delete operations) by multiple drivers due to cache mismatches. 
* We make transaction opening requests immediately return errors instead of waiting for additional transaction operations to be performed (e.g. commit or query).

## Implementation
Details in PR comments to specific lines.

### DIFF
--- a/rust/src/connection/network/transmitter/rpc.rs
+++ b/rust/src/connection/network/transmitter/rpc.rs
@@ -157,7 +157,9 @@ impl RPCTransmitter {
                 match response_source.next().await {
                     Some(Ok(transaction::Server { server: Some(Server::Res(res)) })) => {
                         match TransactionResponse::try_from_proto(res) {
-                            Ok(TransactionResponse::Open { server_duration_millis }) => {
+                            Ok(TransactionResponse::Open { server_duration_millis: _ }) => {
+                                // TODO: Return as a part of Response::TransactionStream and use
+                                // the code below outside:
                                 // let open_latency =
                                 //     Instant::now().duration_since(open_request_start).as_millis() as u64 - server_duration_millis;
                                 // tracker.update_latency(open_latency)
@@ -171,9 +173,9 @@ impl RPCTransmitter {
                             }
                         }
                     }
-                    Some(Ok(unexpected_message)) => panic!("Unexpected message {unexpected_message:?}"),
+                    Some(Ok(unexpected_message)) => panic!("Unexpected message {unexpected_message:?}"), // TODO: return Err()
                     Some(Err(status)) => return Err(status.into()),
-                    None => panic!("Unexpected close"),
+                    None => panic!("Unexpected close"), // TODO: return Err()
                 }
                 Ok(Response::TransactionStream { open_request_id: req_id, request_sink, response_source })
             }
--- a/rust/src/connection/network/transmitter/rpc.rs
+++ b/rust/src/connection/network/transmitter/rpc.rs
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+use futures::StreamExt;
 use tokio::{
     select,
     sync::{
@@ -24,12 +25,13 @@ use tokio::{
         oneshot::channel as oneshot_async,
     },
 };
+use typedb_protocol::{transaction, transaction::server::Server};
 
 use super::{oneshot_blocking, response_sink::ResponseSink};
 use crate::{
     common::{address::Address, RequestID, Result},
     connection::{
-        message::{Request, Response},
+        message::{Request, Response, TransactionResponse},
         network::{
             channel::{open_callcred_channel, open_plaintext_channel, GRPCChannel},
             proto::{FromProto, IntoProto, TryFromProto, TryIntoProto},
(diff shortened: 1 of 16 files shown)


rust/src/connection/network/transmitter/rpc.rs
```diff
@@ -151,7 +153,28 @@ impl RPCTransmitter {
             Request::Transaction(transaction_request) => {
                 let req = transaction_request.into_proto();
                 let req_id = RequestID::from(req.req_id.clone());
-                let (request_sink, response_source) = rpc.transaction(req).await?;
+                let (request_sink, mut response_source) = rpc.transaction(req).await?;
+                match response_source.next().await {
+                    Some(Ok(transaction::Server { server: Some(Server::Res(res)) })) => {
+                        match TransactionResponse::try_from_proto(res) {
+                            Ok(TransactionResponse::Open { server_duration_millis }) => {
+                                // let open_latency =
```
farost, line 161: Will implement it in a half an hour... Does not seem used now though, can be left as a todo?

rust/src/connection/network/transmitter/rpc.rs
```diff
@@ -151,7 +153,28 @@ impl RPCTransmitter {
             Request::Transaction(transaction_request) => {
                 let req = transaction_request.into_proto();
                 let req_id = RequestID::from(req.req_id.clone());
-                let (request_sink, response_source) = rpc.transaction(req).await?;
+                let (request_sink, mut response_source) = rpc.transaction(req).await?;
+                match response_source.next().await {
```
farost, line 157: Maybe it worth making a separate function just to put a lable "initial response" on it

rust/src/connection/network/transmitter/rpc.rs
```diff
@@ -151,7 +153,28 @@ impl RPCTransmitter {
             Request::Transaction(transaction_request) => {
```
farost, line 153: We used to ignore errors in the "initial operation handler"  for transaction opening because we didn't know how to handle this in sync/async environment. However, these is a perfect place for these purposes: just don't return the result from the Transmitter until we get both responses.

rust/src/connection/network/transmitter/rpc.rs
```diff
@@ -151,7 +153,28 @@ impl RPCTransmitter {
...
+                            | Ok(TransactionResponse::Rollback)
+                            | Ok(TransactionResponse::Query(_))
+                            | Ok(TransactionResponse::Close) => {
+                                panic!("Unexpected - transaction open response was not TransactionOpen.")
+                            }
+                        }
+                    }
+                    Some(Ok(unexpected_message)) => panic!("Unexpected message {unexpected_message:?}"),
+                    Some(Err(status)) => return Err(status.into()),
+                    None => panic!("Unexpected close"),
```
farost, line 176: WIll make an error

rust/src/connection/network/transmitter/rpc.rs
```diff
@@ -151,7 +153,28 @@ impl RPCTransmitter {
...
+                            Err(error) => return Err(error),
+                            Ok(TransactionResponse::Commit)
+                            | Ok(TransactionResponse::Rollback)
+                            | Ok(TransactionResponse::Query(_))
+                            | Ok(TransactionResponse::Close) => {
+                                panic!("Unexpected - transaction open response was not TransactionOpen.")
+                            }
+                        }
+                    }
+                    Some(Ok(unexpected_message)) => panic!("Unexpected message {unexpected_message:?}"),
```
farost, line 174: Will make an error

rust/src/database/database.rs
```diff
@@ -67,6 +67,15 @@ impl Database {
         Ok(Self { name, replicas, server_connections })
     }
 
+    #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
+    pub(super) async fn get(name: String, server_connections: HashMap<Address, ServerConnection>) -> Result<Self> {
```
farost, line 71: I actually steal the code from 2.x. If a database is not in the cache, but it got created in the background, we need to "get" it instead of creating it.

rust/src/database/database_manager.rs
```diff
@@ -68,11 +68,18 @@ impl DatabaseManager {
...
         if !self.contains(name.to_owned()).await? {
             self.databases_cache.write().unwrap().remove(name);
-            return Err(ConnectionError::DatabaseDoesNotExist { name: name.to_owned() }.into());
+            return Err(ConnectionError::DatabaseNotFound { name: name.to_owned() }.into());
+        }
+
+        if let Some(cached_database) = self.try_get_cached(name) {
+            return Ok(cached_database);
         }
-        Ok(self.databases_cache.read().unwrap().get(name).unwrap().clone())
```
farost, line 75: We used to trust only us. Now we also read the remote.

rust/src/database/database_manager.rs
```diff
@@ -149,8 +155,18 @@ impl DatabaseManager {
 
     #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
     pub(crate) async fn get_cached_or_fetch(&self, name: &str) -> Result<Arc<Database>> {
-        let cached_database = self.databases_cache.read().unwrap().get(name).cloned();
-        Ok(cached_database.unwrap_or(self.get(name).await?))
+        match self.try_get_cached(name) {
```
farost, line 158: ty_get_cached helps us to release the lock.

rust/src/database/database_manager.rs
```diff
@@ -149,8 +155,18 @@ impl DatabaseManager {
 
     #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
     pub(crate) async fn get_cached_or_fetch(&self, name: &str) -> Result<Arc<Database>> {
-        let cached_database = self.databases_cache.read().unwrap().get(name).cloned();
-        Ok(cached_database.unwrap_or(self.get(name).await?))
```
farost, line 153: `unwrap_or` resolves the value inside of it, that's the place where we spent time.

farost (commented):
Comments

rust/src/connection/network/transmitter/rpc.rs
flyingsilverfin: we should still be figuring out the latency update I think, can you leave a comment `// TODO ` here?

flyingsilverfin (approved):

rust/src/connection/network/transmitter/rpc.rs
dmitrii-ubskii: We can return it as a part of `Response::TransactionStream`

farost: @dmitrii-ubskii Good point! Thanks, will do on Monday. Will leave a todo

PR #716, opened 2024-11-27
Core's Rust driver encryption configuration

lolski:
## Usage and product changes

Implement TLS encryption for the Core's Rust driver, which can be configured through the `ConnectionSettings` object.

## Implementation

- Propagate the TLS configuration in `ConnectionSettings` into GRPC channel


rust/src/connection/network/channel.rs
```diff
@@ -51,7 +51,10 @@ pub(super) fn open_callcred_channel(
     credential: Credential,
     connection_settings: ConnectionSettings,
 ) -> Result<(CallCredChannel, Arc<CallCredentials>)> {
-    let mut builder = Channel::builder(address.into_uri()); // TODO: configure TLS
+    let mut builder = Channel::builder(address.into_uri());
+    if connection_settings.is_tls_enabled() {
+        builder = builder.tls_config(connection_settings.tls_config().clone().unwrap())?;
```
farost, line 56: Didn't you say that TLS can be enabled even without a config?

farost: But we `unwrap` the user-specified config here, which is not correct. What are we going to do if we pass `(true, None)`?

farost (changes requested):
Didn't get it yet!

rust/src/connection/network/channel.rs
lolski: The TLS config object is unset if TLS is disabled, and set if TLS is enabled regardless of whether the CA is set or not. I've changed the `unwrap` to `expect` so it is clarified better to readers: 7dead0549d984c0c1ca4650d5df579f7577f0820.

lolski:
I've merged this PR since @farost is away, and that I believe the issue he raised has been addressed.

PR #718, opened 2024-11-29
Revert get current user API

lolski:
## Usage and product changes
Revert the `get_current_username` which returns the username as a string, back to `get_current_user` which returns the user object.

## Implementation
- Revert the `get_current_username` method in Rust, C, Java, and Python
- Update the BDD tests accordingly


dmitrii-ubskii (approved):
LGTM pending tests

PR #728, opened 2024-12-18
Optimise driver dispatch loop

flyingsilverfin:
## Usage and product changes

We optimise the driver dispatch loop: previously, we could dispatch at most 1000 serial messages per second, though in practice tokio would sometimes sleep longer than expected and the query and transaction latencies increased. This stems from Tokio's approximately millisecond resolution. By moving the collect-dispatch loop to a dedicated thread, we can now dispatch messages after microseconds of batching, improving overall driver performance significantly.

## Implementation

We move the dispatch loop to a dedicated thread, which can sleep for sub-millisecond resolution. We also make the background runtime a multi-threaded runtime for scalability.

### DIFF
--- a/rust/src/connection/network/transmitter/transaction.rs
+++ b/rust/src/connection/network/transmitter/transaction.rs
@@ -260,7 +260,7 @@ impl TransactionTransmitter {
     ) {
         const MAX_GRPC_MESSAGE_LEN: usize = 1_000_000;
         const DISPATCH_INTERVAL: Duration = Duration::from_micros(50);
-        const SLEEP_INTERVAL: Duration = Duration::from_micros(1);
+        const SLEEP_INTERVAL: Duration = Duration::from_micros(10);
         let mut next_dispatch = Instant::now() + DISPATCH_INTERVAL;
 
         let mut request_buffer = TransactionRequestBuffer::default();
--- a/rust/src/connection/network/transmitter/transaction.rs
+++ b/rust/src/connection/network/transmitter/transaction.rs
@@ -22,9 +22,9 @@ use std::{
     future::Future,
     pin::Pin,
     sync::{Arc, RwLock},
+    thread::sleep,
     time::Duration,
 };
-use std::thread::sleep;
 
 use crossbeam::{atomic::AtomicCell, channel::Sender};
 use futures::StreamExt;
@@ -243,13 +243,9 @@ impl TransactionTransmitter {
             callback_handler_sink,
         };
         let cloned_collector = collector.clone();
-        tokio::task::spawn_blocking(move || Self::dispatch_loop(
-            queue_source,
-            request_sink,
-            cloned_collector,
-            on_close_callback_source,
-            shutdown_signal,
-        ));
+        tokio::task::spawn_blocking(move || {
+            Self::dispatch_loop(queue_source, request_sink, cloned_collector, on_close_callback_source, shutdown_signal)
+        });
         tokio::spawn(Self::listen_loop(response_source, collector, shutdown_sink));
     }
 
@@ -265,7 +261,7 @@ impl TransactionTransmitter {
 
         let mut request_buffer = TransactionRequestBuffer::default();
         loop {
-            if let Ok(_) =  shutdown_signal.try_recv() {
+            if let Ok(_) = shutdown_signal.try_recv() {
                 if !request_buffer.is_empty() {
                     request_sink.send(request_buffer.take()).unwrap();
                 }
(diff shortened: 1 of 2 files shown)


rust/src/connection/runtime.rs
```diff
@@ -47,7 +47,7 @@ impl BackgroundRuntime {
     pub(crate) fn new() -> Result<Self> {
         let is_open = AtomicCell::new(true);
         let (shutdown_sink, mut shutdown_source) = unbounded_async();
-        let async_runtime = runtime::Builder::new_current_thread().enable_time().enable_io().build()?;
+        let async_runtime = runtime::Builder::new_multi_thread().enable_time().enable_io().build()?;
```
flyingsilverfin, line 50: @dmitrii-ubskii  do you think this is needed?

rust/src/connection/network/transmitter/transaction.rs
```diff
@@ -241,60 +242,46 @@ impl TransactionTransmitter {
...
+    fn dispatch_loop(
         mut request_source: UnboundedReceiver<(TransactionRequest, Option<ResponseSink<TransactionResponse>>)>,
         request_sink: UnboundedSender<transaction::Client>,
         mut collector: ResponseCollector,
         mut on_close_callback_source: UnboundedReceiver<Box<dyn FnOnce(Option<Error>) + Send + Sync>>,
         mut shutdown_signal: UnboundedReceiver<()>,
     ) {
         const MAX_GRPC_MESSAGE_LEN: usize = 1_000_000;
-        const DISPATCH_INTERVAL: Duration = Duration::from_millis(3);
+        const DISPATCH_INTERVAL: Duration = Duration::from_micros(10);
```
dmitrii-ubskii, line 260: Have you considered nixing this altogether and just sending the requests eagerly? That would remove the sleeps without us having to spawn another thread.

rust/src/connection/runtime.rs
dmitrii-ubskii: No. We're creating a new blocking thread, the number of runtime threads doesn't need to change.

rust/src/connection/network/transmitter/transaction.rs
```diff
@@ -241,60 +242,46 @@ impl TransactionTransmitter {
...
-            queue_source,
-            request_sink,
-            collector.clone(),
-            on_close_callback_source,
-            shutdown_signal,
-        ));
+        let cloned_collector = collector.clone();
+        tokio::task::spawn_blocking(move || {
+            Self::dispatch_loop(queue_source, request_sink, cloned_collector, on_close_callback_source, shutdown_signal)
+        });
```
dmitrii-ubskii, line 248: Minor, but this is the common way of moving a clone into a closure:
```suggestion
        tokio::task::spawn_blocking({
            let collector = collector.clone();
            move || Self::dispatch_loop(queue_source, request_sink, collector, on_close_callback_source, shutdown_signal)
        });
```

rust/src/connection/network/transmitter/transaction.rs
```diff
@@ -241,60 +242,46 @@ impl TransactionTransmitter {
...
-                        }
-                        request_buffer.push(request);
-                    } else {
-                        break;
-                    }
+                request_buffer.push(request);
+            } else if let Ok(callback) = on_close_callback_source.try_recv() {
+                collector.on_close.write().unwrap().push(callback)
+            } else {
+                sleep(DISPATCH_INTERVAL);
```
dmitrii-ubskii, line 282: Wait, so we're not buffering here at all? Pseudocode:
```
loop {
    if shutdown { send(buffer); break; }
    else if request { buffer.push(request); }
    else if callback { register(callback); }
    else { sleep(10us); send(buffer); }
}
``` 
Unless the requests are sent down the channel before this thread gets the chance to sleep, we don't have more than one request at a time, and in either case the sleep is basically redundant. All it does is limit the send rate somewhat.

You could sleep for 1us, and send when dispatch interval is exceeded, but since there was apparently no downside to this I'd like to see the `async` version with no buffering explored as well. I'd prefer not spawning another thread if we can help it.

dmitrii-ubskii (changes requested):

rust/src/connection/network/transmitter/transaction.rs
flyingsilverfin: Good point - I do want to keep the buffering because i know from way back when Haikal did his testing on throughput, without batching of messages we weren't able to maximise loading on the server side. What i'm changing it to is sleep for 1 microsecond, batching up to 20 microseconds worth of stuff (20 messages) , and then send.

dmitrii-ubskii (approved):

PR #733, opened 2025-01-07
Prepend current Docs version to all Docs resource URLs

alexjpwalker:
## Usage and product changes

All resource URLs in our 2.x Docs now include the `{page-version}` tag.

## Motivation

This is necessary in order to support multi-versioned Docs. Without this tag, URLs are always assumed to point to the latest version - 3.x at the time of writing. This would make 2.x docs incorrect and/or unbuildable.

## Implementation

All resource URLs in our 2.x Docs now include the `{page-version}` tag.

(N.B: these Docs pages were originally handwritten - and we're keeping it that way.)

### DIFF
--- a/docs/modules/ROOT/partials/c/api-reference.adoc
+++ b/docs/modules/ROOT/partials/c/api-reference.adoc
@@ -2,85 +2,85 @@
 [#_connection_header]
 == Connection
 
-include::{page-component-version}@api-ref::partial$c/connection/connection.adoc[]
+include::{page-version}@api-ref::partial$c/connection/connection.adoc[]
 
-include::{page-component-version}@api-ref::partial$c/connection/credential.adoc[]
+include::{page-version}@api-ref::partial$c/connection/credential.adoc[]
 
-include::{page-component-version}@api-ref::partial$c/connection/database.adoc[]
+include::{page-version}@api-ref::partial$c/connection/database.adoc[]
 
-include::{page-component-version}@api-ref::partial$c/connection/replica.adoc[]
+include::{page-version}@api-ref::partial$c/connection/replica.adoc[]
 
-include::{page-component-version}@api-ref::partial$c/connection/user.adoc[]
+include::{page-version}@api-ref::partial$c/connection/user.adoc[]
 
 [#_session_header]
 == Session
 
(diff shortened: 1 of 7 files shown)


docs/modules/ROOT/partials/java/api-reference.adoc
```diff
@@ -2,112 +2,112 @@
...
+include::{page-version}@api-ref::partial$java/connection/Database.Replica.adoc[]
 
-include::api-ref::partial$java/connection/UserManager.adoc[]
+include::{page-version}@api-ref::partial$java/connection/UserManager.adoc[]
 
-include::api-ref::partial$java/connection/User.adoc[]
+include::{page-version}@api-ref::partial$java/connection/User.adoc[]
 
 [#_session_header]
 == Session
```
farost, line 22: Oh, I need to update these, too.....

farost: I guess I'll do it in a separate PR

docs/modules/ROOT/partials/c/api-reference.adoc
```diff
@@ -2,85 +2,85 @@
 [#_connection_header]
 == Connection
 
-include::api-ref::partial$c/connection/connection.adoc[]
+include::{page-version}@api-ref::partial$c/connection/connection.adoc[]
```
farost, line 5: Where is this `page-version` declared? Is it used based on the var from the docs repo?

farost (approved):

docs/modules/ROOT/partials/c/api-reference.adoc
alexjpwalker: `{page-version}` is an alias for `{page-component-version}`, an Antora built-in variable that is present on every page of an Antora site and refers to the currently selected version. For us, it'll always be either `3.x` or `2.x`, with `3.x` being the default / latest.

farost: Hmmmmm okay. I still wonder how we integrate these with the docs. Thus, how to make it different for 2.x and 3.x docs (so the current `development` produces 2.x references and the current `3.0` produces 3.x references)

PR #735, opened 2025-01-14
Introduce 3.0 cloud driver tests

farost:
## Usage and product changes
We fix cloud drivers creation interfaces and introduce integration and behavior tests to cover these public methods.

## Implementation
The integration tests are currently just a copypaste of the core integration tests with another interface. Not sure if we want to change it, but it would be smarter to reuse this code differently in the future.

The behavior tests are actually currently similar to the core tests: they don't use encryption and they only use one node. We'll need to update it with more nodes in action when it's ready. Encryption can be added earlier in a separate PR: there are all the relevant TODOs in the code.

### DIFF
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -97,12 +97,13 @@ build:
           export CORE_FAILED= || export CORE_FAILED=1
         tool/test/stop-core-server.sh
         if [[ -n "$CORE_FAILED" ]]; then exit 1; fi
-
-#        source tool/test/start-cloud-servers.sh 3 && # use source to receive export vars
-#          bazel test //rust/tests/integration/cloud/... --test_output=streamed &&
-#          export CLOUD_FAILED= || export CLOUD_FAILED=1
-#        tool/test/stop-cloud-servers.sh
-#        if [[ -n "$CLOUD_FAILED" ]]; then exit 1; fi
+        
+        # TODO: Use cloud server artifact with 3 nodes when available
+        tool/test/start-core-server.sh &&
+          bazel test //rust/tests/integration/cloud/... --test_output=streamed --test_arg=--nocapture &&
+          export CLOUD_FAILED= || export CLOUD_FAILED=1
+        tool/test/stop-core-server.sh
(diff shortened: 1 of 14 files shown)


.factory/automation.yml
```diff
@@ -97,12 +97,13 @@ build:
...
         if [[ -n "$CORE_FAILED" ]]; then exit 1; fi
-
-#        source tool/test/start-cloud-servers.sh 3 && # use source to receive export vars
-#          bazel test //rust/tests/integration/cloud/... --test_output=streamed &&
-#          export CLOUD_FAILED= || export CLOUD_FAILED=1
-#        tool/test/stop-cloud-servers.sh
-#        if [[ -n "$CLOUD_FAILED" ]]; then exit 1; fi
+        
+        # TODO: Use cloud server artifact with 3 nodes when available
+        tool/test/start-core-server.sh &&
```
flyingsilverfin, line 102: Can we use `start-cloud-server` script?

java/test/behaviour/connection/ConnectionStepsCloud.java
```diff
@@ -18,107 +18,102 @@
...
+    }
+
+    @After
+    public synchronized void after() {
+        super.after();
+    }
+
+    @Override
+    Driver createTypeDBDriver(String address, Credentials credentials, DriverOptions driverOptions) {
+        return TypeDB.coreDriver(address, credentials, driverOptions);
```
flyingsilverfin, line 49: So cloudDriver entry point doesn't work at all yet?

.factory/automation.yml
farost: I don't see a need. I don't want to comment out the smart logic it has for connections to multiple endpoints, etc. And we don't have a cloud artifact released to cloudsmith! So there is totally no difference at the moment. 
We can use a temporary `start-cloud-server` script (not serverS) when we at least have the cloud artifacts

java/test/behaviour/connection/ConnectionStepsCloud.java
farost: It's me forgetting to change this line. But it works the same lol

PR #738, opened 2025-01-28
Introduce optional concepts to Concept Rows. Cleanup driver errors. Automate README examples updates.

farost:
## Usage and product changes
* **Introduce optional concepts to Concept Rows**
`get()` and `get_index()` interfaces of Concept Rows always return optional Concepts. If the requested variable name or index exists in the column names of a row, but the actual value for this variable is empty, an empty optional value is returned. This is a natural behavior for optionals (coming to TypeDB soon!) and is already useful for queries like (where the variable `$empty` won't have values):
```
match not {$empty isa user;}; insert $u isa user, has username "Hi";
```

If the requested variable name or index does not exist in the column names of a row, an error is returned.

* **Cleanup driver errors**
Query errors are dissolved in the Java and Python drivers. 
All the arguments passed to the external interfaces are validated to be non-null and of the correct format (e.g., non-negative for `concept_row.get_index(column_index)`).

* **Automate README examples updates**
Cloud driver usage examples are added to all the available READMEs. Additionally, these examples are officially available in the repo as separate files. 

The process of README examples updating is automated by unifying all language updates in a single script. A new CI job is introduced to verify that the README examples are up to date.


dependencies/typedb/repositories.bzl
```diff
@@ -32,8 +32,9 @@ def typedb_protocol():
     )
 
 def typedb_behaviour():
+    # TODO: Return typedb
```
farost, line 35: https://github.com/typedb/typedb-behaviour/pull/328

java/BUILD
```diff
@@ -135,36 +135,56 @@ javadoc_to_adoc(
 test_to_example(
     name = "example_core",
     input = "//java/test/integration/core:ExampleTest.java",
-    output = "java/core_example",
+    output = "java/TypeDBCoreExample.java",
```
farost, line 138: Dmitrii (I think) had asked why we don't really call these files like real examples with extensions. I decided to simplify our work and do so, letting these files exist separately. 
Now, I left it right in the root folder for each language. Shall I create an `example` folder for each language and start placing these files there?

java/BUILD
```diff
@@ -135,36 +135,56 @@ javadoc_to_adoc(
...
+    output = "java/TypeDBCloudExample.java",
+    removed_lines = ["assert", "fail(", "todo"],
+    changed_words = {
+        "com.typedb.driver.test.integration.cloud": "com.typedb.driver",
+        "ExampleTest": "TypeDBCloudExample",
     },
     header_comment_sign = "//",
 )
 
-# Temporary manual usage:
```
farost, line 146: I've introduced a new `tool` script, so it partially automates this manual usage. No need to worry about connecting these two rules (`test_to_example` + `update_markdown_example`) now.

java/README.md
```diff
@@ -55,11 +55,237 @@ Further documentation: https://typedb.com/docs/drivers/java/overview
 
 ## Example usage
 
-### TypeDB Core
+### TypeDB Cloud / Enterprise (Cloud driver)
```
farost, line 58: Placed the Cloud example first (as it's the main product now). We can simplify it later once we decide what to do with the connection API.

Currently, the cloud example contains everything from the core example (every schema/read/write query, etc), the only difference is in the driver setup. I was thinking about simplifying it, but it would be irritating for either cloud or core user to switch between examples, and it's easier to copy the thing you're interested in, so I've made it this way.

HOWEVER, the README looks big now. I can, say, make the Core example collapsible. It does look strange though, and I'm not sure if we want to collapse both examples: we kinda want to showcase what the code looks like. 

Thoughts on these topics?

java/common/exception/ErrorMessage.java
```diff
@@ -80,9 +80,13 @@ public static class Driver extends ErrorMessage {
...
         public static final Driver POSITIVE_VALUE_REQUIRED =
-                new Driver(8, "Value cannot be less than 1, was: '%d'.");
+                new Driver(8, "Value of '%s' should be positive, was: '%d'.");
+        public static final Driver NON_NEGATIVE_VALUE_REQUIRED =
+                new Driver(9, "Value of '%s' should be non-negative, was: '%d'.");
+        public static final Driver NON_NULL_VALUE_REQUIRED =
+                new Driver(10, "Value of '%s' should not be null.");
         public static final Driver UNIMPLEMENTED =
-                new Driver(9, "This operation is not implemented yet.");
+                new Driver(11, "This operation is not implemented yet.");
```
farost, line 89: I don't think anybody relies on these error codes yet. But I can stop doing it and change the enumeration back if needed.

java/common/exception/ErrorMessage.java
```diff
@@ -110,20 +112,6 @@ public static class Concept extends ErrorMessage {
         }
     }
 
-    public static class Query extends ErrorMessage {
-        public static final Query VARIABLE_DOES_NOT_EXIST =
```
farost, line 114: It comes from the driver now.

java/common/exception/ErrorMessage.java
```diff
@@ -110,20 +112,6 @@ public static class Concept extends ErrorMessage {
...
-                new Query(1, "The variable '%s' does not exist.");
-        public static final Query MISSING_QUERY =
```
farost, line 116: None => specific NonNone DriverError. 
Empty => Error from the server (always should be)

python/core_example.py
```diff
@@ -3,6 +3,8 @@
 from typedb.driver import *
 
 
+class TypeDBExample:
```
farost, line 6: For indentation purposes, otherwise, the function floats strangely

java/README.md
flyingsilverfin: Let's just keep both open for the time being.

java/common/exception/ErrorMessage.java
flyingsilverfin: Reorder is fine

java/BUILD
flyingsilverfin: I was actually thinking that having the root contain the example is a bit awkward. However, having it as a single file in `example` is a bit meh as well. I don't have a strong opinion, can you maybe check some other github repos for peple publishing libraries/drivers to see if there's a common theme?

flyingsilverfin (approved):

java/BUILD
farost: > I was actually thinking that having the root contain the example is a bit awkward. However, having it as a single file in `example` is a bit meh as well. I don't have a strong opinion, can you maybe check some other github repos for peple publishing libraries/drivers to see if there's a common theme?

There are two files per repo now! One for Core, one for Cloud.

farost: I'll refactor it in a separate PR if we feel like we want to

PR #741, opened 2025-02-24
Introduce a single driver creation endpoint for all editions of TypeDB

farost:
## Usage and product changes
Introduce a single driver creation endpoint for all editions of TypeDB: all `new_core`, `new_cloud`, `TypeDB.core`, `TypeDB.cloud`, and other alternatives in TypeDB drivers now have a single `new` / `driver` method that accepts a single string as an address.

Use it for any edition of TypeDB 3.x (Community Edition, Cloud, Enterprise) the following way (check out `README` or driver documentation for full usage examples):
Rust:
```rust
let driver = TypeDBDriver::new(
    TypeDBDriver::DEFAULT_ADDRESS,
    Credentials::new("admin", "password"),
    DriverOptions::new(false, None).unwrap(),
)
.await
.unwrap();
```

Python:
```py
driver = TypeDB.driver(TypeDB.DEFAULT_ADDRESS, Credentials("admin", "password"), DriverOptions())
```

Java:
```java
Driver driver = TypeDB.driver(TypeDB.DEFAULT_ADDRESS, new Credentials("admin", "password"), new DriverOptions(false, null));
```

Currently, TypeDB 3.x supports only a single server address, so the list-based method overloading is unnecessary. We plan to preserve the same simplified API after introducing multi-node servers, extending the accepted formats of the input address string instead. Stay tuned for details!

## Implementation
* All excessive driver constructors are removed. Only `new` or `driver` are left and receive a single string address. 
* The Rust and the C drivers still have `new` and `new_with_description` methods. The latter is used to specify a custom driver language, and this intent is highlighted in the docs now.
* Similarly, the C driver acquired a dedicated endpoint for constructing the C driver with the predefined language. It is not exposed in SWIG, so only `new_with_description` can be used by SWIG-based drivers.
* Renamed `core` -> `community`
* Renamed `cloud` -> `cluster`
* Example tests and examples are unified: now, all editions of TypeDB use the same set of interfaces. Separate examples specifically for `cluster`'s features can be written in the future, not affecting these common files. The package structures are adjusted respectively.
* Updated docs and `README`s.
* Fixed minor test environment bugs.


.factory/automation.yml
```diff
@@ -120,28 +120,28 @@ build:
...
-        bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
-        bazel run @typedb_dependencies//distribution/artifact:create-netrc
-
-        # TODO: Use cloud server artifact with 3 nodes when available
-        tool/test/start-core-server.sh &&
-          bazel test //rust/tests/behaviour/...  --//rust/tests/behaviour/config:mode=cloud --test_output=streamed --jobs=1 &&
-          export TEST_SUCCESS=0 || export TEST_SUCCESS=1
-        tool/test/stop-core-server.sh
-        exit $TEST_SUCCESS
+    # TODO: Use cluster server artifact with 3 nodes when available (it would do the same thing as community now)
```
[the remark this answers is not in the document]
farost: We can uncomment it when we at least use the cluster's server. But we only publish docker images, not artifacts. We may want to finally prepare the artifacts as well (at least they have different `main`s).

python/tests/behaviour/background/cluster/environment.py
```diff
@@ -19,18 +19,17 @@
...
 
 def before_all(context: Context):
     environment_base.before_all(context)
-    # context.credentials_root_ca_path = os.environ["ROOT_CA"] # TODO: test root ca with cloud
+    # context.credentials_root_ca_path = os.environ["ROOT_CA"] # TODO: test root ca with cluster
     context.create_driver_fn = lambda host="localhost", port=None, user=None, password=None: \
         create_driver(context, host, port, user, password)
     context.setup_context_driver_fn = lambda host="localhost", port=None, username=None, password=None: \
         setup_context_driver(context, host, port, username, password)
-    context.setup_context_driver_fn()
```
farost, line 33: We only need to call it when it's called in steps

python/tests/behaviour/connection/connection_steps.py
```diff
@@ -66,10 +66,8 @@ def step_impl(context: Context):
 
 @step("connection is open: {is_open:Bool}")
 def step_impl(context: Context, is_open: bool):
-    # real_is_open = hasattr(context, 'driver') and context.driver and context.driver.is_open()
```
farost, line 69: It is a bad habit of us to just leave these forgotten todos. I didn't even know that something was commented out. This could easily lead to a missed bug (we even had such a bug in the Python 2.x driver for years!!!!!!)

farost: It was a test setup. bug, fixed in the `environment.py` files

java/test/behaviour/connection/ConnectionStepsBase.java
```diff
@@ -72,6 +73,15 @@ void beforeAll() {
     } // Can add "before all" setup steps here
 
     void before() {
+        try {
+            // Sleep between scenarios to let the driver close completely
+            // (`close` is not synced and can cause lock failures in CI)
+            // TODO: This might be a bug requiring an investigation. For some reason, it only happens in Java (even not in Python!)
```
farost, line 79: If I'm not mistaken, we had a similar issue in 2.x, and nobody ever noticed anything similar in prod. I think that it's really due to the Java test executor being too fast and eager (as we don't wait until the connection is truly closed), that's why we sometimes fail in CI. It's just sus that it does not happen in Python, so I left a non-prioritized TODO.

python/tests/behaviour/connection/connection_steps.py
flyingsilverfin: Totally agree the TODOs are a bad idea

flyingsilverfin (approved):

PR #743, opened 2025-03-13
Substitute credentials authentication by token authentication with auto-renewal 

farost:
## Usage and product changes
Upgrade [TypeDB protocol](https://github.com/typedb/typedb-protocol/pull/216) to use authentication tokens instead of passing credentials with every API call. 
When sending gRPC requests, drivers include temporary tokens in the request headers, limiting password usage to `connection_open` and token renewal calls, providing better security for sensitive data.

The authentication and token renewal mechanisms are fully automatic and do not require any actions from TypeDB driver users. Thus, the API remains unchanged.

## Implementation
Credential injectors start injecting tokens instead of credentials. 
These tokens are acquired when: 
* user creates a new driver (`connection_open`);
* a request fails because of the token expiration, and we automatically retry this request once with a prior `sign_in` call. This operation is expected to be insignificant in terms of performance effect since the users are not expected to set token TTLs to too low values.


rust/src/connection/network/stub.rs
```diff
@@ -45,15 +48,41 @@ impl<Channel: GRPCChannel> RPCStub<Channel> {
...
     }
 
-    async fn call<F, R>(&mut self, call: F) -> Result<R>
+    async fn call_with_auto_renew_token<F, R>(&mut self, call: F) -> Result<R>
     where
         for<'a> F: Fn(&'a mut Self) -> BoxFuture<'a, Result<R>>,
     {
-        call(self).await
+        match call(self).await {
+            Err(Error::Connection(ConnectionError::TokenCredentialInvalid)) => {
```
farost, line 56: If it failed specifically, renew the token and retry.

rust/src/connection/network/stub.rs
```diff
@@ -45,15 +48,41 @@ impl<Channel: GRPCChannel> RPCStub<Channel> {
...
+            let token = self.grpc.sign_in(request).await?.into_inner().token;
+            call_credentials.set_token(token);
+            trace!("Token renewed");
+        }
+        Ok(())
     }
 
     pub(super) async fn connection_open(&mut self, req: connection::open::Req) -> Result<connection::open::Res> {
-        self.single(|this| Box::pin(this.grpc.connection_open(req.clone()))).await
+        let result = self.single(|this| Box::pin(this.grpc.connection_open(req.clone()))).await;
```
farost, line 78: Sneak into the response to get the token on the `network` level without asking the more business-like logic to work with it for us.

tool/test/start-community-server.sh
```diff
@@ -22,6 +22,7 @@ rm -rf typedb-all
 
 bazel run //tool/test:typedb-extractor -- typedb-all
 BAZEL_JAVA_HOME=$(bazel run //tool/test:echo-java-home)
+# TODO: Can add `--server.authentication.token_ttl_seconds 15` to test token auto-renewal in BDDs!
```
farost, line 25: I don't have the artifacts (I just tested locally), so this feature can be good. We have small integration tests, some small behaviour tests, and some big behaviour tests. Everything will be covered!

farost:
It's expected to be red as I don't have the new server artifacts. It's green locally.

rust/src/common/error.rs
```diff
@@ -275,6 +275,16 @@ impl Error {
         }
     }
 
+    fn try_extracting_connection_error(status: &Status, code: &str) -> Option<ConnectionError> {
+        // TODO: We should probably catch more connection errors instead of wrapping them into
+        // ServerErrors. However, the most valuable information even for connection is inside
+        // stacktraces now.
+        match code {
+            "AUT3" => Some(ConnectionError::TokenCredentialInvalid {}),
```
flyingsilverfin, line 283: Ouu ok make a comment in the server-side that we depend on those error messages client-side and not to change them randomly -- if we don't already have that warning!

rust/src/connection/network/channel.rs
```diff
@@ -65,20 +66,33 @@ pub(super) fn open_callcred_channel(
...
+        *self.token.write().expect("Expected token write lock acquisition on reset") = None;
     }
 
     pub(super) fn inject(&self, mut request: Request<()>) -> Request<()> {
-        request.metadata_mut().insert("username", self.credentials.username().try_into().unwrap());
-        request.metadata_mut().insert("password", self.credentials.password().try_into().unwrap());
+        if let Some(token) = &*self.token.read().expect("Expected token read lock acquisition on inject") {
+            request.metadata_mut().insert(
+                "authorization",
+                format!("Bearer {}", token).try_into().expect("Expected authorization header formatting"),
```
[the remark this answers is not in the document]
flyingsilverfin: I don't have a strong knowledge/opinion on this, maybe @lolski  has some?

rust/src/common/error.rs
farost: We did a similar thing (even for a bigger number of errors) in 2.x in other domains. But I will.

PR #745, opened 2025-03-25
Replay JNI loader fixes

krishnangovindraj:
## Usage and product changes
Replays (#702), updating the jni libraries to include the platform in their filename. 

e.g. `libtypedb_driver_jni-linux-arm64.so` in place of `libtypedb-driver-jni.so` 

 Since the main maven package for the driver depends on all the jni packages, flattening these jars would lead to two`libtypedb-driver.so` being extracted - only one of which survives - leading to errors when the loader attempts to load it.

## Implementation
Check out the files from #702, since nothing on 3.0 had changed since.
Looked at the local jar contents with `jar -tf`

### DIFF
--- a/java/BUILD
+++ b/java/BUILD
@@ -71,6 +71,7 @@ deploy_maven(
 
 swig_native_java_library(
     name = "typedb_driver_jni",
+    library_name_with_platform = "typedb_driver_jni-{platform}",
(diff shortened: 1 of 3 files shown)


flyingsilverfin (approved):
thanks!

PR #746, opened 2025-03-26
Update artifacts and dependencies to test correct transactions management in the updated server's gRPC service

farost:
## Usage and product changes
Update artifacts and dependencies refs.


python/tests/behaviour/config/parameters.py
```diff
@@ -190,8 +191,10 @@ def parse_query_type(value: str) -> QueryType:
 register_type(QueryType=parse_query_type)
 
 
-def parse_list(table: Table) -> list[str]:
-    return [table.headings[0]] + list(map(lambda row: row[0], table.rows))
```
farost, line 194: For Scenario Outlines, it produced `[<sub me>, val, val, val, val]`

PR #749, opened 2025-04-03
Introduce Rust driver token-based authentication

farost:
## Usage and product changes
We update the protocol version and introduce token-based authentication for all the available drivers. Now, instead of sending usernames and passwords for authentication purposes, authorization tokens are implicitly added to every network request to a TypeDB server. It enhances the authentication speed and security. 

These tokens are acquired as a result of driver instantiation and are renewed automatically. This feature does not require any user-side changes.

Additionally, as a part of the HTTP client introduction work, we rename Concept Documents key style from snake_case to camelCase, which is a common convention for JSONs (it affects only value types: `value_type` -> `valueType`).

## Implementation
Every network call adds tokens instead of user credentials and is repeated after a token renewal request if it receives an authentication error.

### DIFF
--- a/rust/src/connection/network/stub.rs
+++ b/rust/src/connection/network/stub.rs
@@ -67,7 +67,7 @@ impl<Channel: GRPCChannel> RPCStub<Channel> {
             trace!("Renewing token...");
             call_credentials.reset_token();
             let request = call_credentials.credentials().clone().try_into_proto()?;
-            let token = self.grpc.sign_in(request).await?.into_inner().token;
+            let token = self.grpc.authentication_token_create(request).await?.into_inner().token;
             call_credentials.set_token(token);
             trace!("Token renewed");
         }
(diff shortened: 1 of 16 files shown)


rust/src/connection/network/stub.rs
```diff
@@ -45,15 +48,41 @@ impl<Channel: GRPCChannel> RPCStub<Channel> {
...
+            Err(Error::Connection(ConnectionError::TokenCredentialInvalid)) => {
+                debug!("Request rejected because token credential was invalid. Renewing token and trying again...");
+                self.renew_token().await?;
+                call(self).await
+            }
+            res => res,
+        }
+    }
+
+    async fn renew_token(&mut self) -> Result {
```
flyingsilverfin, line 65: good!

tool/test/start-community-server.sh
```diff
@@ -22,6 +22,7 @@ rm -rf typedb-all
 
 bazel run //tool/test:typedb-extractor -- typedb-all
 BAZEL_JAVA_HOME=$(bazel run //tool/test:echo-java-home)
+# TODO: Can add `--server.authentication.token_ttl_seconds 15` to test token auto-renewal in BDDs!
```
flyingsilverfin, line 25: Go ahead!

flyingsilverfin (approved):

flyingsilverfin (approved):

PR #751, opened 2025-04-15
Update dependencies

farost:


dmitrii-ubskii:
For future reference:

After a recent update of the crates, rustls added aws-lc-rs as a default dependency. This caused runtime issues during loading of the shared library where the dynamic linker could not find some symbols from aws-lc: ```
ImportError: dlopen(.../native_driver_python.so, 0x0002): symbol not found in flat namespace '_aws_lc_0_28_0_EVP_aead_aes_128_gcm'
```

We decided to force the use of ring as the cryptographic provider for rustls instead.

PR #753, opened 2025-04-24
Introduce transaction and query options

farost:
## Usage and product changes
**Introduce transaction options:**
* `transaction_timeout`: If set, specifies a timeout for killing transactions automatically, preventing memory leaks in unclosed transactions,
* `schema_lock_acquire_timeout`: If set, specifies how long the driver should wait if opening a transaction is blocked by an exclusive schema write lock.

Rust examples:
```rust
let options = TransactionOptions::new().transaction_timeout(Duration::from_secs(10));
let transaction =
    driver.transaction_with_options(database.name(), TransactionType::Schema, options).await.unwrap();
```

Python example:
```py
options = TransactionOptions(transaction_timeout_millis=10_000)
tx = driver.transaction(database.name, TransactionType.SCHEMA, options)
```

Java example:
```java
TransactionOptions transactionOptions = new TransactionOptions().transactionTimeoutMillis(10_000);
Transaction transaction = driver.transaction(database.name(), Transaction.Type.SCHEMA, transactionOptions);
```

**Introduce query options:**
* `include_instance_types`: If set, specifies if types should be included in instance structs returned in ConceptRow answers.

Rust examples:
```rust
let options = QueryOptions::new().include_instance_types(true);
let answer = transaction.query_with_options("match $x isa person;", options).await.unwrap();
```

Python example:
```py
options = QueryOptions(include_instance_types=True)
answer = tx.query("match $x isa person;").resolve()
```

Java example:
```java
QueryOptions queryOptions = new QueryOptions().includeInstanceTypes(true);                
QueryAnswer matchAnswer = transaction.query("match $x isa person;", queryOptions).resolve();                
```

## Implementation
Add new structs and classes for `TransactionOptions` and `QueryOptions` and extend `transaction` / `query` interfaces to accept optional options.

Add relevant integration and behaviour tests.


java/api/Transaction.java
```diff
@@ -67,6 +67,20 @@ public interface Transaction extends AutoCloseable {
     @CheckReturnValue
     Promise<? extends QueryAnswer> query(String query) throws com.typedb.driver.common.exception.TypeDBDriverException;
```
flyingsilverfin, line 68: I believe you could make this a `default` method that calls the one with options (with default options) so you don't have to implement twice?

java/connection/TransactionImpl.java
```diff
@@ -81,9 +83,14 @@ public boolean isOpen() {
 
     @Override
     public Promise<? extends QueryAnswer> query(String query) throws TypeDBDriverException {
+        return query(query, new QueryOptions());
```
flyingsilverfin, line 86: Ah ok you've done that here, works for me

PR #756, opened 2025-05-02
Fix incorrect resource ownership transmission in schema retrieval C functions

farost:
## Usage and product changes
Fix a crash caused by an incorrect database's ownership acquisition in the C layer of the `schema` and `type_schema` functions, affecting Python and Java drivers.

Add respective BDD steps implementations for these functions in Rust, Python, and Java.

## Implementation
Borrow instead of taking ownership of the referenced database resource.

### DIFF
--- a/c/src/database.rs
+++ b/c/src/database.rs
@@ -49,13 +49,13 @@ pub extern "C" fn database_delete(database: *const Database) {
 /// A full schema text as a valid TypeQL define query string.
 #[no_mangle]
 pub extern "C" fn database_schema(database: *const Database) -> *mut c_char {
-    try_release_string(take_arc(database).schema())
+    try_release_string(borrow(database).schema())
 }
 
 /// The types in the schema as a valid TypeQL define query string.
 #[no_mangle]
 pub extern "C" fn database_type_schema(database: *const Database) -> *mut c_char {
-    try_release_string(take_arc(database).type_schema())
+    try_release_string(borrow(database).type_schema())
 }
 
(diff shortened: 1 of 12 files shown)


c/src/database.rs
```diff
@@ -49,13 +49,13 @@ pub extern "C" fn database_delete(database: *const Database) {
 /// A full schema text as a valid TypeQL define query string.
 #[no_mangle]
 pub extern "C" fn database_schema(database: *const Database) -> *mut c_char {
-    try_release_string(take_arc(database).schema())
+    try_release_string(borrow(database).schema())
```
farost, line 52: Lol!

flyingsilverfin, line 52: when do we use borrow vs take_arc?

flyingsilverfin (approved):

c/src/database.rs
farost: `take_arc` takes the ownership. We use it when we release the resource, e.g. on deletion -> `(database: Database)`
`borrow` is for borrowing resources -> `(database: &Database)`

farost:
```
pub(super) fn take_arc<T>(raw: *const T) -> Arc<T> {
    trace!("Taking ownership of arced <{}> @ {:?}", std::any::type_name::<T>(), raw);
    unsafe { Arc::from_raw(raw) }
}
```

```
pub(super) fn borrow<T>(raw: *const T) -> &'static T {
    trace!("Borrowing <{}> @ {:?}", std::any::type_name::<T>(), raw);
    assert!(!raw.is_null());
    unsafe { &*raw }
}
```

PR #758, opened 2025-05-28
Introduce file-based database export and import

farost:
## Usage and product changes
Introduce interfaces to export databases into schema definition and data files and to import databases using these files. Database import supports files exported from both TypeDB 2.x and TypeDB 3.x.

Both operations are blocking and may take a significant amount of time to execute for large databases. Use parallel connections to continue operating with the server and its other databases.

Usage examples in Rust:
```rust
// export
let db = driver.databases().get(db_name).await.unwrap();
db.export_to_file(schema_file_path, data_file_path).await.unwrap();

// import
let schema = read_to_string(schema_file_path).unwrap();
driver.databases().import_from_file(db_name2, schema, data_file_path).await.unwrap();
```

Usage examples in Python:
```py
# export
database = driver.databases.get(db_name)
database.export_to_file(schema_file_path, data_file_path)

# import
with open(schema_file_path, 'r', encoding='utf-8') as f:
    schema = f.read()
driver.databases.import_from_file(db_name2, schema, data_file_path)
```

Usage examples in Java:
```java
// export
Database database = driver.databases().get(dbName);
database.exportToFile(schemaFilePath, dataFilePath);

// import
String schema = Files.readString(Path.of(schemaFilePath));
driver.databases().importFromFile(dbName2, schema, dataFilePath);
```

## Implementation
Implemented the updated [protocol](https://github.com/typedb/typedb-protocol/pull/224).

As both operations work with streaming, the implementation is similar to transactions. The behavior is split into the file processing logic and networking (specialized for sync and async modes). The exposed interfaces present only the file-based versions, but additional interfaces for direct work with streams can be presented in future updates.

In Rust, paths are accepted as Rust `Path`s. In other languages working through C interfaces, it's pure strings for C layer transmission.

### Database export 
Implemented through the `database` interface and accepts two target files for export. Does not require a specific format of naming (so it's necessarily `.typeql` or `.typedb`. If any of the target files already exist, an error is returned.

The export operation consists of these steps:
* prepare the output files
* open a unidirectional GRPC stream from the server to the client
* "block" on server response listening until an error or a "done" is received (blocking is implemented through a loop which resolves a `listen` promise presented by the network layer)
* if there is a schema message, write it to the schema file and flush it right away
* if there is a data items message, encode the items and write them to the data file 
* in case of an error, the output files are deleted (we own them as we create them at the beginning)

The network layer is basically just a task listening for the GRPC stream and transmitting the converted messages to the processing loop.

### Database import
Implemented through the `database_manager` interface and accepts a database name, a schema definition query string (can be read from the exported file), and an exported data file. No naming requirements as well. 

The import operation consists of these steps:
* open the input file
* open a bidirectional GRPC stream between the server and the client, send the initial request with the database's name and schema
* eagerly start reading and decoding data items from the input file one by one, storing up to 250 items in the buffer (this number can be easily changed)
* once the buffer is full, attempt items sending operation: this operation will check a potential early error signal from the server and then send the batch, returning to processing the rest of the file
* once the file is read, send a "done" message and block until the server responds with either an error or its "done" message

The network layer consists of a blocking task for client-side requests and a listening task waiting for a one-shot signal from the server (either an error or a "done" message). Errors can be received at any time of processing, while "done" is expected only after a client-side "done" request. When a response is received, either an async or a sync sink receives this message, which should be checked before any client-side network operations to ensure proper interruption.

### DIFF
--- a/rust/src/connection/database/export_stream.rs
+++ b/rust/src/connection/database/export_stream.rs
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+use std::fmt;
+
+#[cfg(not(feature = "sync"))]
+use futures::{stream, StreamExt};
+
+use crate::{
+    common::{stream::Stream, Promise, Result},
+    connection::{message::DatabaseExportResponse, network::transmitter::DatabaseExportTransmitter},
+    database::migration::DatabaseExportAnswer,
+    error::ConnectionError,
+    promisify, Error,
+};
+
+pub(crate) struct DatabaseExportStream {
+    export_transmitter: DatabaseExportTransmitter,
+}
+
+impl DatabaseExportStream {
+    pub(crate) fn new(export_transmitter: DatabaseExportTransmitter) -> Self {
+        Self { export_transmitter }
+    }
+
+    pub(crate) fn listen<'a>(&'a mut self) -> impl Promise<'a, Result<DatabaseExportAnswer>> + 'a {
+        let mut stream = self.export_transmitter.stream();
+        promisify! {
+            #[cfg(feature = "sync")]
+            let response = stream.next();
+            #[cfg(not(feature = "sync"))]
+            let response: Option<Result<DatabaseExportResponse>> = stream.next().await;
+
+            let response = match response {
+                None => return Err(ConnectionError::QueryStreamNoResponse.into()),
+                Some(Err(err)) => return Err(err),
+                Some(Ok(response)) => response,
+            };
+            match response {
+                DatabaseExportResponse::Schema(schema) => Ok(DatabaseExportAnswer::Schema(schema)),
+                DatabaseExportResponse::Items(items) => Ok(DatabaseExportAnswer::Items(items)),
+                DatabaseExportResponse::Done => Ok(DatabaseExportAnswer::Done),
+                DatabaseExportResponse::Error(error) => Err(Error::Server(error)),
+            }
+        }
+    }
+}
+
+impl fmt::Debug for DatabaseExportStream {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("DatabaseExportStream").finish()
+    }
+}
--- a/rust/src/connection/database/export_stream.rs
+++ b/rust/src/connection/database/export_stream.rs
@@ -39,7 +39,7 @@ impl DatabaseExportStream {
         Self { export_transmitter }
     }
 
-    pub(crate) fn listen<'a>(&'a mut self) -> impl Promise<'a, Result<DatabaseExportAnswer>> + 'a {
+    pub(crate) fn next<'a>(&'a mut self) -> impl Promise<'a, Result<DatabaseExportAnswer>> + 'a {
         let mut stream = self.export_transmitter.stream();
         promisify! {
             #[cfg(feature = "sync")]
--- a/rust/src/connection/database/export_stream.rs
+++ b/rust/src/connection/database/export_stream.rs
@@ -47,16 +47,14 @@ impl DatabaseExportStream {
             #[cfg(not(feature = "sync"))]
             let response: Option<Result<DatabaseExportResponse>> = stream.next().await;
 
-            let response = match response {
-                None => return Err(ConnectionError::QueryStreamNoResponse.into()),
-                Some(Err(err)) => return Err(err),
-                Some(Ok(response)) => response,
-            };
             match response {
-                DatabaseExportResponse::Schema(schema) => Ok(DatabaseExportAnswer::Schema(schema)),
-                DatabaseExportResponse::Items(items) => Ok(DatabaseExportAnswer::Items(items)),
-                DatabaseExportResponse::Done => Ok(DatabaseExportAnswer::Done),
-                DatabaseExportResponse::Error(error) => Err(Error::Server(error)),
+                None => return Err(ConnectionError::DatabaseExportStreamNoResponse.into()),
+                Some(Err(err)) => return Err(err),
+                Some(Ok(response)) => match response {
+                    DatabaseExportResponse::Schema(schema) => Ok(DatabaseExportAnswer::Schema(schema)),
+                    DatabaseExportResponse::Items(items) => Ok(DatabaseExportAnswer::Items(items)),
+                    DatabaseExportResponse::Done => Ok(DatabaseExportAnswer::Done),
+                }
             }
         }
     }
--- a/rust/src/connection/database/export_stream.rs
+++ b/rust/src/connection/database/export_stream.rs
@@ -20,14 +20,14 @@
 use std::fmt;
 
 #[cfg(not(feature = "sync"))]
-use futures::{stream, StreamExt};
+use futures::StreamExt;
 
 use crate::{
     common::{stream::Stream, Promise, Result},
     connection::{message::DatabaseExportResponse, network::transmitter::DatabaseExportTransmitter},
     database::migration::DatabaseExportAnswer,
     error::ConnectionError,
-    promisify, Error,
+    promisify,
 };
 
 pub(crate) struct DatabaseExportStream {
--- a/rust/src/connection/database/import_stream.rs
+++ b/rust/src/connection/database/import_stream.rs
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+use std::{fmt, iter, pin::Pin, sync::Arc};
+
+#[cfg(not(feature = "sync"))]
+use futures::{stream, StreamExt};
+use typedb_protocol::migration;
+
+use crate::{
+    common::{stream::Stream, Promise, Result},
+    connection::{
+        message::{DatabaseImportRequest, TransactionResponse},
+        network::transmitter::DatabaseImportTransmitter,
+    },
+    promisify, resolve,
+};
+
+pub(crate) struct DatabaseImportStream {
+    import_transmitter: DatabaseImportTransmitter,
+}
+
+impl DatabaseImportStream {
+    pub(crate) fn new(import_transmitter: DatabaseImportTransmitter) -> Self {
+        Self { import_transmitter }
+    }
+
+    pub(crate) fn items(&self, items: Vec<migration::Item>) -> impl Promise<'static, Result> {
+        let promise = self.single(DatabaseImportRequest::ItemPart { items });
+        promisify! { resolve!(promise) }
+    }
+
+    pub(crate) fn done(self) -> impl Promise<'static, Result> {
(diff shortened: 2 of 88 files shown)


rust/src/connection/database/import_stream.rs
```diff
@@ -0,0 +1,58 @@
...
+pub(crate) struct DatabaseImportStream {
+    import_transmitter: DatabaseImportTransmitter,
+}
+
+impl DatabaseImportStream {
+    pub(crate) fn new(import_transmitter: DatabaseImportTransmitter) -> Self {
+        Self { import_transmitter }
+    }
+
+    pub(crate) fn items(&mut self, items: Vec<migration::Item>) -> Result {
```
farost, line 41: Send a bucket of migration items.

rust/src/connection/network/proto/message.rs
```diff
@@ -93,6 +96,34 @@ impl TryIntoProto<database_manager::create::Req> for Request {
...
+            other => Err(InternalError::UnexpectedRequestType { request_type: format!("{other:?}") }.into()),
+        }
+    }
+}
+
+impl IntoProto<migration::import::Client> for DatabaseImportRequest {
+    fn into_proto(self) -> migration::import::Client {
+        let req = match self {
+            DatabaseImportRequest::Initial { name, schema } => {
+                migration::import::client::Client::InitialReq(migration::import::client::InitialReq { name, schema })
```
farost, line 114: This double wrapping is something we have to deal with because of the protocol (see here: https://github.com/typedb/typedb-protocol/pull/224/files#r2108713254)

rust/src/connection/network/transmitter/export.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub(in crate::connection) fn stream(&mut self) -> &mut impl Stream<Item = Result<DatabaseExportResponse>> {
+        &mut self.stream
+    }
+
+    async fn start_workers(
+        response_source: Streaming<database::export::Server>,
+        response_sender: UnboundedSender<Result<DatabaseExportResponse>>,
+        shutdown_signal: UnboundedReceiver<()>,
+    ) {
+        tokio::spawn(Self::listen_loop(response_source, response_sender, shutdown_signal));
```
farost, line 67: `ExportTransmitter` just listens for server responses. I decided that it's better to make it similar to `TransactionTransmitter`. I don't see a need for a trait, though.

rust/src/connection/network/transmitter/import.rs
```diff
@@ -0,0 +1,173 @@
...
+impl DatabaseImportTransmitter {
+    pub(in crate::connection) fn new(
+        background_runtime: Arc<BackgroundRuntime>,
+        request_sink: UnboundedSender<database_manager::import::Client>,
+        response_source: Streaming<database_manager::import::Server>,
+    ) -> Self {
+        let (buffer_sink, buffer_source) = unbounded_async();
+        let (shutdown_sink, shutdown_source) = unbounded_async();
+
+        let (result_sink, result_source) = oneshot();
```
farost, line 59: We send migration items controlled by the client, but we also listen for a single result from the server. The server can either say "ERROR" once or "DONE" once ("DONE" is expected only on "DONE" from the client side, otherwise it's a bug).

We use `ResponseSink`, either an async or a sync one, to check the answer. The design is very similar to what we have in `transmitter/transaction.rs`. If you have any questions regarding the idea, please check that file first.

rust/src/connection/network/transmitter/import.rs
```diff
@@ -0,0 +1,173 @@
...
+        #[cfg(feature = "sync")]
+        let result_sink = ResponseSink::BlockingOneShot(result_sink);
+        #[cfg(not(feature = "sync"))]
+        let result_sink = ResponseSink::AsyncOneShot(result_sink);
```
farost, line 63: Could not come up with a better idea how to express this.

rust/src/connection/server_connection.rs
```diff
@@ -224,7 +265,7 @@ impl ServerConnection {
                     TransactionTransmitter::new(self.background_runtime.clone(), request_sink, response_source);
                 let transmitter_shutdown_sink = transmitter.shutdown_sink().clone();
                 let transaction_stream = TransactionStream::new(transaction_type, options, transmitter);
-                self.transaction_shutdown_senders.lock().unwrap().insert(request_id, transmitter_shutdown_sink);
```
farost, line 227: We don't use request_ids here, we only need it to broadcast the shutdown. And I don't want to add "import"/"export" pseudo-ids.

rust/src/database/database.rs
```diff
@@ -157,6 +162,40 @@ impl Database {
...
+    #[cfg_attr(not(feature = "sync"), doc = "database.export_file(schema_path, data_path).await;")]
+    /// ```
+    #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
+    pub async fn export_file(&self, schema_file_path: impl AsRef<Path>, data_file_path: impl AsRef<Path>) -> Result {
+        let schema_file_path = schema_file_path.as_ref();
+        let data_file_path = data_file_path.as_ref();
+
+        let result = self
+            .run_failsafe(|database| async move {
+                let schema_file = try_creating_export_file(schema_file_path)?;
```
farost, line 186: It errors if the file already exists. So deleting it below is fine

rust/src/database/database_manager.rs
```diff
@@ -156,6 +162,60 @@ impl DatabaseManager {
...
+        data_file_path: impl AsRef<Path>,
+    ) -> Result {
+        const ITEM_BATCH_SIZE: usize = 250;
+
+        let name = name.into();
+        let schema: String = schema.into();
+        let schema_ref: &str = schema.as_ref();
+        let data_file_path = data_file_path.as_ref();
+        self.run_failsafe(name, |server_connection, name| async move {
+            let mut import_stream = server_connection.import_database(name, schema_ref.to_string()).await?;
```
farost, line 196: We don't wait for a server response here and proceed to read the file and send its items. However, if an error is returned, we will stop on one of the `.items(...)` calls.

farost:
CI is red because there is no server artifact yet.

rust/src/connection/network/transmitter/shutdown_guard.rs
```diff
@@ -0,0 +1,47 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+use std::ops::Deref;
+
+use tokio::sync::mpsc::UnboundedSender;
+
+// Makes sure that the shutdown signal is sent when it's dropped. Can try sending redundant shutdown
+// signals after another shutdown source: the channel is expected to be closed (no-op).
+pub(crate) struct ShutdownGuard<T: Default> {
```
farost, line 25: Previously, we always counted on the server to drop the stream. However, for import, we are the owners of the stream, and we should make sure to terminate it for both 
1. GRPC connection dropping because of the server or the network
2. Our error leading to resource deallocation <- this struct helps specifically with this case

rust/src/connection/message.rs
```diff
@@ -94,9 +100,13 @@ pub(super) enum Response {
...
         schema: String,
     },
-    DatabaseRuleSchema {
-        schema: String,
+    DatabaseExportStream {
+        response_source: Streaming<database::export::Server>,
+    },
+    DatabaseExportAnswer {
+        response_source: Streaming<database::export::Server>,
     },
```
dmitrii-ubskii, line 108: Is `Response::DatabaseExportAnswer` used? It seems to be a doublet of `DatabaseExportStream`.

rust/src/connection/network/stub.rs
```diff
@@ -132,6 +148,17 @@ impl<Channel: GRPCChannel> RPCStub<Channel> {
         self.single(|this| Box::pin(this.grpc.database_type_schema(req.clone()))).await
     }
 
+    pub(super) async fn database_export(
+        &mut self,
+        req: database::export::Req,
+    ) -> Result<Streaming<database::export::Server>> {
+        self.call_with_auto_renew_token(|this| {
+            let req = req.clone();
```
dmitrii-ubskii, line 156: Is this necessary? Other requests get away with inlining `req.clone()` directly in the closure.

rust/src/connection/network/transmitter/mod.rs
```diff
@@ -17,12 +17,20 @@
...
-use crossbeam::channel::{bounded as bounded_blocking, Receiver as SyncReceiver, Sender as SyncSender};
+use crossbeam::channel::{
+    bounded as bounded_blocking, Receiver as SyncReceiver, Sender as SyncSender, TryRecvError as SyncTryRecvError,
+};
 
-pub(in crate::connection) use self::{rpc::RPCTransmitter, transaction::TransactionTransmitter};
+pub(in crate::connection) use self::{
+    export::DatabaseExportTransmitter, import::DatabaseImportTransmitter, rpc::RPCTransmitter,
+    transaction::TransactionTransmitter,
+};
```
dmitrii-ubskii, line 27: New imports, apparently unused.

rust/src/connection/network/transmitter/import.rs
```diff
@@ -0,0 +1,184 @@
...
+            }
+            sleep(DISPATCH_INTERVAL);
+            if let Ok(request) = request_source.try_recv() {
+                let client_req = database_manager::import::Client { client: Some(request.into_proto()) };
+                request_sink.send(client_req).ok();
+            }
+        }
+    }
+
+    async fn listen(
```
dmitrii-ubskii, line 171: Same as `listen` above, this should be renamed

rust/src/database/database.rs
```diff
@@ -395,6 +434,33 @@ impl ServerDatabase {
...
+                        let mut buf = Vec::new();
+                        item.encode_length_delimited(&mut buf)
+                            .map_err(|_| Error::Migration(MigrationError::CannotEncodeExportedConcept))?;
+                        data_writer.write_all(&buf)?;
+                    }
+                }
+            }
+        }
+
+        data_writer.flush()?;
```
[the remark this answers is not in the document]
dmitrii-ubskii: Is there a benefit? If the driver crashes you'd have to restart the export process anyway.

rust/src/database/database_manager.rs
```diff
@@ -156,6 +162,61 @@ impl DatabaseManager {
...
+    /// * `data_file_path` — The exported database file to import the data from
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    #[cfg_attr(feature = "sync", doc = "driver.databases().import_file(name, schema, data_path);")]
+    #[cfg_attr(not(feature = "sync"), doc = "driver.databases().import_file(name, schema, data_path).await;")]
+    /// ```
+    #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
+    pub async fn import_file(
```
dmitrii-ubskii, line 183: `import_database_from_file`

rust/src/database/database_manager.rs
```diff
@@ -156,6 +162,60 @@ impl DatabaseManager {
...
+
+            let file = try_opening_import_file(data_file_path)?;
+            let mut item_buffer = Vec::with_capacity(ITEM_BATCH_SIZE);
+            let mut read_item_iterator = ProtoMessageIterator::<Item, _>::new(BufReader::new(file));
+
+            while let Some(item) = read_item_iterator.next() {
+                let item = item?;
+                item_buffer.push(item);
+                if item_buffer.len() >= ITEM_BATCH_SIZE {
+                    import_stream.items(item_buffer.split_off(0))?;
```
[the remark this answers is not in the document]
dmitrii-ubskii: Another option is `mem::replace(&mut item_buffer, Vec::with_capacity(ITEM_BATCH_SIZE))`. 
Neither is particularly readable, but `split_off` is at least shorter.

rust/src/database/migration.rs
```diff
@@ -0,0 +1,143 @@
...
+                _ => return Some(Err(Error::Migration(MigrationError::CannotDecodeImportedConcept))),
+            }
+        }
+
+        let mut message_buf = self.get_message_buf(message_len);
+        Some(M::decode(&mut message_buf).map_err(|_| Error::Migration(MigrationError::CannotDecodeImportedConcept)))
+    }
+}
+
+pub(crate) fn try_creating_export_file(path: impl AsRef<Path>) -> Result<File> {
```
dmitrii-ubskii, line 119: `try_create_...` and so on, `ing` is against convention.

dmitrii-ubskii (approved):
Naming and style comments aside, looks good.

rust/src/connection/database/export_stream.rs
```diff
@@ -0,0 +1,67 @@
...
+pub(crate) struct DatabaseExportStream {
+    export_transmitter: DatabaseExportTransmitter,
+}
+
+impl DatabaseExportStream {
+    pub(crate) fn new(export_transmitter: DatabaseExportTransmitter) -> Self {
+        Self { export_transmitter }
+    }
+
+    pub(crate) fn listen<'a>(&'a mut self) -> impl Promise<'a, Result<DatabaseExportAnswer>> + 'a {
```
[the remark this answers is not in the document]
farost: `next` is good, thansk

rust/src/database/database.rs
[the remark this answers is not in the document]
farost: True. Yeah, I think it knows how to optimally write itself. I was more concerned about buffers for a moment.

rust/src/connection/message.rs
farost: Yep, it got redundant after a couple of protocol refactoring iterations. Good catch

rust/src/connection/network/transmitter/mod.rs
farost: They are used!

rust/src/database/database_manager.rs
farost: `_database_ is clearly redundant, we don't call `create` `create_database` instead. `from` is fine, I will use it, alright.

PR #777, opened 2025-07-25
Enable TLS by default in Python

flyingsilverfin:
## Usage and product changes

We want to enable a secure-by-default setting in TypeDB Drivers. In Java and Rust, `DriverOptions` have to be explicitly set, and there are no defaults. However, Python features a disabled TLS default. While this is compatible with TypeDB CE, it's an insecure default & not compatible with TypeDB Cloud without explicitly enabling it.

Instead, we set the default to TLS being __enabled__ in Python. This means when using an insecure, plaintext connection the user must explicitly set it, and is more likely to become aware of the plaintext communication.

## Implementation

- Fix a documentation string in Rust
- Set the default TLS connection in Python to be TLS enabled, instead of disabled

- Update some forgotten HTTP-TS docs generation and testing targets

### DIFF
--- a/http-ts/BUILD
+++ b/http-ts/BUILD
@@ -110,6 +110,7 @@ checkstyle_test(
         "**/*.md",
         "pnpm-lock.yaml",
         "tsup.config.ts",
+        ".npmrc",
         "LICENSE",
         "docs/**/*",
         "dist/**/*",


farost:
It's weird that python BDDs are green, because they create default DriverOptions. Maybe we have a bug somewhere! They should not be able to connect to an unprotected server (which is booted up by default)

I may also miss smth

```
def create_driver(context, host="localhost", port=None, username=None, password=None) -> Driver:
    if port is None:
        port = int(context.config.userdata["port"])
    if username is None:
        username = "admin"
    if password is None:
        password = "password"
    credentials = Credentials(username, password)
    return TypeDB.driver(address=f"{host}:{port}", credentials=credentials, driver_options=DriverOptions())
```

PR #784, opened 2025-09-03
Fix database importer decoding error

farost:
## Usage and product changes
Stabilize the database import function by eliminating rare decoding errors that could occur during the import of large datasets. All the exported files that the import function could not process are still valid and should be correctly imported after the proposed changes.

## Implementation
* For better control of the manual bytes processing, replace `VecDeque` with `prost::bytes::*` and isolate the file reader's moving in a single function (`Iterator::next`).
* Since `read` functions can read fewer bytes than requested even without meeting EOF, replace single `if` calls with `loop`s <- this was probably the main issue before.

I was not able to reduce the bug to a single test case, even with thousands of records (however, we had a report of a failure after processing 200 entries), so no new behavior tests were introduced. I decided not to spend more time searching for such cases.

### DIFF
--- a/rust/src/database/migration.rs
+++ b/rust/src/database/migration.rs
@@ -48,6 +48,7 @@ pub struct ProtoMessageIterator<M: Message + Default, R: BufRead> {
 
 impl<M: Message + Default, R: BufRead> ProtoMessageIterator<M, R> {
     const BUF_CAPACITY: usize = 8 * 1024;
+    // prost's length delimiters take up to 10 bytes
     const MAX_LENGTH_DELIMITER_LEN: usize = 10;
 
     pub fn new(reader: R) -> Self {


rust/src/database/migration.rs
```diff
@@ -40,35 +42,41 @@ pub(crate) enum DatabaseExportAnswer {
...
     const MAX_LENGTH_DELIMITER_LEN: usize = 10;
 
     pub fn new(reader: R) -> Self {
-        Self { reader, buffer: VecDeque::with_capacity(Self::BUF_CAPACITY), _phantom_data: PhantomData }
+        Self { reader, buffer: BytesMut::with_capacity(Self::BUF_CAPACITY), _phantom_data: PhantomData }
     }
 
-    fn try_read_more(&mut self, bytes_to_read: usize) -> std::io::Result<usize> {
-        let mut addition = vec![0; bytes_to_read];
+    fn read_more(&mut self, bytes_to_read: usize) -> std::io::Result<usize> {
```
farost, line 57: No "not-try" methods, so it's probably not a `try`

rust/src/database/migration.rs
```diff
@@ -40,35 +42,41 @@ pub(crate) enum DatabaseExportAnswer {
...
+        }
+        let mut addition = vec![0u8; max(bytes_to_read, 1)];
         let bytes_read = self.reader.read(&mut addition)?;
-        self.buffer.extend(&addition[..bytes_read]);
+        self.buffer.extend_from_slice(&addition[..bytes_read]);
         Ok(bytes_read)
     }
 
-    fn try_get_next_message_len(&mut self) -> Result<Option<usize>> {
+    fn decode_next_len(&mut self) -> Result<Option<(usize /*len*/, usize /*consumed*/)>> {
```
farost, line 67: Return `consumed` to move the buffer/cursor/reader only in a single function when needed

rust/src/database/migration.rs
```diff
@@ -40,35 +42,41 @@ pub(crate) enum DatabaseExportAnswer {
...
-                return Ok(Some(len));
-            } else {
-                if self.buffer.len() < Self::MAX_LENGTH_DELIMITER_LEN {
-                    assert!(Self::MAX_LENGTH_DELIMITER_LEN < Self::BUF_CAPACITY);
-                    let to_read = max(Self::MAX_LENGTH_DELIMITER_LEN, Self::BUF_CAPACITY - self.buffer.len());
-                    match self.try_read_more(to_read) {
+            let mut cursor: &[u8] = &self.buffer;
+            match prost::decode_length_delimiter(&mut cursor) {
+                Ok(len) => {
+                    let consumed = self.buffer.len() - cursor.len();
```
farost, line 72: `cursor` points to `buffer`

rust/src/database/migration.rs
```diff
@@ -79,40 +87,34 @@ impl<M: Message + Default, R: BufRead> ProtoMessageIterator<M, R> {
...
         };
 
-        if self.buffer.len() < message_len {
-            let required = message_len - self.buffer.len();
-            let to_read = max(required, Self::BUF_CAPACITY);
-            match self.try_read_more(to_read) {
-                Ok(bytes_read) if bytes_read >= required => {}
-                _ => return Some(Err(Error::Migration(MigrationError::CannotDecodeImportedConcept))),
+        let required = consumed + message_len;
+        while self.buffer.len() < required {
```
farost, line 107: More read retries

rust/src/database/migration.rs
```diff
@@ -79,40 +87,34 @@ impl<M: Message + Default, R: BufRead> ProtoMessageIterator<M, R> {
...
+            let to_read = required - self.buffer.len();
+            match self.read_more(max(to_read, Self::BUF_CAPACITY)) {
+                Ok(0) | Err(_) => return Some(Err(Error::Migration(MigrationError::CannotDecodeImportedConcept))),
+                Ok(_) => {}
             }
         }
 
-        let mut message_buf = self.get_message_buf(message_len);
-        Some(M::decode(&mut message_buf).map_err(|_| Error::Migration(MigrationError::CannotDecodeImportedConcept)))
+        self.buffer.advance(consumed);
```
farost, line 115: Advance in a single place

flyingsilverfin (approved):
To be honest I kinda skimmed it but if it works i'm happy!

PR #792, opened 2025-10-02
Improve on close callback

flyingsilverfin:

### DIFF
--- a/c/src/transaction.rs
+++ b/c/src/transaction.rs
@@ -66,8 +66,8 @@ pub extern "C" fn transaction_close(txn: *mut Transaction) {
 
 /// Forcibly closes this transaction. To be used in exceptional cases.
 #[no_mangle]
-pub extern "C" fn transaction_force_close(txn: *mut Transaction) {
-    borrow_mut(txn).force_close();
+pub extern "C" fn transaction_force_close(txn: *mut Transaction) -> *mut VoidPromise {
+    release(VoidPromise(Box::new(borrow_mut(txn).force_close())))
 }
 
 /// Commits the changes made via this transaction to the TypeDB database.
--- a/c/src/transaction.rs
+++ b/c/src/transaction.rs
@@ -20,7 +20,7 @@
 use std::{ffi::c_char, ptr::null_mut};
 
 use typedb_driver::{
-    DatabaseManager, Error, QueryOptions, Transaction, TransactionOptions, TransactionType, TypeDBDriver,
+    Error, QueryOptions, Transaction, TransactionOptions, TransactionType, TypeDBDriver,
 };
 
 use super::memory::{borrow, borrow_mut, free, release, take_ownership};
@@ -60,14 +60,14 @@ pub extern "C" fn transaction_query(
 
 /// Closes the transaction and frees the native rust object.
 #[no_mangle]
-pub extern "C" fn transaction_close(txn: *mut Transaction) {
+pub extern "C" fn transaction_submit_close(txn: *mut Transaction) {
     free(txn);
 }
 
 /// Forcibly closes this transaction. To be used in exceptional cases.
 #[no_mangle]
-pub extern "C" fn transaction_force_close(txn: *mut Transaction) -> *mut VoidPromise {
-    release(VoidPromise(Box::new(borrow_mut(txn).force_close())))
+pub extern "C" fn transaction_close(txn: *mut Transaction) -> *mut VoidPromise {
+    release(VoidPromise(Box::new(borrow_mut(txn).close())))
 }
 
 /// Commits the changes made via this transaction to the TypeDB database.
@@ -100,7 +100,9 @@ pub extern "C" fn transaction_on_close(
     txn: *const Transaction,
     callback_id: usize,
     callback: extern "C" fn(usize, *mut Error),
-) {
-    borrow(txn)
-        .on_close(move |error| callback(callback_id, error.map(|err| release(err.into())).unwrap_or(null_mut())));
+) -> *mut VoidPromise {
+    release(VoidPromise(Box::new(
+        borrow(txn)
+        .on_close(move |error| callback(callback_id, error.map(|err| release(err.into())).unwrap_or(null_mut())))
+    )))
 }
--- a/c/src/transaction.rs
+++ b/c/src/transaction.rs
@@ -19,9 +19,7 @@
 
 use std::{ffi::c_char, ptr::null_mut};
 
-use typedb_driver::{
-    Error, QueryOptions, Transaction, TransactionOptions, TransactionType, TypeDBDriver,
-};
+use typedb_driver::{Error, QueryOptions, Transaction, TransactionOptions, TransactionType, TypeDBDriver};
 
 use super::memory::{borrow, borrow_mut, free, release, take_ownership};
 use crate::{answer::QueryAnswerPromise, error::try_release, memory::string_view, promise::VoidPromise};
@@ -103,6 +101,6 @@ pub extern "C" fn transaction_on_close(
 ) -> *mut VoidPromise {
     release(VoidPromise(Box::new(
         borrow(txn)
-        .on_close(move |error| callback(callback_id, error.map(|err| release(err.into())).unwrap_or(null_mut())))
+            .on_close(move |error| callback(callback_id, error.map(|err| release(err.into())).unwrap_or(null_mut()))),
     )))
 }
--- a/c/typedb_driver.i
+++ b/c/typedb_driver.i
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+%module(threads=1) typedb_driver
 %module(directors="1") typedb_driver
 %{
 extern "C" {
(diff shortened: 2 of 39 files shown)


c/src/transaction.rs
```diff
@@ -60,14 +58,14 @@ pub extern "C" fn transaction_query(
 
 /// Closes the transaction and frees the native rust object.
 #[no_mangle]
-pub extern "C" fn transaction_close(txn: *mut Transaction) {
+pub extern "C" fn transaction_submit_close(txn: *mut Transaction) {
```
flyingsilverfin, line 61: New paradigm i chose:
`close()` returns a promise and does not take ownership.
We also have a `submit_close()`, which a fire-and-forget equivalent that is used by `drop` since we can't block or resolve promises there.

Note: we still use `force_close` on driver for server connections. Didn't change it there!

c/typedb_driver.i
```diff
@@ -116,12 +117,75 @@ struct TransactionCallbackDirector {
 %{
 #include <memory>
 #include <iostream>
+#include <mutex>
 #include <unordered_map>
-static std::unordered_map<size_t, TransactionCallbackDirector*> transactionOnCloseCallbacks {};
```
flyingsilverfin, line 120: This had a race condition if the driver calls a callback at the same time as a user adding one to a new transaction.

c/typedb_driver.i
```diff
@@ -116,12 +117,75 @@ struct TransactionCallbackDirector {
...
+    // 2. The static mutex to manage access
+    static std::mutex s_mutex;
+
+public:
+    // Delete copy/move constructors and assignment operators
+    // to prevent accidental copying of the singleton-like structure
+    ThreadSafeTransactionCallbacks(const ThreadSafeTransactionCallbacks&) = delete;
+    ThreadSafeTransactionCallbacks& operator=(const ThreadSafeTransactionCallbacks&) = delete;
+
+    // --- Core Operations ---
```
flyingsilverfin, line 137: yes this was AI generated

c/typedb_driver.i
```diff
@@ -132,11 +196,11 @@ static void transaction_callback_execute(size_t ID, Error* error) {
 %ignore transaction_on_close;
 %inline %{
 #include <atomic>
-void transaction_on_close_register(const Transaction* transaction, TransactionCallbackDirector* handler) {
+VoidPromise* transaction_on_close_register(const Transaction* transaction, TransactionCallbackDirector* handler) {
```
flyingsilverfin, line 199: `on_close` callback registration also returns a promise, since we're not guaranteed that the on_close callback is registered when this returns.

python/typedb/connection/transaction.py
```diff
@@ -78,7 +78,8 @@ def is_open(self) -> bool:
         return transaction_is_open(self.native_object)
 
     def on_close(self, function: callable):
-        transaction_on_close(self.native_object, _Transaction.TransactionOnClose(function).__disown__())
+        callback = _Transaction.TransactionOnClose(function)
+        void_promise_resolve(transaction_on_close(self.native_object, callback.__disown__()))
```
flyingsilverfin, line 82: Python also actively resolves on_close and close() for the user

rust/src/connection/network/transmitter/transaction.rs
```diff
@@ -118,15 +117,77 @@ impl TransactionTransmitter {
...
     }
 
-    pub(in crate::connection) fn force_close(&self) {
+    #[cfg(not(feature = "sync"))]
+    pub(in crate::connection) fn close(&self) -> impl Promise<'_, Result<()>> {
+        box_promise(async move {
+            if self.background_runtime.is_open() && self.is_open.compare_exchange(true, false).is_ok() {
+                let (closed_sink, mut closed_source) = unbounded_async();
+                let close_notifier_callback = Box::new(move |error| {
+                    closed_sink.send(()).unwrap();
```
flyingsilverfin, line 126: This is kinda a hack but honestly might be ok

rust/src/connection/network/transmitter/transaction.rs
```diff
@@ -118,15 +117,77 @@ impl TransactionTransmitter {
...
         }
     }
 
-    pub(in crate::connection) fn on_close(&self, callback: impl FnOnce(Option<Error>) + Send + Sync + 'static) {
-        self.on_close_register_sink.send(Box::new(callback)).ok();
+    #[cfg(not(feature = "sync"))]
+    pub(in crate::connection) fn on_close(
+        &self,
+        callback: impl FnOnce(Option<Error>) + Send + Sync + 'static,
+    ) -> impl Promise<'_, Result<()>> {
```
flyingsilverfin, line 171: on_close is also implemented as a promise now

rust/src/connection/network/transmitter/transaction.rs
```diff
@@ -242,66 +306,77 @@ impl TransactionTransmitter {
             on_close: Default::default(),
             callback_handler_sink,
         };
+        trace!("... Rust spawning sync dispatch loop");
         tokio::task::spawn_blocking({
             let collector = collector.clone();
-            move || {
-                Self::dispatch_loop(queue_source, request_sink, collector, on_close_callback_source, shutdown_signal)
-            }
+            move || Self::sync_dispatch_loop(queue_source, request_sink, collector, shutdown_signal)
```
flyingsilverfin, line 312: Since we just have to get the on_close_callback into the Collector in some way, we can choose whether we lsiten to the channel in the async listen loop or the sync dispatch loop. To minimize waits, we're going to let Tokio manage it in the Async loop.

rust/src/connection/network/channel.rs
```diff
@@ -58,6 +60,7 @@ pub(super) fn open_callcred_channel(
             driver_options.tls_config().clone().expect("TLS config object must be set when TLS is enabled");
         builder = builder.tls_config(tls_config)?;
     }
+    builder = builder.http2_keep_alive_interval(Duration::from_secs(10)).keep_alive_while_idle(true);
```
flyingsilverfin, line 63: so this is a bit of mystery:
without it (this was AI suggested), it appears that server responses (in particular, the transaction open response) never gets delivered into our code.

Ie. we see:
Driver: send open transaction request
Server: receive open txn request OpenTransaction.Req
Server: open txn, response with OpenTransaction.Res

(confirmed the above with wireshark).

The client side actually receives _something (from stub.rs):
```
this
                    .grpc
                    .transaction(UnboundedReceiverStream::new(receiver))
```

when awaited, this actually returns a stream successfully -- however, the `OpenTransaction.Res` message doesn't arrive until "something else" happens, such as the stream closing because of a transaction timeout...

It's very strange but since this solves it and it's sucked up a ton of time, I'm going to leave this in.

rust/src/connection/network/transmitter/rpc.rs
```diff
@@ -100,9 +101,15 @@ impl RPCTransmitter {
             request = request_source.recv() => request,
             _ = shutdown_signal.recv() => None,
         } {
+            trace!("RPC dispatcher loop received request {:?}", request);
```
flyingsilverfin, line 104: If we want to see everything happening, on TRACE we'll basically see every message in and out

rust/src/connection/transaction_stream.rs
```diff
@@ -90,10 +87,17 @@ impl TransactionStream {
...
 
-    pub(crate) fn on_close(&self, callback: impl FnOnce(Option<Error>) + Send + Sync + 'static) {
+    pub(crate) fn on_close(
+        &self,
+        callback: impl FnOnce(Option<Error>) + Send + Sync + 'static,
+    ) -> impl Promise<'_, Result<()>> {
         self.transaction_transmitter.on_close(callback)
     }
 
+    pub(crate) fn close(&self) -> impl Promise<'_, Result<()>> {
```
flyingsilverfin, line 97: So here's a question: why is it ok for close() to be a true promise, while the variants below for commit and rollback are resolved and re-emitted as promises? We should probably pick 1 approach

cc @farost  @dmitrii-ubskii

rust/tests/integration/driver.rs
```diff
@@ -0,0 +1,84 @@
...
+        transaction
+            .on_close(Box::new({
+                let clone = close_called.clone();
+                move |error| {
+                    clone.store(true, Ordering::SeqCst);
+                }
+            }))
+            .await;
+
+        transaction.close().await;
```
flyingsilverfin, line 74: verify: if we await the close() future, we have guaranteed all thea llocated callbacks are executed.

c/src/transaction.rs
flyingsilverfin: @dmitrii-ubskii  @farost important change!

python/typedb/connection/transaction.py
flyingsilverfin: @dmitrii-ubskii  @farost important to know in the new  settings

flyingsilverfin: (Java's the same)

rust/src/connection/transaction_stream.rs
flyingsilverfin: Yeah we do - but should we? I think close() definitely should be blocking unless the user wants it. Rollback I think probably makes the most sense to block - otherwise the transaction might be in a weird state when you keep using it? Although I think we preserve message ordering so it might be fine not to resolve?

Then there's commit - I think resolving the commit probably makes sense, since users could very easily async fire off huge amounts of say write transactions and there's no slowing down the client at all. Too easy to forget the commit().await/resolve

PR #793, opened 2025-10-03
Update dependencies after servers relation index fix for CI tests

farost:
## Usage and product changes
Ensure drivers run new BDD tests for migration written with the core's relation index fix https://github.com/typedb/typedb/pull/7594


krishnangovindraj (approved):
Trivial, lgtm

PR #799, opened 2025-10-22
Set up configurable FFI/Rust driver logging

flyingsilverfin:
## Usage and product changes

Rather than having Python/Java/other wrappers invoke logging initialization explicitly, we now just do it on open of the Rust driver automatically.

This should fix warnings in Python:
```
Failed to initialize logging: attempted to set a logger after the logging system was already initialized
```

We also remove uses of the `log` create in the `//c` layer, and use `tracing` instead.

We can now configure logging in the Rust-driver component with the following variables: `TYPEDB_DRIVER_LOG=info|debug|warn|trace` or using the usual `RUST_LOG` variable. By default, ffi-based drivers calling `init_logging` will use INFO logging.

Rust applications should continue set up their own logging subscriber, as usual.

## Implementation

Remove C-layer static logging initialization, and rely on just the Rust driver's once lock that is execute on driver open.

Drive-by: add claude.md

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -128,9 +128,14 @@ BDD tests use Gherkin specifications from the `typedb-behaviour` repository.
 
 Driver logging is controlled via environment variables:
 
-- `TYPEDB_DRIVER_LOG`: Log level for driver core (`debug`, `info`, `warn`, `error`, `trace`)
-- `TYPEDB_DRIVER_CLIB_LOG`: Log level for C FFI layer
-- `RUST_LOG`: Standard Rust logging (lower priority than above)
+- `TYPEDB_DRIVER_LOG`: Fine-grained control using the same syntax as `RUST_LOG`.
+  Example: `TYPEDB_DRIVER_LOG=typedb_driver=debug,typedb_driver_clib=trace`
+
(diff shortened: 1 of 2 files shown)


c/src/error.rs
```diff
@@ -35,12 +35,41 @@ thread_local! {
...
+        } else {
+            "info".to_owned()
+        };
+        // Try to get log level from TYPEDB_DRIVER_LOG first
+        let env_filter = if let Ok(typedb_log_level) = std::env::var("TYPEDB_DRIVER_LOG") {
+            EnvFilter::new(&format!("typedb_driver={},typedb_driver_clib={}", typedb_log_level, clib_level))
+        } else if let Ok(rust_log) = std::env::var("RUST_LOG") {
+            // If RUST_LOG is set, use it but scope it to typedb_driver only
+            EnvFilter::new(&format!("typedb_driver={},typedb_driver_clib={}", rust_log, clib_level))
+        } else {
```
dmitrii-ubskii: What happens if `RUST_LOG` is set to e.g. `tonic=trace`?

dmitrii-ubskii (changes requested):

PR #802, opened 2025-10-28
Use naiive date in Python

flyingsilverfin:
## Usage and product changes

Python `Date` objects were timezone-aware, which means that it was possible to insert, for example `2010-10-10` (recorded on the server-side in UTC :00-00-00), and read it back as `2010-10-09` when in a negative timezone relative to UTC!

We now parse the date received from the server naiively as a datetime, using UTC as the set point, and extract the naiive date from there.


krishnangovindraj (approved):
Makes sense to me. Can we add a reliable test for this or is it hard to trigger it in a test?

PR #805, opened 2025-11-02
Align HTTP TS analyze endpoints with GRPC drivers

krishnangovindraj:
## Usage and product changes
Updates the HTTP TypeScript driver to reflect the GRPC driver.
We also introduce an optional query structure into the GRPC response for pipelines without fetch. This can be requested using the new include_query_structure flag in QueryOptions.

## Implementation
**Reflects Protocol changes**
* Introduces AnalyzeOptions in the protocol <ins>**but does not expose it in rust or FFI**</ins>.
* Introduces AnalyzeOptions it in the HTTP-TS driver.
* Introduces query structure in GRPC query response
* Introduces include_query_structure in GRPC query options.
**Tests**
* Also implements analyze steps for HTTP TS driver BDD.


flyingsilverfin:
- align GRPC and HTTP analyze
-

http-ts/src/index.ts
```diff
@@ -107,6 +108,10 @@ export class TypeDBHttpDriver {
         return this.apiPost(`/v1/transactions/${encodeURIComponent(transactionId)}/rollback`, {});
     }
 
+    analyze(transactionId: string, query: string, queryOptions?: QueryOptions): Promise<ApiResponse<AnalyzeResponse>> {
```
flyingsilverfin: Shouldn't analyze take a different kind of query options?

java/api/answer/ConceptRow.java
```diff
@@ -55,6 +56,17 @@ public interface ConceptRow {
     @CheckReturnValue
     QueryType getQueryType();
 
+    /**
+     * Retrieves the executed query's structure of this <code>ConceptRow</code>.
+     * Shared between all the rows in a QueryAnswer.
```
flyingsilverfin: Maybe add a note to these docs throughout "When enabled as a query option, retrieves the executed query structure of this ConceptRow" or something like that

http-ts/src/index.ts
krishnangovindraj: Good catch.

PR #830, opened 2025-12-17
Introduce TypeDB cluster support

farost:
## Usage and product changes

Introduce cluster support across all TypeDB 3.x driver languages (Rust, Java, Python, C#, C/C++, HTTP-TypeScript).

### gRPC

**This release contains breaking changes to public interfaces and will require minor code updates as a part of the migration.**

**This release is not compatible with the old GRPC protocol. The C headers have changed significantly, requiring updates in the dependent drivers.**

- Implement cluster-aware connection management. The driver discovers all replicas on connection open and automatically routes operations to the primary replica for strong consistency. You can specify multiple replicas of a cluster or just a single address -- even one replica is enough to get the whole information about the cluster and establish a consistent and automatically redirecting connection.
- Change `DriverOptions`. Add `DriverTlsConfig` for explicit TLS configuration: `disabled()`, `enabled_with_native_root_ca()`, `enabled_with_root_ca(path)`. Replaces the previous `is_tls_enabled` boolean.
- Add `ServerRouting` for controlling request routing:
  - `Auto` — driver selects the primary replica automatically, with failover on leadership changes. Currently, always strongly consistent.
  - `Direct(address)` — route to a specific replica. Useful for read-only workloads on secondaries where it's possible.
- Add `Server` API exposing replica metadata: `id`, `role` (Primary/Candidate/Secondary), `term`, `address`, and `version`.
- Add address translation for cloud/NAT deployments. Supports mapping public addresses to private connection addresses.
- Add primary failover with configurable retries (`DriverOptions::primary_failover_retries`, default: 1). On leadership change, the driver seeks the new primary and retries the operation.
- Refactor `Database` API: remove unused `ReplicaInfo` / `DatabaseReplicas`. Replica information is now server-level via `driver.servers()`.
- Add `request_timeout` to `DriverOptions` to ensure any operation of the driver (excluding operations within already opened transactions: queries, commits, ...) is executed within the set time limit. The driver returns an error if the time limit is reached without a response from a server. This prevents the request from hanging indefinitely if the server is online, but does not respond. 
- Rename `community` back to `core` to highlight the technical side of the edition used.
- Complete the format and the content of the API references.

### HTTP-TS

**The HTTP-TS driver IS backwards compatible. Updating it will not introduce build or runtime errors, but will let your applications support automatic redirects when putting multiple addresses in the constructor.** 

- Implement cluster-aware connection management. Unlike gRPC, its retry logic is not configurable, but follows a similar pattern: the driver tries executing a request against the first address specified in the constructor; in case of an explicit redirect in the error response it reauthenticates & redirects the request to the specified location; in case the error is preserved the driver tries the rest of the addresses in the list; if no addresses work, the request fails. 
- Add new messages for retrieving `Server` statuses.

#### Server discovery and connection management

`ServerManager` orchestrates cluster topology. On connection open, the initial server responds with the full replica list (`ServerManager.All.Res`). The manager maintains:
- `replicas: RwLock<HashSet<AvailableServer>>` — discovered servers with replication status
- `replica_connections: RwLock<HashMap<Address, ServerConnection>>` — connection pool to each known replica
- `address_translation: RwLock<AddressTranslation>` — public-to-private address mapping for cloud deployments

Discovery happens at startup and is refreshed during failover via `refresh_replicas()`. New connections are created lazily, and stale connections (to removed replicas) are cleaned up.

#### Routing and failover

`ServerRouting::Auto` routes to the primary replica for strongly consistent operations. The primary is determined by the highest Raft term among replicas with `ReplicationRole::Primary`.

On failure, `execute_strongly_consistent()` distinguishes between:
- `ClusterReplicaNotPrimary` — leadership changed, seek new primary in existing replica set
- Network errors — the old primary is unreachable, seek primary excluding the failed address

Primary seeking (`seek_primary_replica_in`) connects to available replicas, calls `servers_all()` to get updated topology, and selects the primary with the highest term. Configurable via `DriverOptions::primary_failover_retries` (default: 1 retry).

#### Address translation

`Addresses` enum supports three modes:
- `Direct(Vec<Address>)` — single or multiple server addresses
- `Translated(HashMap<Address, Address>)` — public-to-private address mapping

Address translation enables cloud deployments where servers advertise public addresses but the driver connects via private addresses. The translation is applied bidirectionally: discovered public addresses are mapped to private addresses for connections, and private addresses are mapped back to public for display.

#### Server metadata

`AvailableServer` carries `ReplicationStatus` (id, role, term) from the Raft consensus layer. The `Replica` trait provides a uniform interface for accessing server metadata across the public and internal APIs.

`ServerVersion` stores the server distribution name and version string, retrieved via the `server_version` RPC.

#### TLS configuration

`DriverTlsConfig` replaces the previous boolean flag with explicit configuration:
- `disabled()` — plaintext, not recommended for production
- `enabled_with_native_root_ca()` — uses system trust roots (default)
- `enabled_with_root_ca(path)` — custom CA certificate for self-signed or internal CAs

### Language spefifics

All language drivers implement the same cluster-aware API:

**Java** (`java/`):
- `Driver` constructor accepts `DriverOptions` with TLS config and failover settings
- `ServerRouting` enum: `Auto` / `new Direct(address)`
- `Server` interface with `id()`, `role()`, `term()`, `address()`, `version()`
- `ReplicationRole` enum: `PRIMARY`, `CANDIDATE`, `SECONDARY`

**Python** (`python/`):
- `TypeDB.driver(addresses, credentials, driver_options)` factory
- `ServerRouting.Auto()` / `ServerRouting.Direct(address)`
- `Server` class with replication properties
- Moved `Database` and `DatabaseManager` to `typedb.api.database` package

**C#** (`csharp/`):
- `TypeDB.Driver(addresses, credentials, driverOptions)` constructor
- `ServerRouting.Auto` / `ServerRouting.Direct`
- `IServer` interface with `ReplicationRole` enum
- `DriverTlsConfig` with builder pattern

**C** (`c/`):
- New types: `Credential`, `DriverOptions`, `DriverTlsConfig`, `Server`, `ServerRouting`
- C API functions: `driver_new`, `driver_servers`, `server_get_address`, etc.
- SWIG bindings updated for Java and C# interop, fixed missing or excessive fields

**HTTP-TypeScript** (`http-ts/`):
- `Server` type with replication metadata
- `ServersListResponse` for the `/servers` endpoint

### Testing infrastructure

- BDD test steps updated for cluster scenarios across all languages
- Test encryption certificates regenerated
- `start-cluster-servers.sh` updated for new cluster configuration
- `tool/test/resources/config.yml` added for test server configuration

### Known limitations

- Blocking `thread::sleep` in async context during primary seeking (FIXME in `server_manager.rs:365`)


.circleci/windows/java/test_deploy_snapshot.bat
```diff
@@ -25,7 +25,7 @@ powershell -Command "Move-Item -Path bazel-typedb-driver\external\typedb_artifac
 7z x typedb-all-windows.zip
 RD /S /Q typedb-all-windows
 powershell -Command "Move-Item -Path typedb-all-windows-* -Destination typedb-all-windows"
-START /B "" typedb-all-windows\typedb server --development-mode.enable=true
+START /B "" typedb-all-windows\typedb server --development-mode.enabled=true --server.http.enabled=false
```
farost, line 28: bbbuuuuuugggggg

.circleci/config.yml
```diff
@@ -134,21 +134,21 @@ commands:
     steps:
       - install-pip-requirements
       - run: |
-          tool/test/start-community-server.sh
+          tool/test/start-core-server.sh
```
farost, line 137: `community` is very confusing, and we agreed that we use `core` when we talk about the repo, not the product.

.circleci/config.yml
```diff
@@ -793,7 +793,8 @@ jobs:
           bazel-arch: amd64
       - deploy-crate-release-unix
       - deploy-maven-release-unix
-      - deploy-http-ts-npm-release-unix
+# TODO: http-ts is disabled in the clustering branch
```
farost: TODO: Uncomment

.circleci/config.yml
```diff
@@ -845,32 +846,32 @@ workflows:
       - deploy-snapshot-linux-arm64:
           filters:
             branches:
-              only: [master]
+              only: [master, cluster-support-feature-branch]
```
farost: TODO: remove these

.factory/automation.yml
```diff
@@ -104,49 +104,52 @@ build:
...
-    #     exit $TEST_SUCCESS
+    test-rust-behaviour-cluster:
+      image: typedb-ubuntu-20.04 # Ubuntu 20.04 has GLIBC version 2.31 (2020) which we should verify to compile against
+      command: |
+        export ARTIFACT_USERNAME=$REPO_TYPEDB_USERNAME
+        export ARTIFACT_PASSWORD=$REPO_TYPEDB_PASSWORD
+        bazel run --config=ci @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
+        bazel run --config=ci @typedb_dependencies//distribution/artifact:create-netrc
+ 
+        # TODO: run `bazel test --config=ci //rust/tests/behaviour/...` instead of multiple tests (cannot do Clustered migration now)
```
farost, line 144: Will be available later in the future

farost: I could add an ignore to the BDD steps declarations, but it would be unfair, because we still expect them to work as a requirement!

c/src/common/error.rs
farost, line 1: Refactoring the packaging in the `c` dir a little so it's easier to coordinate things

c/src/common/mod.rs
```diff
@@ -42,3 +46,10 @@ pub extern "C" fn string_iterator_next(it: *mut StringIterator) -> *mut c_char {
 pub extern "C" fn string_iterator_drop(it: *mut StringIterator) {
     free(it);
 }
+
+pub(crate) fn iterators_to_map<K: Eq + Hash, V>(
```
farost, line 50: Address translation

c/src/concept/concept.rs
```diff
@@ -84,22 +76,6 @@ impl DatetimeAndTimeZone {
             is_fixed_offset,
         }
     }
-
-    pub fn get_datetime_in_nanos(self) -> DatetimeInNanos {
```
farost, line 88: SWIG generates its own thing

farost: Although C getters might require these since the fields are not public... TODO check!

c/src/concept/mod.rs
```diff
@@ -21,30 +21,15 @@ use std::ptr::addr_of_mut;
...
+    iterator::{iterator_try_next, CIterator},
+    memory::{free, take_ownership},
+};
 
 mod concept;
 mod instance;
 
-/// Promise object representing the result of an asynchronous operation.
-/// Use \ref concept_promise_resolve(ConceptPromise*) to wait for and retrieve the resulting boolean value.
-pub struct ConceptPromise(BoxPromise<'static, Result<Option<Concept>>>);
```
farost, line 32: Not used

c/src/driver.rs
```diff
@@ -0,0 +1,249 @@
...
+
+// TODO: These constructors get out of hand. Consider driver builders based on a config!
+
+/// Open a TypeDB C Driver to a TypeDB server available at the provided address.
+///
+/// @param address The address on which the TypeDB Server is running
+/// @param credentials The <code>Credentials</code> to connect with
+/// @param driver_options The <code>DriverOptions</code> to connect with
+#[no_mangle]
+pub extern "C" fn driver_new(
```
farost: Renamed `driver_open` to `driver_new`. `open` is an outdated name from back when it was `connection_open`. This will break the other libs that already depend on our headers, but it's not the only breaking change, so I decided that it's time for a nice cleanup.

c/swig/typedb_driver_csharp.swg
```diff
@@ -150,17 +150,37 @@
 %noexception user_get_name;
 %noexception user_get_password_expiry_seconds;
 
-/* Replica info getters do not throw */
```
farost, line 153: Counted, compared between csharp and java, removed the unused records, and ensured that every existing method is covered equally everywhere. Reordered so it's easier to compare all the alternatives in the future.

c/swig/typedb_driver_java.swg
```diff
@@ -439,8 +458,6 @@
 
 %promise(AnalyzedQuery, analyzed_query)
 %promise(QueryAnswer, query_answer)
-%promise(Concept, concept)
```
farost, line 442: Removed

c/swig/typedb_driver_java.swg
```diff
@@ -439,8 +458,6 @@
...
-%promise(Rule, rule)
```
farost, line 443: Trash left from the 2.x days

c/docs_structure.bzl
```diff
@@ -21,127 +21,117 @@
...
-
-    # Value types (re-exported from Rust driver)
-    "decimal": "value",
-    "duration": "value",
-    "timezone": "value",
-    "kind": "concept",
-    "valuetype": "concept",
-
     # Analyze
+    "analyzed": "analyze",
```
farost, line 25: Adjusted to the same ordering and typing in all the `x/docs_structure.bzl`

csharp/Analyze/Function.cs
```diff
@@ -24,6 +24,8 @@
 using TypeDB.Driver.Api.Analyze;
 using TypeDB.Driver.Common;
 
+using AnalyzeError = TypeDB.Driver.Common.Error.Analyze;
```
farost, line 27: Mirrors the `java` impl. The previous implementors (claude) were lazy.

csharp/Api/DriverTlsConfig.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <summary>
+        /// Creates a TLS configuration with TLS disabled.
+        /// WARNING: Disabling TLS causes credentials and data to be transmitted in plaintext.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// DriverTlsConfig tlsConfig = DriverTlsConfig.Disabled();
+        /// </code>
+        /// </example>
+        public static DriverTlsConfig Disabled()
```
farost, line 52: Configuring TLS is much easier now. No more combinations of boolean flags and exceptions for unclear states. Just an enum of three options.

csharp/Test/Behaviour/Query/QuerySteps.cs
```diff
@@ -124,8 +124,8 @@ private IValue GetRowGetValue(int rowIndex, string conceptKind, string variable,
         private IQueryAnswer ExecuteQuery(string queryText)
         {
             if (_queryOptions != null)
-                return Tx.Query(queryText, _queryOptions).Resolve();
-            return Tx.Query(queryText).Resolve();
+                return Tx.Query(queryText, _queryOptions).Resolve()!;
+            return Tx.Query(queryText).Resolve()!;
```
farost, line 128: Remove warnings

http-ts/src/index.ts
```diff
@@ -131,6 +132,10 @@ export class TypeDBHttpDriver {
         return this.apiGet(`/v1/version`);
     }
 
+    getServers(): Promise<ApiResponse<ServersListResponse>> {
```
farost, line 174: New HTTP endpoints that mirror gRPC updates

http-ts/tests/behaviour/feature/BUILD
```diff
@@ -16,34 +16,37 @@
...
-    data = [],
 )
 
 typedb_behaviour_http_ts_test(
     name = "concept",
     features = ["@typedb_behaviour//driver:concept.feature"],
-    data = [],
+)
+
+# Cluster-only test (no core equivalent)
```
farost, line 48: Spent quite some time making the encryption work. Not sure if it works 100%, will finish it after the rebases

http-ts/tests/behaviour/steps/connection.ts
```diff
@@ -57,3 +63,58 @@ Then("connection has {int} user(s)", async (expectedUserCount: number) => {
 });
 
 When("connection closes", closeConnection);
+
+// Cluster/Replica steps
```
farost: New steps that can be implemented in HTTP (no server routing, etc)

http-ts/tests/behaviour/steps/context.ts
```diff
@@ -89,12 +93,105 @@ export function setConcurrentAnswers(answers: QueryResponse[]) {
...
+
+export const isClusterMode = process.env.TYPEDB_CLUSTER_MODE === "true";
+export const CLUSTER_ADDRESSES = [
+    "https://127.0.0.1:18000",
+    "https://127.0.0.1:28000",
+    "https://127.0.0.1:38000"
+];
+
+// For cluster mode tests, replace global fetch with one that handles mTLS
+if (isClusterMode) {
```
farost: TODO: ensure it works nicely and clean up if needed

rust/src/common/address/address_translation.rs
```diff
@@ -0,0 +1,44 @@
...
+ * under the License.
+ */
+
+use std::collections::HashMap;
+
+use crate::common::address::Address;
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub(crate) enum AddressTranslation {
+    // public : private
```
farost, line 26: Could not decide how to make the interface more explicit (not to confuse public and private)

rust/src/common/address/addresses.rs
```diff
@@ -0,0 +1,232 @@
...
+    fmt::{Formatter, Write},
+};
+
+use itertools::Itertools;
+
+use crate::common::address::{address_translation::AddressTranslation, Address};
+
+/// A collection of server addresses used for connection.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum Addresses {
```
farost, line 32: For ease of use and explicitness, addresses are now an `enum`. Yes, I love enums.

rust/src/common/error.rs
```diff
@@ -128,50 +128,48 @@ error_messages! { ConnectionError
...
-    DatabaseNotFound { name: String } =
-        8: "Database '{name}' not found.",
+        8: "The transaction is closed because of the error(s):\n{errors}",
     MissingResponseField { field: &'static str } =
         9: "Missing field in message received from server: '{field}'. This is either a version compatibility issue or a bug.",
     UnknownRequestId { request_id: RequestID } =
         10: "Received a response with unknown request id '{request_id}'",
     UnexpectedResponse { response: String } =
         11: "Received unexpected response from server: '{response}'. This is either a version compatibility issue or a bug.",
-    InvalidResponseField { name: &'static str } =
```
farost, line 150: Decided to break the enumeration like in server. Not sure when it's time to freeze the indices

rust/src/common/error.rs
```diff
@@ -413,45 +448,64 @@ impl From<Status> for Error {
     fn from(status: Status) -> Self {
```
farost, line 444: Tried to make it nicer

rust/src/connection/network/proto/server.rs
```diff
@@ -0,0 +1,69 @@
...
+    common::Result,
+    connection::server::{server_version::ServerVersion, ReplicationRole, ReplicationStatus, Server},
+    error::ConnectionError,
+};
+
+impl TryFromProto<ServerProto> for Server {
+    fn try_from_proto(proto: ServerProto) -> Result<Server> {
+        let replication_status =
+            proto.replica_status.map(|status| ReplicationStatus::try_from_proto(status)).transpose()?;
+        match proto.address {
```
farost, line 36: We receive only one address from proto: connection address!

rust/src/connection/network/transmitter/rpc.rs
```diff
@@ -150,37 +155,15 @@ impl RPCTransmitter {
...
-                            })),
-                        }
-                    }
-                    Some(Ok(other)) => {
-                        Err(Error::Connection(ConnectionError::UnexpectedResponse { response: format!("{other:?}") }))
-                    }
-                    Some(Err(status)) => Err(status.into()),
-                    None => Err(Error::Connection(ConnectionError::UnexpectedConnectionClose)),
-                }
+                let timeout = rpc.request_timeout();
```
farost, line 162: Transaction opening is under the request timeout! Commands within transactions are not affected by request timeout.

rust/src/connection/network/channel.rs
```diff
@@ -54,11 +55,22 @@ pub(super) fn open_callcred_channel(
     credentials: Credentials,
     driver_options: DriverOptions,
 ) -> Result<(CallCredChannel, Arc<CallCredentials>)> {
-    let mut builder = Channel::builder(address.into_uri());
-    if driver_options.is_tls_enabled() {
-        let tls_config =
-            driver_options.tls_config().clone().expect("TLS config object must be set when TLS is enabled");
-        builder = builder.tls_config(tls_config)?;
+    let is_tls_enabled = driver_options.tls_config.is_enabled();
```
farost, line 58: An explicit TLS configuration

rust/src/connection/server/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+pub(crate) mod server_connection;
+pub(crate) mod server_manager;
+pub mod server_routing;
+pub mod server_version;
+
+use crate::common::address::{address_translation::AddressTranslation, Address};
+
+pub const DEFAULT_SERVER_ID: u64 = 0;
+
+pub trait Replica: Clone {
```
farost, line 29: Split servers to `AvailableServer` and `UnavailableServer` for Rust only. `AvailableServer`s can be communicated with, `Unavailable` ones are only for the info. Helps type checking a lot.

rust/src/connection/server/server_manager.rs
farost, line 1: A new component that manages everything related to servers, including connections!

farost: Driver becomes just a wrapper for the logic.

rust/src/connection/server/server_manager.rs
```diff
@@ -0,0 +1,488 @@
...
+    credentials: Credentials,
+    driver_options: DriverOptions,
+    driver_lang: String,
+    driver_version: String,
+}
+
+impl ServerManager {
+    const PRIMARY_REPLICA_SELECTION_TIMEOUT: Duration = Duration::from_secs(2);
+
+    // TODO: Introduce a timer-based connections update to have a more actualized connections list
```
farost, line 73: User-facing methods always do the network calls to receive the most recent info. If it happens, we update our caches. However, the caches can be stale without user calls, and, say, an inactive driver can live for a day, and then start sending request to an outdated primary. It will require some time to find a new primary. Also, the configuration could change drastically. 

Not really needed, but it might be nice to have. I'm currently inclined to remove this, especially with the introduction of hints in the GRPC errors (I don't yet use them here). TODO!

rust/src/connection/server/server_manager.rs
```diff
@@ -0,0 +1,488 @@
...
+    #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
+    pub(crate) async fn new(
+        background_runtime: Arc<BackgroundRuntime>,
+        addresses: Addresses,
+        credentials: Credentials,
+        driver_options: DriverOptions,
+        driver_lang: impl AsRef<str>,
+        driver_version: impl AsRef<str>,
+    ) -> Result<Self> {
+        let (source_connections, replicas) = Self::fetch_servers_from_addresses(
```
farost, line 84: We use the addresses passed to the driver only for the initial connection. You can pass a single address from a cluster of 10 -- the driver will fetch and automatically remember all the peers of this single replica, because the replica returns them. It returns the `connection-address`es of the peers, not their serving addresses, so it just... works out of the box!

rust/src/connection/server/server_manager.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    pub(crate) fn username(&self) -> Result<String> {
+        match self.read_replica_connections().iter().next() {
+            Some((_, replica_connection)) => Ok(replica_connection.username().to_string()),
+            None => Err(ConnectionError::ServerConnectionIsClosed {}.into()),
+        }
+    }
+
+    #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
+    pub(crate) async fn execute<F, P, R>(&self, server_routing: ServerRouting, task: F) -> Result<R>
```
farost, line 196: Execute based on the server routing

rust/src/connection/server/server_manager.rs
```diff
@@ -0,0 +1,488 @@
...
+    where
+        F: Fn(ServerConnection) -> P,
+        P: Future<Output = Result<R>>,
+    {
+        match server_routing {
+            ServerRouting::Auto => self.execute_strongly_consistent(task).await,
```
farost, line 202: Currently, if the routing is automatic, we execute strongly (by picking the current primary replica).  In the future, we can have other configurations for execution strategies (e.g., find the closest and most easily accessible replica for a read operation)

rust/src/connection/server/server_manager.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
+    async fn execute_on_any<F, P, R>(&self, replicas: impl IntoIterator<Item = AvailableServer>, task: F) -> Result<R>
+    where
+        F: Fn(ServerConnection) -> P,
+        P: Future<Output = Result<R>>,
+    {
+        let mut connection_errors = HashMap::new();
+        for replica in replicas.into_iter() {
+            // TODO: We won't ever reconnect to disconnected / new replicas when using execute_on_any.
```
farost, line 276: TODO for myself

rust/src/connection/server/server_manager.rs
```diff
@@ -0,0 +1,488 @@
...
+                if &result == &Err(Error::Connection(ConnectionError::NoPrimaryServer {})) {
+                    // TODO: Cannot actually happen
+                    Self::wait_for_primary_replica_selection().await;
+                }
+                result
+            })
+            .await;
+
+        match result {
+            // TODO: We may not check newly added configured addresses, and we need to somehow integrate it here...
```
farost: TODO for myself

rust/src/connection/server/server_manager.rs
```diff
@@ -0,0 +1,488 @@
...
+            self.refresh_replica_connections().await?;
+            Ok(replica)
+        } else {
+            Err(ConnectionError::NoPrimaryServer {}.into())
+        }
+    }
+
+    #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
+    async fn wait_for_primary_replica_selection() {
+        // FIXME: blocking sleep! Can't do agnostic async sleep.
```
farost, line 368: Copied from the 2.x implementation.

rust/src/connection/driver_options.rs
```diff
@@ -17,46 +17,77 @@
...
+const DEFAULT_REDIRECT_FAILOVER_RETRIES: usize = 1;
+
+/// TypeDB driver connection options.
+/// `DriverOptions` object can be used to override the default driver behavior while connecting to
+/// TypeDB.
+///
+/// # Examples
+///
+/// ```rust
+/// let options = DriverOptions::new(DriverTlsConfig::default()).request_timeout(Duration::from_secs(30));
```
farost, line 35: New API for `DriverOptions` -- same as for other options in the lib.

rust/src/database/database.rs
```diff
@@ -202,276 +178,4 @@ impl Database {
...
-                    debug!("Unable to connect to {}. Attempting next server.", replica.server);
-                }
-                res => return res,
-            }
-        }
-        Err(Self::unable_to_connect_error(&self.server_connections))
-    }
-
-    #[cfg_attr(feature = "sync", maybe_async::must_be_sync)]
-    pub(super) async fn run_on_primary_replica<F, P, R>(&self, task: F) -> Result<R>
```
farost, line 242: You can see that I've stolen some of the ideas from before. Now, it got reshaped and moved to `server_manager`.

rust/src/driver.rs
```diff
@@ -67,21 +62,17 @@ impl TypeDBDriver {
     /// ```rust
     #[cfg_attr(
         feature = "sync",
-        doc = "TypeDBDriver::new(\"127.0.0.1:1729\", Credentials::new(\"username\", \"password\"), DriverOptions::new(true, None))"
+        doc = "TypeDBDriver::new(Addresses::try_from_address_str(\"127.0.0.1:1729\").unwrap(), Credentials::new(\"username\", \"password\"), DriverOptions::new(true, None))"
```
farost, line 65: New API for parsing addresses

tool/test/resources/encryption/ext-grpc-certificate.pem
farost, line 1: They got outdated

tool/test/resources/encryption/int-grpc-root-ca.pem
farost, line 1: Can be used for internal clustering protocols of clusters

tool/test/start-cluster-servers.sh
farost, line 1: New logic for clustered typedb!

RELEASE_TEMPLATE.md
```diff
@@ -1,3 +1,6 @@
+**This is an alpha release for CLUSTERED TypeDB 3.x. Do not use this as a stable version of TypeDB.**
```
farost: TODO: remove

.circleci/windows/csharp/test_deploy_snapshot.bat
```diff
@@ -26,6 +26,7 @@ powershell -Command "Move-Item -Path bazel-typedb-driver\external\typedb_artifac
 RD /S /Q typedb-all-windows
 powershell -Command "Move-Item -Path typedb-all-windows-* -Destination typedb-all-windows"
 START /B "" typedb-all-windows\typedb server --development-mode.enable=true
+START /B "" typedb-all-windows\typedb server --development-mode.enabled=true --server.http.enabled=false
```
farost, line 28: bbuuuuuugggg

rust/tests/integration/cluster/clustering.rs
farost, line 1: This test will be a `failback` test, working on it in the background! Should be done soon we a major enhancement of the test scripts

.circleci/windows/csharp/test_deploy_snapshot.bat
flyingsilverfin: why did we have to disable http? Ahh it would clash with the other server? Could change port or no? Doesn't grpc port clash too?

.factory/automation.yml
```diff
@@ -49,7 +49,7 @@ build:
         bazel run --config=ci @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
         bazel run --config=ci @typedb_dependencies//distribution/artifact:create-netrc
         # TODO: Temporarily build only the 3.0-reimplemented drivers:
-        # bazel build //...
+        # bazel build --config=ci //...
```
flyingsilverfin, line 52: i believe we haven't re-enabled C++ yet? does it build? if not, just make sure the list is complete

.factory/automation.yml
flyingsilverfin: I dont mind our current setup of just running them one at a time wdyt?

c/src/concept/concept.rs
```diff
@@ -345,9 +321,9 @@ pub extern "C" fn concept_get_datetime(concept: *const Concept) -> DatetimeInNan
 /// Returns the value of this datetime-tz value concept as seconds and nanoseconds parts since the start of the UNIX epoch and timezone information.
 /// If the value has another type, the error is set.
 #[no_mangle]
-pub extern "C" fn concept_get_datetime_tz(concept: *const Concept) -> DatetimeAndTimeZone {
+pub extern "C" fn concept_get_datetime_tz(concept: *const Concept) -> *mut DatetimeAndTimeZone {
     match borrow(concept).try_get_datetime_tz() {
-        Some(value) => DatetimeAndTimeZone::new(&value),
+        Some(value) => release(DatetimeAndTimeZone::new(&value)),
```
flyingsilverfin, line 334: memory issue before?

c/src/concept/concept.rs
flyingsilverfin: wait what do you mean - these are in the c API already?

c/src/server/server_routing.rs
```diff
@@ -0,0 +1,112 @@
...
+    error::unwrap_or_default,
+    memory::{borrow_optional, free, release, release_string, string_free, string_view},
+};
+
+/// <code>ServerRoutingTag</code> is used to represent server routing directives in FFI.
+/// It is the tag part, which is combined with optional fields to form an instance of the original
+/// enum.
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+pub enum ServerRoutingTag {
```
flyingsilverfin: Tag... maybe Type or Method?

c/src/server/server_version.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use std::{ffi::c_char, ptr::null_mut};
+
+use crate::common::memory::{free, release_string, string_free};
+
+/// <code>ServerVersion</code> is an FFI representation of a full server's version specification.
+#[repr(C)]
+#[derive(Debug)]
+pub struct ServerVersion {
+    distribution: *mut c_char,
```
flyingsilverfin, line 29: do we always (and also need to) hold mut?

c/src/driver.rs
```diff
@@ -0,0 +1,190 @@
...
+///        Nullable: pass NULL to default to "c".
+#[no_mangle]
+pub extern "C" fn driver_new_with_address_translation(
+    public_addresses: *const *const c_char,
+    private_addresses: *const *const c_char,
+    credentials: *const Credentials,
+    driver_options: *const DriverOptions,
+    driver_lang: *const c_char,
+) -> *mut TypeDBDriver {
+    let addresses = iterators_to_map(string_array_view(public_addresses), string_array_view(private_addresses));
```
flyingsilverfin, line 105: will this just stop short if the arrays are mismatched, or error?

c/swig/typedb_driver_csharp.swg
flyingsilverfin: fantastic, thank you!

flyingsilverfin: my goodness i hope we've not got any memory leaks anywhere. Actually, after this merges maybe we can write a little stress test in each driver that opens drivers and transaction, doing a small bulk load, to see if the memory just increases :?

csharp/Answer/ConceptRow.cs
```diff
@@ -87,6 +89,7 @@ public QueryType QueryType
         /// <inheritdoc/>
         public IConcept? GetIndex(long columnIndex)
         {
+            Validator.RequireNonNegative(columnIndex, nameof(columnIndex));
```
flyingsilverfin, line 92: nice catch

csharp/Api/Server/IServer.cs
```diff
@@ -0,0 +1,52 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+namespace TypeDB.Driver.Api
+{
+    /// <summary>
+    /// The metadata and state of an individual raft server of a driver connection.
```
flyingsilverfin: from the user point of view we generally shouldn't expose implementation (raft) - they only need to know about Clusters!

csharp/Api/DriverTlsConfig.cs
flyingsilverfin: great

csharp/Api/IDriver.cs
```diff
@@ -43,31 +40,30 @@ public interface IDriver : IDisposable
...
         /// <code>
-        /// driver.Users
+        /// driver.Users;
         /// </code>
         /// </example>
         IUserManager Users { get; }
 
         /// <summary>
-        /// Opens a transaction to perform read or write queries on the database.
+        /// Opens a communication tunnel (transaction) to the given database on the running TypeDB server.
```
flyingsilverfin: try to avoid new terminology - 'tunnel' isn't something we every use. You can just say 'transaction' here it's fine

csharp/Api/ServerRouting.cs
```diff
@@ -0,0 +1,93 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+namespace TypeDB.Driver.Api
+{
+    /// <summary>
+    /// Server routing directive for operations against a distributed server. All driver methods have
```
flyingsilverfin, line 23: distributed 'cluster' imo

csharp/Test/Deployment/NugetApplicationTest.cs
```diff
@@ -37,7 +37,7 @@ public static void Main()
 
             string dbName = "nuget-test-database";
 
-            using (var driver = TypeDB.Driver(TypeDB.DefaultAddress, new Credentials("admin", "password"), new DriverOptions(false, null)))
+            using (var driver = TypeDB.Driver(TypeDB.DefaultAddress, new Credentials("admin", "password"), new DriverOptions(DriverTlsConfig.Disabled())))
```
flyingsilverfin, line 40: Much better yea!

csharp/Test/Integration/FailoverTest.cs
```diff
@@ -0,0 +1,262 @@
...
+        };
+        private const string Username = "admin";
+        private const string Password = "password";
+        private const string DatabaseName = "test-failover";
+        private const int FailoverIterations = 10;
+        private const int PrimaryPollRetries = 20;
+        private const int PrimaryPollIntervalSecs = 2;
+        private const int PrimaryFailoverRetries = 5;
+        private static readonly string ClusterServerScript =
+            Environment.GetEnvironmentVariable("CLUSTER_SERVER_SCRIPT")
```
flyingsilverfin, line 47: do you use this to start up and shut down? How do you shut down?

csharp/TypeDB.cs
```diff
@@ -17,39 +17,66 @@
...
-    /// Entry point for creating TypeDB driver connections.
-    /// </summary>
     public static class TypeDB
     {
         /// <summary>
-        /// The default address for TypeDB Server.
+        /// The default address of the TypeDB server.
         /// </summary>
-        public const string DefaultAddress = "localhost:1729";
+        public const string DefaultAddress = "127.0.0.1:1729";
```
flyingsilverfin, line 32: Interesting - why the change?

dependencies/typedb/artifacts.bzl
```diff
@@ -25,20 +25,15 @@ def typedb_artifact():
...
-
-#maven_artifacts = {
-#    'com.typedb:typedb-runner': '2.28.3',
-#    'com.typedb:typedb-cloud-runner': '2.28.3',
-#}
+def typedb_cluster_artifact():
+    native_artifact_files(
+        name = "typedb_cluster_artifact",
+        group_name = "typedb-cluster-all-{platform}",
+        artifact_name = "typedb-cluster-all-{platform}-{version}.tar.gz", # TODO: Make {ext} instead of tar.gz
```
flyingsilverfin: remember this todo

flyingsilverfin (changes requested):
I've left a set of comments on csharp for now, assuming that these can broadly be applied throughout? I'm not sure i have bandwidth to review everything! Perhaps in the next pass I can review Python since that's the most popular entry point!

.factory/automation.yml
farost: This will be eliminated once database import is available in clusters. Since you don't mind the current set up, we'll leave it as is for now

c/src/concept/concept.rs
farost: All the C struct members are pub by default, so yes

c/swig/typedb_driver_csharp.swg
farost: It sounds like a very good idea, I've been striving for more stability in TypeDB and the tools. It's much easier with claude now. Created an issue: https://github.com/typedb/typedb-driver/issues/865

.circleci/windows/csharp/test_deploy_snapshot.bat
farost: We don't need the HTTP in these tests, and it started failing because of the taken port. Why change the port if we don't need it?

.factory/automation.yml
farost: The list is complete

c/src/concept/concept.rs
farost: Yeah, risky. `DatetimeAndTimeZone` contains a string and a custom `drop` for it.

c/src/server/server_routing.rs
farost: Changed to `Type`

c/src/server/server_version.rs
farost: We don't **have to**, but the type used in memory management (`release_string`), so that's what we store without casting.

c/src/driver.rs
farost: `zip_eq` asserts both iters have equal lengths, so it will panic just like in 2.x. It's okay for C, the drivers on top of it will probably use more high-level containers that will make freaking this up very hard.

csharp/Api/Server/IServer.cs
farost: I think we exposed this in docs in 2.x, but I changed all `raft` to `cluster`.

csharp/Api/ServerRouting.cs
farost: `distributed server` is `cluster` in the context of the protocol to me. We are a client, and there is a server. It contains multiple `replica`s. We did remove `replica`s in favor of `server`s, so it became a little weird generally, but this is fine in terms of the domain to me.

csharp/Test/Integration/FailoverTest.cs
farost: In order to make all the integration and behaviour tests in sync, I use this script only for killing and restarting the already running servers. You need to start up all the servers first and then shut them down manually.

csharp/TypeDB.cs
farost: Both work (unless you generate SSL certificates for `127.0.0.1` only), but I want to use explicit loopback/broadcast ips everywhere for the server and the clients to stop confusing `127.0.0.1` and `localhost`. `listen` addresses default to `0.0.0.0`, `advertise` default to `127.0.0.1`. We used to have mixed-up values in different places and sometimes in different drivers.

dependencies/typedb/artifacts.bzl
farost: Never forget

http-ts/src/index.ts
```diff
@@ -17,21 +17,25 @@
...
     QueryResponse,
+    ServersListResponse,
     SignInResponse,
     TransactionOpenResponse,
     UsersListResponse,
     VersionResponse
 } from "./response";
 
 const HTTP_UNAUTHORIZED = 401;
+const HTTP_MISDIRECTED = 421;
```
farost, line 38: We use this code instead of direct redirects, so the simplified clients (like `curl`) don't automatically hit the unauthorized endpoint

python/tests/behaviour/behave_rule.bzl
```diff
@@ -32,30 +32,29 @@ def py_behave_test(*, name, background, native_typedb_artifact, steps, feats, de
...
+def typedb_behaviour_py_test_core(name, **kwargs):
     py_behave_test(
-        name = name + "-community",
-        background = "@//python/tests/behaviour/background:community",
-        native_typedb_artifact = "@//tool/test:native-typedb-artifact",
+        name = name + "-core",
+        background = "@//python/tests/behaviour/background:core",
         toolchains = ["@rules_python//python:current_py_toolchain"],
-        typedb_port = "1729",
+        typedb_port = "1729", # TODO: Not needed?
```
flyingsilverfin, line 43: todos

python/tests/deployment/test.py
```diff
@@ -35,7 +35,8 @@ class TestDeployedPythonDriver(TestCase):
     def setUpClass(cls):
         super(TestDeployedPythonDriver, cls).setUpClass()
         global driver
-        driver = TypeDB.driver(TypeDB.DEFAULT_ADDRESS, Credentials("admin", "password"), DriverOptions(is_tls_enabled=False))
+        driver = TypeDB.driver(TypeDB.DEFAULT_ADDRESS, Credentials("admin", "password"),
```
flyingsilverfin, line 38: So in the next release, all the users will have to update their driver APIs right

python/tests/integration/cluster/BUILD
```diff
@@ -15,23 +15,28 @@
...
+    ],
+    data = [
+        "//python:native-driver-binary-link",
+        "//python:native-driver-wrapper-link",
+        "//tool/test:cluster-server.sh",
+        "//tool/test/resources:encryption/ext-grpc-root-ca.pem",
+    ],
+    env = {
+        "CLUSTER_SERVER_SCRIPT": "$(rootpath //tool/test:cluster-server.sh)",
+        "ROOT_CA": "$(rootpath //tool/test/resources:encryption/ext-grpc-root-ca.pem)",
```
flyingsilverfin, line 36: Cool - are these still valid?? amazing

python/typedb/api/answer/query_answer.py
```diff
@@ -37,8 +37,6 @@ def query_type(self) -> QueryType:
         """
         Retrieves the executed query's type of this ``QueryAnswer``.
 
-        :return:
```
flyingsilverfin, line 40: didn't we need these in order for the adoc to get generate right :? dont quite remember - as long as it looks good i dont really care!

python/example.py
```diff
@@ -8,7 +8,8 @@ class TypeDBExample:
     def typedb_example(self):
         # Open a driver connection. Specify your parameters if needed
         # The connection will be automatically closed on the "with" block exit
-        with TypeDB.driver(TypeDB.DEFAULT_ADDRESS, Credentials("admin", "password"), DriverOptions(is_tls_enabled=False)) as driver:
+        with TypeDB.driver(TypeDB.DEFAULT_ADDRESS, Credentials("admin", "password"),
```
flyingsilverfin, line 11: This should be top-line of the release notes for driver (along with other languages!)

python/rules.bzl
```diff
@@ -77,7 +77,7 @@ def native_driver_versioned(python_versions):
             ],
             url = "https://github.com/typedb/typedb-driver",
             author = "TypeDB Community",
-            author_email = "community@typedb.com",
+            author_email = "core@typedb.com",
```
flyingsilverfin: keep community in these ones pls - that's our contact email i believe

flyingsilverfin (approved):
Approved with comments

python/tests/behaviour/behave_rule.bzl
farost: Thanks

python/tests/deployment/test.py
farost: Yes

python/tests/integration/cluster/BUILD
farost: I had to regenerate the pem files, but they will last for a couple of years again

python/typedb/api/answer/query_answer.py
farost: Maaaaybe something will break on the website, but I don't see a need for these -- the docs generations look fine. If the website view is broken, I'll get it fixed with no issue

python/example.py
farost: Ok!

python/rules.bzl
farost: ohhh it's clearly a bug. I thought I caught this when changing the globals. Sorry and thanks

PR #868, opened 2026-05-11
Restore C# cluster BDD and integration tests

lolski:
## Usage and product changes
Restore C# tests in order to maintain quality control.

## Implementation
- Restore `test-csharp-behaviour-cluster` job
- Restore `test-csharp-integration` job


farost (approved):
Looking good!

PR #869, opened 2026-05-12
Fix http-ts failover logic

farost:
## Usage and product changes

When the cluster primary is no longer responsive, the HTTP-TS driver may route its next signin to a secondary that's briefly returning `SRV14 "Not yet initialised"` — e.g. a just-restarted node whose system DB isn't loaded into memory yet, or a node mid-promotion.

This made the failover incomplete and the `cluster-failover.ts` test flaky in CI: when fallback landed on a stuck secondary before a healthy origin, the driver burned the test budget on a single dead-end.

## Implementation
- `tokenErrToResult` now treats `SRV14` the same as `HDR2`: surface as `null` so the caller iterates to the next origin.
- Removed the 30s SRV14 retry loop in `refreshToken`. Fallback to a healthy origin is in the millisecond range; callers (and the test harness) already retry on returned errors, which covers the rarer "every origin is briefly transient" case at a coarser cadence.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -240,7 +240,7 @@ native_artifact_files.artifact(
     name = "typedb_cluster_artifact",
     group_name = "typedb-cluster-all-{platform}",
     artifact_name = "typedb-cluster-all-{platform}-{version}.{ext}",
-    tag = "3.10.0-alpha-1",
+    commit = "598c47cc13c8400ae64061b390bb035de46a64d0",
     private = True,
 )
 
--- a/http-ts/src/index.ts
+++ b/http-ts/src/index.ts
@@ -37,11 +37,13 @@ import {
 const HTTP_UNAUTHORIZED = 401;
 const HTTP_MISDIRECTED = 421;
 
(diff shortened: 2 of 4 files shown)


http-ts/src/index.ts
```diff
@@ -332,12 +326,7 @@ export class TypeDBHttpDriver {
     }
 
     private async refreshToken(): Promise<ApiResponse<SignInResponse>> {
-        const deadline = Date.now() + SIGNIN_TRANSIENT_RETRY_BUDGET_MS;
```
farost, line 335: This old hack was too aggressive and still not fully helpful

MODULE.bazel
```diff
@@ -240,7 +240,7 @@ native_artifact_files.artifact(
     name = "typedb_cluster_artifact",
     group_name = "typedb-cluster-all-{platform}",
     artifact_name = "typedb-cluster-all-{platform}-{version}.{ext}",
-    tag = "3.10.0-alpha-1",
+    commit = "598c47cc13c8400ae64061b390bb035de46a64d0",
```
farost, line 243: Not to depend on an alpha version

PR #874, opened 2026-05-20
Add driver for python 3.14

krishnangovindraj:
## Usage and product changes
The python driver now deploys a package supporting python 3.14


krishnangovindraj:
Updating `rules_python` leads to 
```
    from typedb.native_driver_wrapper import TypeDBDriverExceptionNative
  File "/<sandbox>/bin/python/tests/integration/test_values.runfiles/_main/python/typedb/native_driver_wrapper.py", line 10, in 
    from . import native_driver_python
```
[Advice](https://github.com/bazel-contrib/rules_python/issues/824) is to use: `@rules_python//python/cc:current_py_cc_libs` , though we have an explicit dependency of the native library on `@python_3xx//:libpython`.

Actually, we might just need to do this: https://rules-python.readthedocs.io/en/1.7.0-rc3/howto/linking-libpython.html 

And I'd also like us to bump protobuf to 33.0 or above so we get prebuilt compilers

MODULE.bazel.lock
farost, line 1: That's probably not the way

flyingsilverfin (approved):

PR #875, opened 2026-05-20
Implement "given rows" as inputs to queries

krishnangovindraj:
## Usage and product changes
Adds the machinery required for passing rows of concepts to TypeQL queries with 'given' stages, and updates analyzed query datastructures for the given stage.

#### Examples:
(Note: `eugene` and `fred` are concepts returned from previous queries)
**rust**
```rust
let mut given_rows = GivenRows::new(vec!["x".to_string(), "v".to_string()], 2);
given_rows.push_row(vec![GivenRowEntry::Entity(eugene), GivenRowEntry::Value(Value::Integer(12))]).unwrap();
given_rows.push_row(vec![fred.into(), 34.into()]).unwrap(); // From<T> is implemented for many native values, and concepts.
let query = "given $x: person, $v: integer; insert $x has age == $v;";
let inserted_answer = transaction.query_with_rows(query, given_rows).await.unwrap();
```

**Java**
```java
String query = "given $x: person, $v: integer; insert $x has age == $v;";
List<Map<String, Object>> givenRows = List.of(
    java.util.Map.ofEntries(
        Map.entry("x", personEugene),
        Map.entry("v", TypeDB.Concept.newInteger(12))
    ),
    java.util.Map.ofEntries(
        Map.entry("x", personFred),
        Map.entry("v", TypeDB.Concept.newInteger(34))
    )
);
QueryAnswer inserted = transaction.query(query, givenRows).resolve();
```

**python**
```
query = "given $x: person, $v: integer; insert $x has age == $v;"
given_rows = [
    {"x": person_eugene, "v": TypeDB.Concept.new_integer(12)},
    {"x": person_fred, "v": TypeDB.Concept.new_integer(34)},
]
inserted = tx.query(query, given_rows=given_rows).resolve()
```

**C#**
```c#
 var query = "given $x: person, $v: integer; insert $x has age == $v;";
var givenRows = new List<Dictionary<string, object?>>
{
    new Dictionary<string, object?> { { "x", personEugene }, { "v", TypeDB.Concept.NewInteger(12) } },
    new Dictionary<string, object?> { { "x", personFred }, { "v", TypeDB.Concept.NewInteger(34) } }
};
  var inserted = transaction.Query(query, new QueryOptions(), givenRows).Resolve()!;
```

The HTTP QueryRequest has expects the (backward compatible) format:
```typescript
export type QueryRequest = { 
    query: string,
    queryOptions: QueryOptions | undefined, 
    givenRows:  GivenRows | undefined,
}

export type GivenRows = { [varName: string]: GivenRowEntry }[];

export type GivenRowEntry = Value | Entity | Relation | Attribute
    | boolean // for boolean values
    | number  // For integer and double values
    | string; // For all other types, as well as IIDs.
```


csharp/Connection/Transaction.cs
```diff
@@ -152,6 +152,53 @@ public Promise<IQueryAnswer> Query(string query, QueryOptions options)
...
+                {
+                    var concept = rowEntries[i];
+                    if (concept == null)
+                    {
+                        Pinvoke.typedb_driver.given_row_set_index_to_empty(nativeRow, (uint)i);
+                    }
+                    else
+                    {
+                        if (concept.IsType())
+                            throw new TypeDBDriverException(DriverError.INVALID_TYPE_AS_GIVEN_INPUT, concept.GetLabel(), rowIndex);
```
krishnangovindraj: These can be replaced by C side errors.

csharp/Test/Behaviour/Concept/ConceptSteps.cs
```diff
@@ -611,21 +611,6 @@ public void AnswerGetRowGetConceptKindTryGetValueTypeIsNot(
                 $"Value should not equal '{notExpectedValue}' but it does");
         }
 
-        // Pattern: answer get row(N) get attribute(VAR) get value is: VALUE (non-quoted)
```
krishnangovindraj, line 614: Moved so we can access it from elsewhere too

dependencies/typedb/repositories.bzl
krishnangovindraj, line 1: No idea where this snuck in from

csharp/Concept/ValueFactory.cs
krishnangovindraj, line 1: Architectural question 1: Where should the methods to create new values live?

csharp/Test/Integration/Data/ValuesTest.cs
```diff
@@ -619,6 +621,62 @@ public void TestDatetimeTZFixedOffset()
             }
         }
 
+        [Test]
```
krishnangovindraj, line 623: This is how we test the value constructors in each language.

http-ts/src/analyze.ts
krishnangovindraj, line 1: TODO: The http driver doesn't have the types for given_rows yet.

rust/src/common/error.rs
```diff
@@ -315,6 +315,7 @@ pub enum Error {
     Internal(InternalError),
     Server(ServerError),
     Other(String),
+    FFI(String),
```
krishnangovindraj, line 333: Could have used "Other", but why not be more specific.

rust/src/connection/network/proto/analyze.rs
```diff
@@ -20,24 +20,25 @@
...
     common::Result,
-    concept::{Kind, Value, ValueType, type_},
+    concept::{type_, Kind, Value, ValueType},
     connection::{
         message::AnalyzeResponse,
         network::proto::{FromProto, TryFromProto},
     },
     error::{AnalyzeError, ServerError},
 };
+use crate::analyze::Given;
```
krishnangovindraj: Needs a rustfmt

rust/src/connection/network/proto/concept.rs
```diff
@@ -344,3 +346,76 @@ impl TryFromProto<i32> for Kind {
 fn naive_datetime_from_timestamp(seconds: i64, nanos: u32) -> NaiveDateTime {
     DateTime::from_timestamp(seconds, nanos).unwrap().naive_utc()
 }
+
+// Into proto for selected concepts
```
krishnangovindraj, line 349: Based on reversing the `try_from_proto` implementations.

rust/src/connection/message.rs
```diff
@@ -38,6 +38,8 @@ use crate::{
     error::ServerError,
     info::UserInfo,
 };
+use crate::concept::{Attribute, Entity, Relation, Value};
```
krishnangovindraj: unused?

rust/src/transaction.rs
```diff
@@ -87,10 +88,23 @@ impl Transaction {
...
+
+    pub fn query_with_inputs(&self, query: impl AsRef<str>, inputs: QueryGivenRows) -> impl Promise<'static, Result<QueryAnswer>> {
+        self.query_with_options_and_inputs(query, QueryOptions::new(), Some(inputs))
+    }
+
+    pub fn query_with_options_and_inputs(
+        &self,
+        query: impl AsRef<str>,
+        options: QueryOptions,
+        inputs: Option<QueryGivenRows>
```
krishnangovindraj: Do we want this to be non-optional given the name is "and_inputs"? 
I might prefer renaming it because it's nice to have a version where I can just pass `None` instead of having to use a different function

krishnangovindraj (commented):
one self review

python/typedb/connection/transaction.py
```diff
@@ -72,11 +76,16 @@ def analyze(self, query: str) -> Promise[AnalyzedQuery]:
         promise = transaction_analyze(self.native_object, query)
         return Promise.map(_AnalyzedQuery, lambda: analyzed_query_promise_resolve(promise))
 
-    def query(self, query: str, options: Optional[QueryOptions] = None) -> Promise[QueryAnswer]:
+    def query(self, query: str, options: Optional[QueryOptions] = None, given_rows: Optional[List[List[Optional["Concept"]]]] = None) -> Promise[QueryAnswer]:
```
flyingsilverfin: I think this might read better with a dict `{ "x": Value::integer(10), "y": Value::string("john") }`

rust/tests/behaviour/steps/query.rs
```diff
@@ -171,22 +171,49 @@ fn concept_get_type(concept: &Concept) -> Concept {
...
+    context.given_rows = Some(QueryGivenRows(given_rows))
 }
 
 #[apply(generic_step)]
-#[step(expr = "get answers of typeql schema query")]
-#[step(expr = "get answers of typeql write query")]
-#[step(expr = "get answers of typeql read query")]
-pub async fn get_answers_of_typeql_query(context: &mut Context, step: &Step) {
+#[step(expr = "get answers of typeql schema query{with_given}")]
+#[step(expr = "get answers of typeql write query{with_given}")]
```
flyingsilverfin, line 246: Ah it's here ok

rust/tests/behaviour/steps/query.rs
```diff
@@ -171,22 +171,49 @@ fn concept_get_type(concept: &Concept) -> Concept {
...
     context.cleanup_answers().await;
-    may_error.check(run_query(context.transaction(), step.docstring().unwrap(), context.query_options).await);
+    let given_rows = may_take_given_rows(context, with_given);
+    may_error.check(run_query(context.transaction(), step.docstring().unwrap(), given_rows, context.query_options).await);
+}
+
+
+#[cucumber::given(expr = "set answers of typeql read query as given rows with order: {variable_list}")]
+#[cucumber::when(expr = "set answers of typeql read query as given rows with order: {variable_list}")]
+async fn set_given_rows(context: &mut Context, var_list: VariableList, step: &Step) {
```
flyingsilverfin, line 193: Was coming over here to see an example of how the API looks now. That being said - you're not constructing the Given inputs in the steps with this approach :D

Can you update the root examples in each language to show how it would be used please - helpful for me and users as docs.

flyingsilverfin:
I think a good title for this is `Implement bulk 'given' queries`

c/src/given_rows.rs
```diff
@@ -0,0 +1,94 @@
...
+#[unsafe(no_mangle)]
+pub extern "C" fn given_rows_new(size_hint: usize) -> *mut QueryGivenRows {
+    release(QueryGivenRows(Vec::with_capacity(size_hint)))
+}
+
+/// Creates a new <code>givenRow</code> of the specified capacity,
+#[unsafe(no_mangle)]
+pub extern "C" fn given_rows_push(rows: *mut QueryGivenRows, row: *mut QueryGivenRow) {
+    borrow_mut(rows).0.push(take_ownership(row))
+}
```
flyingsilverfin: I see my a dict/map might be much harder to implement hmmmm can we have a think about if we can implement it? Maybe we can serialize across the C layer as a list of pairs (var name, concept) from the source language map and reassemble on the Rust side? I just think that API is much nicer than just a list of lists

c/src/given_rows.rs
```diff
@@ -0,0 +1,94 @@
...
+/// Creates a new <code>givenRow</code> of the specified width,
+/// initialised with <code>QueryGivenEntry::Empty</code>
+#[unsafe(no_mangle)]
+pub extern "C" fn given_row_new(width: usize) -> *mut QueryGivenRow {
+    let row = vec![QueryGivenEntry::Empty; width];
+    release(QueryGivenRow(row))
+}
+
+/// Sets the entry at `index` in the given row to the specified entity
+/// Will panic if out-of-bounds
```
flyingsilverfin: panic = exception on the caller? Can it set an error and return instead?

c/src/transaction.rs
```diff
@@ -63,6 +63,24 @@ pub extern "C" fn transaction_query(
...
+/// @param given_rows <code>QueryGivenRows</code> the rows given as input to the query.
+#[unsafe(no_mangle)]
+pub extern "C" fn transaction_query_given_rows(
+    transaction: *mut Transaction,
+    query: *const c_char,
+    options: *const QueryOptions,
+    given_rows: *mut QueryGivenRows,
+) -> *mut QueryAnswerPromise {
+    release(QueryAnswerPromise::new(Box::new(
+        borrow(transaction).query_with_options_and_inputs(string_view(query), *borrow(options), Some(take_ownership(given_rows))),
```
flyingsilverfin: `query_with_options_and_rows` might actually a be a nice API. They're the "given rows"

csharp/Api/ITransaction.cs
```diff
@@ -79,6 +80,15 @@ public interface ITransaction : IDisposable
...
         Promise<IQueryAnswer> Query(string query, QueryOptions options);
 
+        /// <summary>
+        /// Executes a TypeQL query with input rows in this transaction.
+        /// </summary>
+        /// <param name="query">The TypeQL query string to execute.</param>
+        /// <param name="options">Query options.</param>
+        /// <param name="givenRows">Input rows; each inner enumerable is one row, <c>null</c> entries represent empty variables.</param>
+        /// <returns>A promise that resolves to the query answer containing the results.</returns>
+        Promise<IQueryAnswer> Query(string query, QueryOptions options, List<List<IConcept?>> givenRows);
```
flyingsilverfin: Ok my idea continues to refine. It can be 'rows'... but our "row" type when we return a row from a query uses a associative list or something right? Has the idea of a 'header'? Maybe an idea we can borrow there? Although i think for users sending a dict/map or a list is the easiest...

csharp/Concept/ValueFactory.cs
flyingsilverfin: My instinct it could be directly on `ValueAPI`? A factory is ok too, but historically we have inlined our factories into static methods on the classes in Java for example

rust/src/transaction.rs
flyingsilverfin: Yeah in rust we can keep it Optional i spose! I'd call it `given_rows` instead of `inputs` as the var name.

csharp/Concept/ValueFactory.cs
krishnangovindraj: I don't think every language supports it being on the API, because you can't have abstract static methods.

python/typedb/connection/transaction.py
krishnangovindraj: Might be worth supporting both. No point of forcing users to go to dicts if they're happy to use rows.

c/src/transaction.rs
```diff
@@ -63,6 +63,24 @@ pub extern "C" fn transaction_query(
...
+/// @param transaction The <code>Transaction</code> to execute the query within.
+/// @param query The query string.
+/// @param options <code>QueryOptions</code> to configure the executed query.
+/// @param given_rows <code>QueryGivenRows</code> the rows given as input to the query.
+#[unsafe(no_mangle)]
+pub extern "C" fn transaction_query_given_rows(
+    transaction: *mut Transaction,
+    query: *const c_char,
+    options: *const QueryOptions,
+    given_rows: *mut GivenRows,
```
krishnangovindraj, line 79: I could just add this to `transaction_query` and allow it to be null, though that would be an API breaking change for the C driver.

c/typedb_driver.i
```diff
@@ -264,9 +282,19 @@ VoidPromise* transaction_on_close_register(const Transaction* transaction, Trans
...
 
+%newobject given_rows_header_builder_new;
+%delobject given_rows_header_builder_finish;
+%newobject given_rows_header_builder_finish;
+%newobject given_rows_builder_new;
+%delobject given_rows_builder_finish;
+%newobject given_rows_builder_finish;
+// GivenRows is consumed by any function that takes it (transaction_query_given_rows)
+%apply SWIGTYPE *DISOWN { GivenRows *given_rows };
+
```
krishnangovindraj: @dmitrii-ubskii , Do you know if this makes sense? Do you think I should manually test this?

csharp/Test/Behaviour/Query/QuerySteps.cs
```diff
@@ -235,6 +314,54 @@ public void GetAnswersOfTypeqlQueryFailsWithMessage(string expectedMessage, DocS
...
+        [Given(@"get answers of typeql read query with given rows; fails with a message containing: ""(.*)""")]
+        [When(@"get answers of typeql read query with given rows; fails with a message containing: ""(.*)""")]
+        [Then(@"get answers of typeql read query with given rows; fails with a message containing: ""(.*)""")]
+        public void GetAnswersOfTypeqlQueryWithGivenRowsFailsWithMessage(string expectedMessage, DocString query)
+        {
+            ClearAnswers();
+            var given = TakeGivenRows();
+            var exception = Assert.ThrowsAny<Exception>(() => { _queryAnswer = ExecuteQuery(query.Content, given); });
+            Assert.Contains(expectedMessage, exception.Message);
+        }
```
krishnangovindraj: TODO: If we've not had this before, I'm not sure why it's here now.

csharp/BUILD
```diff
@@ -48,6 +48,7 @@ csharp_library(
         "//csharp/Api:api",
         "//csharp/Common:common",
         "//csharp/Connection:connection",
+        "//csharp/Concept:concept",
```
krishnangovindraj, line 51: This is for the ConceptFactory. Are we ok with this?

csharp/ConceptFactory.cs
krishnangovindraj, line 1: New at the root level.

c/typedb_driver.i
krishnangovindraj: Ok disrergard for now. It doesn't work and I'll look into it:
`c/typedb_driver.i:292: Warning 453: Can't apply (SWIGTYPE *DISOWN). No typemaps are defined.`

csharp/ConceptFactory.cs
```diff
@@ -0,0 +1,189 @@
...
+            var converted = givenRows.Select(row =>
+                row.ToDictionary(
+                    kv => kv.Key,
+                    kv => kv.Value == null ? null :
+                          kv.Value is IConcept concept ? concept :
+                          (IConcept?)TryConvertToValue(kv.Value)
+                )
+            ).ToList();
+            return BuildGivenRowsFrom(converted);
+        }
```
krishnangovindraj: I'm not a huge fan of having this variant. imo it's not too much to expect the user to just call the `ConceptFactory.New` of the type they want.

csharp/Test/Behaviour/Query/QuerySteps.cs
```diff
@@ -983,8 +1074,6 @@ public void AnswerContainsDocument(DocString expectedDocument)
             CollectDocumentsAnswerIfNeeded();
             var expected = JSON.Parse(expectedDocument.Content);
             bool found = _collectedDocuments!.Any(doc => {
-                Console.WriteLine("Checking if this document is an expected document it could be:");
-                Console.WriteLine(doc);
```
krishnangovindraj, line 987: Cleanup

http-ts/tests/behaviour/steps/query.ts
```diff
@@ -149,8 +178,8 @@ Then(`answer get row\\({int}\\) get variable{is_by_var_index}\\({var}\\)${EXPECT
 Then('answer get row\\({int}\\) get variable{is_by_var_index}\\({var}\\) {is_or_not} empty', (rowIdx: number, indexed: boolean, variable: string, is: boolean) => {
     if (indexed) return; // http does not have indices
     if (answers.answerType === "ok" || answers.answerType === "conceptDocuments") assert.fail("Expected conceptRow answers");
-    if (is) assert.equal(answers.answers[rowIdx].data[variable], "");
-    else assert.notEqual(answers.answers[rowIdx].data[variable], "");
+    if (is) assert.equal(answers.answers[rowIdx].data[variable], null);
+    else assert.notEqual(answers.answers[rowIdx].data[variable], null);
```
krishnangovindraj, line 182: Was never hit till now.

java/BUILD
```diff
@@ -41,7 +41,9 @@ java_library(
         # Internal dependencies
         "//java/api",
         "//java/common",
+        "//java/concept",
         "//java/connection",
+        "//java:typedb_driver_jni",
```
krishnangovindraj, line 46: For ConceptFactory.

python/typedb/common/datetime.py
```diff
@@ -68,6 +68,9 @@ def __init__(self, timestamp_seconds: int, subsec_nanos: int, tz_name: Optional[
 
         self._nanos = subsec_nanos
 
+    def to_seconds_and_nanos(self) -> Tuple[int, int]:
+        return (int(self._datetime_of_seconds.timestamp()), self._nanos)
+
```
krishnangovindraj: Internal use for the FFI.

python/typedb/connection/transaction.py
```diff
@@ -72,11 +73,15 @@ def analyze(self, query: str) -> Promise[AnalyzedQuery]:
...
-    def query(self, query: str, options: Optional[QueryOptions] = None) -> Promise[QueryAnswer]:
+    def query(self, query: str, options: Optional[QueryOptions] = None, given_rows: Optional[GivenRows] = None) -> Promise[QueryAnswer]:
         require_non_null(query, "query")
         if not options:
             options = QueryOptions()
-        promise = transaction_query(self.native_object, query, options.native_object)
+        if given_rows is None:
+            promise = transaction_query(self.native_object, query, options.native_object)
+        else:
+            given_rows._native.thisown = 0
```
krishnangovindraj, line 88: We could do this in every language, though it's ideal if we get swig to do it.

python/typedb/concept_factory.py
```diff
@@ -0,0 +1,197 @@
...
+            {
+                var: (None if val is None else (val if isinstance(val, _ConceptApi) else ConceptFactory.try_convert_to_value(val)))
+                for var, val in row.items()
+            }
+            for row in rows
+        ]
+        return ConceptFactory.build_given_rows_from_map(converted)
+
+    @staticmethod
+    def try_convert_to_value(value: Any):
```
krishnangovindraj: Not a big fan of exposing this.

rust/src/common/error.rs
```diff
@@ -265,6 +265,18 @@ error_messages! { InternalError
...
+    GivenRowIndexOutOfBounds { index: usize, width: usize } =
+        1: "Attempted to set a row entry at index: {index}. Row width is {width}.",
+    GivenRowUnknownVariable { variable: String } =
+        2: "Attempted to set a row entry for a variable {variable} which was not in the header.",
+    GivenRowsSizeMismatch { actual: usize, expected: usize } =
+        3: "Cannot push a row of width: {actual} . Expected width: {expected}",
+    GivenRowsReceivedType { variable_or_index: String } =
+        4: "A type was passed as given row entry for '{variable_or_index}'. Only instances and values are allowed.",
+}
+
```
krishnangovindraj, line 283: Questionable naming this "QueryError" given it's mostly to do with the given rows input.

rust/tests/integration/values.rs
```diff
@@ -828,3 +830,76 @@ fn test_decimal_via_database() {
...
 }
+
+#[test]
+#[serial]
+fn test_round_trips() {
+    async_std::task::block_on(async {
+        let driver = setup().await;
+        let database = driver.databases().get(DATABASE_NAME).await.unwrap();
+
+        compile_error!("This is some terrible UX");
```
krishnangovindraj: Not great UX.

rust/tests/integration/values.rs
```diff
@@ -828,3 +830,76 @@ fn test_decimal_via_database() {
...
+        // Belfast is UTC+1 (BST) in September, so local 16:40:05 = UTC 15:40:05
+        let belfast_naive_utc = NaiveDate::from_ymd_opt(2024, 9, 20).unwrap().and_hms_opt(15, 40, 5).unwrap();
+        let belfast_dt = Value::DatetimeTZ(
+            TimeZone::IANA(Tz::from_str("Europe/Belfast").unwrap()).from_utc_datetime(&belfast_naive_utc),
+        );
```
krishnangovindraj: Especially this, if this is the only way to do it. Could be that Claude is being silly.

rust/src/given.rs
```diff
@@ -0,0 +1,135 @@
...
+    Value(Value),
+}
+
+/// Rows of data to be used as input to a TypeQL query using a `given` stage.
+///
+/// # Examples
+///
+/// ```rust
+/// let rows =
+/// transaction.query_with_options_and_rows(query, options, Some(rows))
```
krishnangovindraj: There needs to be atleast one example here.

krishnangovindraj (commented):
Self review pass

c/src/concept/concept.rs
```diff
@@ -424,6 +425,87 @@ pub extern "C" fn concept_to_string(concept: *const Concept) -> *mut c_char {
     release_string(format!("{:?}", borrow(concept)))
 }
 
+// Constructors
```
farost: not a fan of these separators, they will get mixed up with the other content in the future pretty easily

rust/tests/integration/values.rs
krishnangovindraj: I've used from_local_time since.

c/src/given_rows.rs
krishnangovindraj: Done

c/typedb_driver.i
krishnangovindraj: Ok, I removed this and called `release()` in the guest drivers instead, which seems to be convention for us.

MODULE.bazel
```diff
@@ -65,15 +65,15 @@ git_override(
 bazel_dep(name = "typedb_protocol", version = "0.0.0")
 git_override(
     module_name = "typedb_protocol",
-    remote = "https://github.com/typedb/typedb-protocol",
-    commit = "9a45ab625534a01afe5ecf89acaa06cecc92cafb",
+    remote = "https://github.com/krishnangovindraj/typedb-protocol",
+    commit = "bb9c513d4dcff269feb88015dcf99354ad378657",
```
dmitrii-ubskii: Note

MODULE.bazel
```diff
@@ -65,15 +65,15 @@ git_override(
...
 )
 
 bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
-    remote = "https://github.com/typedb/typedb-behaviour",
-    commit = "cd585f09c61381a31dc19c2e129cbf40d36acc3d",
+    remote = "https://github.com/krishnangovindraj/typedb-behaviour",
+    commit = "3031eed40edabadfddb0e080eaeaac3b99dfb77a",
```
dmitrii-ubskii: And here

c/src/concept/concept.rs
```diff
@@ -424,6 +426,97 @@ pub extern "C" fn concept_to_string(concept: *const Concept) -> *mut c_char {
...
+/// provided as a string (without the dec suffix).
+#[unsafe(no_mangle)]
+pub extern "C" fn concept_new_decimal_from_string(str: *const c_char) -> *mut Concept {
+    match Decimal::from_str(string_view(str)) {
+        Ok(value) => release(Concept::Value(Value::Decimal(value))),
+        Err(err) => {
+            record_error(err);
+            null_mut()
+        }
+    }
```
dmitrii-ubskii: This should have been `try_release(... .map(...))`

c/src/common/error.rs
```diff
@@ -79,7 +79,7 @@ pub(crate) fn unwrap_void(result: Result) {
     ok_record(result);
 }
 
-fn record_error(err: Error) {
+pub(crate) fn record_error(err: Error) {
     debug!("Encountered error {err} in typedb-driver-rust");
     LAST_ERROR.with(|prev| *prev.borrow_mut() = Some(err));
 }
```
dmitrii-ubskii, line 85: Keep this private and use `try_release` instead.

c/src/given_rows.rs
```diff
@@ -0,0 +1,205 @@
...
+#[unsafe(no_mangle)]
+pub extern "C" fn given_rows_builder_finish(builder: *mut GivenRowsBuilder) -> *mut GivenRows {
+    let builder = take_ownership(builder);
+    if builder.active_row.is_some() {
+        record_error(typedb_driver::error::Error::FFI(
+            "given_rows_builder_finish: There is an active, uncommitted row.".to_owned(),
+        ));
+        return null_mut();
+    };
+    release(builder.rows)
```
dmitrii-ubskii: This could have been `try_release(builder.finish())`

c/src/given_rows.rs
```diff
@@ -0,0 +1,205 @@
...
+    builder: *mut GivenRowsBuilder,
+    index: usize,
+    concept: *const Concept,
+) {
+    let result = borrow_mut(builder)
+        .active_row()
+        .and_then(|mut row| row.set_at(index, to_given_row_entry(borrow(concept), index)?));
+    if let Err(err) = result {
+        record_error(err);
+    }
```
dmitrii-ubskii: This should use `unwrap_void`.

flyingsilverfin (changes requested):
Lol my review never submitted -- oh it's empty weird

python/typedb/concept_factory.py
farost: Nah, everything is exposed in Python anyway

rust/src/common/error.rs
farost: It's fine. We used to have `QueryError`s, but I moved the whole logic to the server with, if possible, not pre-checks. You could consider removing the ones that are not very needed for simplification to let the server handle an incorrect input (because we still need to be protected from it for the HTTP endpoint). Then, if it suits `ConceptErro`s, it can be moved there, too. Otherwise, it's fine to have a new type, we may need to add something else. But do consider refactoring this if you feel the new category is excessive.

farost (commented):
comments

csharp/ConceptFactory.cs
dmitrii-ubskii: I feel like this should be split into the actual `ConceptFactory` or even `ValueFactory` that would deal with typedb native values (probably in `Concept`) and a module that would build the `GivenRows`.

csharp/Concept/GivenRows.cs
dmitrii-ubskii, line 1: Why is this in `Connection` and not `Api` with `QueryOptions`?

csharp/Connection/Transaction.cs
```diff
@@ -19,13 +19,12 @@
...
 using System;
 using System.Collections.Generic;
-
 using TypeDB.Driver;
 using TypeDB.Driver.Analyze;
 using TypeDB.Driver.Answer;
 using TypeDB.Driver.Api;
-using TypeDB.Driver.Api.Analyze;
 using TypeDB.Driver.Api.Answer;
+using TypeDB.Driver.Api.Analyze;
```
dmitrii-ubskii, line 26: Huh?

rust/src/analyze/mod.rs
```diff
@@ -36,10 +36,12 @@ pub mod pipeline;
 pub struct AnalyzedQuery {
     /// The original TypeQL query string
     pub source: String,
-    /// A representation of the query as a <code>Pipeline</code>
-    pub query: Pipeline,
     /// A representation of the <code>Function</code>s in the preamble of the query
     pub preamble: Vec<Function>,
+    /// A representation of the <code>Given</code> stage of the query, if it has one
+    pub given: Option<Given>,
```
dmitrii-ubskii, line 42: Side question: is there a set order between `given` and `with`? Does `given` come after all the `with` stages?

rust/src/common/error.rs
```diff
@@ -233,6 +233,8 @@ error_messages! { ConceptError
         1: "Cannot get concept from a concept row by variable '{variable}'.",
     UnavailableRowIndex { index: usize } =
         2: "Cannot get concept from a concept row by index '{index}'.",
+    ErrorParsingDecimal { unparsed: String, reason: ParseIntError } =
+        3: "Could not parse the value'{unparsed}' as decimal: {reason}.",
```
dmitrii-ubskii:
```suggestion
        3: "Could not parse the value '{unparsed}' as decimal: {reason}.",
```

rust/src/common/error.rs
dmitrii-ubskii: `Given` is a query stage, I think that's fine.

rust/src/common/error.rs
dmitrii-ubskii: Is this just so that we can construct one in the C layer and still haave it be of this type? That's a bit leaky but okay I guess.

rust/src/common/error.rs
```diff
@@ -335,7 +353,9 @@ impl Error {
             Self::Concept(error) => error.message(),
             Self::Migration(error) => error.message(),
             Self::Internal(error) => error.message(),
+            Self::Query(error) => error.message(),
             Self::Server(error) => error.message(),
+            Self::FFI(_error) => String::new(),
```
dmitrii-ubskii: Surely it should be this?
```suggestion
            Self::FFI(error) => error.clone(),
```

rust/src/concept/value.rs
```diff
@@ -284,7 +287,35 @@ impl Sub for Decimal {
...
+
+        let (integer_part, fractional_part) = str.split_once(".").unwrap_or((str, "0"));
+        let integer = integer_part
+            .parse()
+            .map_err(|reason| Error::Concept(ConceptError::ErrorParsingDecimal { unparsed: str.to_owned(), reason }))?;
+        let num_fractional_digits = fractional_part.len() as u32;
+        let parsed_fractional = fractional_part
+            .parse::<u64>()
+            .map_err(|reason| Error::Concept(ConceptError::ErrorParsingDecimal { unparsed: str.to_owned(), reason }))?;
+        let fractional = parsed_fractional * 10u64.pow(Self::FRACTIONAL_PART_DENOMINATOR_LOG10 - num_fractional_digits);
```
dmitrii-ubskii, line 383: There are two missing checks here that is also missing server-side:
- `parsed_fractional` may be `>= FRACTIONAL_PART_DENOMINATOR` while still fitting in `u64`;
- `fractional` may overflow.
Both can be avoided if we just check that the length of the fractional part is at most `FRACTIONAL_PART_DENOMINATOR_LOG10`.

rust/src/concept/value.rs
```diff
@@ -284,7 +287,35 @@ impl Sub for Decimal {
...
+    fn from_str(mut str: &str) -> Result<Self, Self::Err> {
+        if str.ends_with("dec") {
+            str = str.trim_end_matches("dec");
+        }
+        let is_negative = if str.starts_with("-") {
+            str = str.trim_start_matches('-');
+            true
+        } else {
+            false
+        };
```
dmitrii-ubskii: I'm also now realizing that this will accept `----2.2` and result in a value of `-2.2`.

dmitrii-ubskii: On the server side we are at least relying on the fact that the literal was parsed by typeql grammar so that shouldn't be an issue there. Still worth fixing.

rust/src/connection/network/proto/concept.rs
```diff
@@ -344,3 +345,76 @@ impl TryFromProto<i32> for Kind {
...
+        ValueProto { value: Some(value) }
+    }
+}
+
+impl IntoProto<EntityProto> for Entity {
+    fn into_proto(self) -> EntityProto {
+        EntityProto { iid: self.iid.bytes_cloned(), entity_type: None }
+    }
+}
+impl IntoProto<RelationProto> for Relation {
```
dmitrii-ubskii, line 374:
```suggestion
}

impl IntoProto<RelationProto> for Relation {
```

rust/src/connection/network/proto/message.rs
```diff
@@ -228,13 +229,56 @@ impl IntoProto<transaction::Req> for TransactionRequest {
...
+        let entries = self.into_iter().map(|entry| entry.into_proto()).collect();
+        typedb_protocol::query::req::GivenRow { entries }
+    }
+}
+
+impl IntoProto<typedb_protocol::query::req::GivenEntry> for GivenRowEntry {
+    fn into_proto(self) -> typedb_protocol::query::req::GivenEntry {
+        use typedb_protocol::{
+            Thing as ThingProto, query::req::given_entry::Entry as EntryProto, thing::Thing as ThingProtoInner,
+        };
```
dmitrii-ubskii, line 260: Why here and not top level?

rust/src/given.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub fn push_row(&mut self, row: Vec<GivenRowEntry>) -> Result<()> {
+        if row.len() == self.header.width() {
+            self.rows.push(row);
+            Ok(())
+        } else {
+            Err(Error::Query(QueryError::GivenRowsSizeMismatch { actual: row.len(), expected: self.header.width() }))
+        }
+    }
+
+    pub fn push_map(&mut self, values: impl IntoIterator<Item = (String, GivenRowEntry)>) -> Result<()> {
```
dmitrii-ubskii: Neat.

dmitrii-ubskii: Also I believe we have `Result<()>` <=> `Result`?

dmitrii-ubskii (changes requested):
Only looked at C, Rust, and C# so far.

rust/src/analyze/mod.rs
krishnangovindraj: Given is an optional initial stage of the pipeline, so it comes just before the main pipeline.

rust/src/connection/network/proto/message.rs
krishnangovindraj: I thought we had it as a convention to not import everything and name them `Proto`, but I guess not.
and that mainly because you then even have ThingProto and ThingProtoInner.

rust/src/common/error.rs
krishnangovindraj: Good point. It's a bit leaky but a lot of the GivenRows API is that way because of the FFI anyway

java/concept/GivenRowsImpl.java
flyingsilverfin, line 1: /connection package is the wrong place for this no?

java/connection/TransactionImpl.java
```diff
@@ -99,6 +102,21 @@ public Promise<? extends QueryAnswer> query(String query, QueryOptions options)
         }
     }
 
+    @Override
+    public Promise<? extends QueryAnswer> query(String query, GivenRows givenRows) throws TypeDBDriverException {
+        return query(query, new QueryOptions(), givenRows);
+    }
+
+    @Override
+    public Promise<? extends QueryAnswer> query(String query, QueryOptions options, GivenRows givenRows) throws TypeDBDriverException {
```
flyingsilverfin, line 133: Let's add an overload for the List of maps that we convert like in Python what do you think?

python/tests/integration/test_values.py
```diff
@@ -493,5 +493,89 @@ def test_duration(self):
                 assert_that(Duration.fromstring("P999Y12M31DT24H59M59.999999999S"), is_(typedb_duration))
 
 
+    def test_roundtrips_with_raw_value(self):
```
flyingsilverfin, line 496: This test name (and similar ones in other lang's) isn't indicating that this is about `given`. Perhaps `test_given_roundtrip_with_raw_value`?

python/typedb/api/connection/transaction.py
```diff
@@ -84,12 +85,13 @@ def options(self) -> TransactionOptions:
...
 
     @abstractmethod
-    def query(self, query: str, options: Optional[QueryOptions] = None) -> Promise[QueryAnswer]:
+    def query(self, query: str, options: Optional[QueryOptions] = None, given_rows=None) -> Promise[QueryAnswer]:
         """
         Execute a TypeQL query in this transaction.
 
         :param query: The query to execute.
         :param options: The ``QueryOptions`` to execute the query with..
+        :param given_rows: Rows given to the query as input. Will throw an exception if a type is passed.
```
flyingsilverfin: Don't forget to update the comment if changing the API

python/typedb/api/concept/given_rows.py
flyingsilverfin, line 1: Wrong package

flyingsilverfin (changes requested):

python/tests/integration/test_values.py
krishnangovindraj: This is more to test the value constructors than the given stage.

java/concept/GivenRowsImpl.java
krishnangovindraj: True, but it's unclear which package it should be in. It's most like answer/

cpp/docs_structure.bzl
```diff
@@ -61,6 +61,7 @@ dir_mapping = {
     "Attribute.adoc": "data",
     "Value.adoc": "data",
     "Thing.adoc": "data",
+    "ConceptFactory.adoc": "data",
```
flyingsilverfin: This file might need updating

flyingsilverfin (approved):

dmitrii-ubskii (approved):

PR #877, opened 2026-05-21
Use connection address for standalone connections

farost:
## Usage and product changes

Make the driver use the connection address in case the server is not a part of a cluster and does not advertise an address for connection. 

This feature is required to allow easy connections to servers behind a proxy or inside a container -- otherwise, they'd have to provide advertise addresses, which makes the initial TypeDB set up more complicated (requires https://github.com/typedb/typedb/pull/7817).

For clustered connections, all clustered servers are still expected to provide the address through the protocol. 

## Implementation

Do not change the network layer -- just add a special case for non-clustered (no replication info) server without an advertise address.

HTTP-TS does not use the returned address in its connection, so it requires no changes.

### DIFF
--- a/rust/src/connection/server/server_manager.rs
+++ b/rust/src/connection/server/server_manager.rs
@@ -393,7 +393,8 @@ impl ServerManager {
             match replica_connection {
                 Ok((replica_connection, replicas)) => {
                     debug!("Fetched replicas from configured address '{address}': {replicas:?}");
-                    let translated_replicas = Self::translate_replicas(replicas, &address_translation);
+                    let translated_replicas =
+                        Self::translate_replicas(replicas, &address_translation, &replica_connection);
                     let mut source_connections = HashMap::with_capacity(translated_replicas.len());
                     source_connections.insert(address.clone(), replica_connection);
                     return Ok((source_connections, translated_replicas));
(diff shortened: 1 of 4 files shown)


rust/src/connection/server/server_manager.rs
```diff
@@ -438,14 +439,28 @@ impl ServerManager {
...
         address_translation: &AddressTranslation,
+        source: &ServerConnection,
     ) -> HashSet<Server> {
-        replicas.into_iter().map(|replica| replica.translated(address_translation)).collect()
+        replicas
+            .into_iter()
+            .map(|replica| match replica {
+                // standalone server without advertised address: use connection address directly
+                Server::Unavailable { replication_status: None } => {
+                    Server::available_from_private(source.address().clone(), None)
```
flyingsilverfin, line 458: is that the right name `from_private`? perhaps it's `available_from_initial` or `from_external`?

flyingsilverfin (approved):
Approved w 1 idea

rust/src/connection/server/server_manager.rs
farost: It uses the notion of public/private addresses for address translation. Not a new method

PR #886, opened 2026-06-08
Enhance redirection logic in grpc drivers

farost:
## Usage and product changes

Fixes a failover hole: when a clustered driver's cached primary is alive but no longer primary, the server's `CSV9` redirect ("not primary, here's who is") was bubbled up to the caller as a generic `Error::Server` with no retry.

The driver now uses the correct error codes and routes the retry directly to the hinted node when it's presented, falling back to the existing seek-primary path otherwise.

## Implementation
- `rust/src/common/error.rs`: collapse error messages to only currently used variants and introduce the hinted non-primary error option.
- `rust/src/connection/server/server_manager.rs`: `execute_strongly_consistent` short-circuits to the hinted private address when known; otherwise falls through to seek. Hint short-circuit consumes one retry slot, so `primary_failover_retries` is still the hard upper bound on total attempts.
- `rust/src/connection/driver_options.rs` + Python / Java / C# / C source + adoc partials: identical doc text everywhere.
- `rust/tests/integration/cluster/failover.rs`: a new redirect test. Single-address connect, kill primary, wait for re-election, restart killed node as secondary, then issue a read without refreshing the driver's view first (exactly the case we hit).


csharp/Api/Server/ServerVersion.cs
```diff
@@ -31,10 +31,7 @@ namespace TypeDB.Driver.Api
     /// </example>
     public class ServerVersion : NativeObjectWrapper<Pinvoke.ServerVersion>
     {
-        /**
-         * @private
```
farost, line 35: Turns out this is not recognized by the C# doc tool, it gives a warning we used to ignore

c/src/driver_options.rs
```diff
@@ -50,10 +50,11 @@ pub extern "C" fn driver_options_get_tls_config(options: *const DriverOptions) -
     release(borrow(options).tls_config.clone())
 }
 
-/// Sets the number of times the driver retries finding and re-routing to the primary replica
-/// on connection failures. This value is used both for polling during leader election (up to
-/// N+1 attempts with a 2-second sleep between each) and for re-executing a failed request on
-/// a newly discovered primary. Defaults to 1.
+/// Specifies the number of retries the driver performs to find and reach the cluster primary
```
farost, line 53: Update docs for this option everywhere

docs/modules/ROOT/partials/rust/errors/ConnectionError.adoc
```diff
@@ -25,7 +25,6 @@ a| `ListsNotImplemented`
 a| `MissingPort`
 a| `MissingResponseField`
 a| `NoPrimaryServer`
-a| `NotPrimaryOnReadOnly`
```
farost, line 28: Not used now and we'll probably just return the server error in this case, it will contain all the info. No need for a redeclaration

rust/src/common/error.rs
```diff
@@ -152,8 +152,8 @@ error_messages! { ConnectionError
         12: "Didn't receive any server responses for the query.",
     UnexpectedQueryType { query_type: i32 } =
         13: "Unexpected query type in message received from server: {query_type}. This is either a version compatibility issue or a bug.",
-    ClusterServerNotPrimary =
-        14: "The server is not primary.",
+    ClusterServerNotPrimary { hint_address: Option<Address> } =
+        14: "The server is not primary (primary hint: {hint_address:?}).",
```
farost, line 156: I really want to collapse it to single error, I guess printing `Option` is fine here?

rust/src/common/error.rs
```diff
@@ -196,8 +196,6 @@ error_messages! { ConnectionError
         34: "Server is not yet initialized.",
     AnalyzeNoResponse =
         35: "Didn't receive any server responses for the analyze request.",
-    NotPrimaryOnReadOnly { address: Address } =
```
farost, line 199: Not used

rust/src/common/error.rs
```diff
@@ -342,11 +340,14 @@ impl Error {
...
-    fn try_extracting_connection_error_code(code: &str) -> Option<ConnectionError> {
-        match code {
+    fn try_extracting_connection_error_from_info(error_info: &ErrorInfo) -> Option<ConnectionError> {
+        match error_info.reason.as_str() {
             "AUT1" | "AUT2" | "AUT3" => Some(ConnectionError::TokenCredentialInvalid {}),
             "SRV14" | "RFT1" | "CSV7" => Some(ConnectionError::ServerIsNotInitialised {}),
-            "CSV8" => Some(ConnectionError::ClusterServerNotPrimary {}),
+            "CSV8" => Some(ConnectionError::ClusterServerNotPrimary { hint_address: None }),
+            "CSV9" => Some(ConnectionError::ClusterServerNotPrimary {
+                hint_address: error_info.metadata.get("redirect_address").and_then(|addr| addr.parse::<Address>().ok()),
```
farost, line 349: New server's format from https://github.com/typedb/typedb/pull/7841

csharp/Api/Server/ServerVersion.cs
farost: Instead, make this `internal` and make `internal` accessible by the `Connection` package (see below)

csharp/Api/AssemblyInfo.cs
```diff
@@ -0,0 +1,22 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("TypeDB.Driver.Connection")]
```
farost, line 22: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-10.0.   

that's life

rust/tests/integration/cluster/failover.rs
```diff
@@ -240,3 +240,57 @@ fn primary_failover() {
         cleanup_database().await;
     });
 }
+
+#[test]
+#[serial]
+fn primary_failover_redirect_hint() {
```
farost, line 246: Essentially, there is a need for specific failover tests (since this is a grpc-specific error, http reconnections are covered by the cluster itself)

farost:
The builds will get green once we merge and propagate this 2dbbe636af503bedb525712a8c315e9b364d4f66

c/src/driver_options.rs
```diff
@@ -62,11 +63,11 @@ pub extern "C" fn driver_options_set_primary_failover_retries(
     borrow_mut(options).primary_failover_retries = primary_failover_retries as usize;
 }
 
-/// Returns the value set for the primary failover retries limit in this <code>DriverOptions</code> object.
-/// Sets the number of times the driver retries finding and re-routing to the primary replica
-/// on connection failures. This value is used both for polling during leader election (up to
-/// N+1 attempts with a 2-second sleep between each) and for re-executing a failed request on
-/// a newly discovered primary.
+/// Specifies the number of retries the driver performs to find and reach the cluster primary
```
flyingsilverfin, line 67: This one isn't specifying but is 'getting' right?

c/src/driver_options.rs
```diff
@@ -62,11 +63,11 @@ pub extern "C" fn driver_options_set_primary_failover_retries(
...
+/// after a failed request, before giving up. Total attempts per user request = <code>N + 1</code>.
+/// Each retry either follows the server's redirect address (fast path) or polls the
+/// known replicas with a 2-second sleep between polls (slow path).
+/// Set to 0 to disable failover. Defaults to 1.
```
flyingsilverfin, line 71: the old comment looks more right?

csharp/Api/AssemblyInfo.cs
flyingsilverfin: dayum

docs/modules/ROOT/partials/c/connection/driveroptions.adoc
```diff
@@ -38,7 +38,7 @@ int64_t driver_options_get_primary_failover_retries(const struct DriverOptions*
 
 
 
-Returns the value set for the primary failover retries limit in this ``DriverOptions`` object. Sets the number of times the driver retries finding and re-routing to the primary replica on connection failures. This value is used both for polling during leader election (up to N+1 attempts with a 2-second sleep between each) and for re-executing a failed request on a newly discovered primary.
+Specifies the number of retries the driver performs to find and reach the cluster primary after a failed request, before giving up. Total attempts per user request = ``N + 1``. Each retry either follows the server's redirect address (fast path) or polls the known replicas with a 2-second sleep between polls (slow path). Set to 0 to disable failover. Defaults to 1.
```
flyingsilverfin: same here

java/api/DriverOptions.java
```diff
@@ -115,11 +115,11 @@ public DriverOptions requestTimeoutMillis(long requestTimeoutMillis) {
     }
 
     /**
-     * Returns the value set for the primary failover retries limit in this <code>DriverOptions</code> object.
-     * Sets the number of times the driver retries finding and re-routing to the primary server
-     * on connection failures. This value is used both for polling during leader election (up to
-     * N+1 attempts with a 2-second sleep between each) and for re-executing a failed request on
-     * a newly discovered primary.
+     * Specifies the number of retries the driver performs to find and reach the cluster primary
```
flyingsilverfin, line 119: eh

python/typedb/api/connection/driver_options.py
```diff
@@ -124,11 +124,11 @@ def request_timeout_millis(self, request_timeout_millis: int):
     @property
     def primary_failover_retries(self) -> int:
         """
-        Returns the value set for the primary failover retries limit in this ``DriverOptions`` object.
-        Sets the number of times the driver retries finding and re-routing to the primary server
-        on connection failures. This value is used both for polling during leader election (up to
-        N+1 attempts with a 2-second sleep between each) and for re-executing a failed request on
-        a newly discovered primary. Defaults to 1.
+        Specifies the number of retries the driver performs to find and reach the cluster primary
```
flyingsilverfin, line 128: here

c/src/driver_options.rs
farost: It's a general description for the field, both for setters and getters. Supporting different forms is a nightmare in all languages, I just go the easy way.

rust/src/common/error.rs
flyingsilverfin: yeah can do!

c/src/driver_options.rs
farost: Okay, I'll fix it to `Returns`, it really was it =(

flyingsilverfin (approved):
Approved w comments

c/src/driver_options.rs
farost: Are you talking about the description or the fact that it does not use `Returns` anymore?

PR #889, opened 2026-06-08
Use Bazel `exclusive` tag for tests instead of CLI --jobs=1

flyingsilverfin:
## Usage and product changes
Replace the CI Bazel `--jobs=1` flag to ensure jobs aren't running in parallel and clashing for resources with the `exclusive` bazel tag.

Note that the `exclusive` flag allows parallel builds but blocks parallel test phase execution. In addition, if we ever add it, it prevents remote execution.

## Implementation
All integration and behaviour tests run against a single shared TypeDB server (started out-of-band by tool/test/start-core-server.sh / start-cluster-servers.sh), so two test targets cannot run concurrently. CI worked around this by passing --jobs=1 on every test command, which serializes the *entire* Bazel invocation.

Tag the test targets with `exclusive` instead, which serializes them against each other at the Bazel level while letting the build (and other, non-exclusive tests) parallelize. --jobs=1 is then removed from every test command in .factory/automation.yml (the //csharp/... `bazel build --jobs=1` is left as-is: it works around a C# build mutex, not test concurrency).

Behaviour tests are tagged via their shared macros (both -core and -cluster variants are covered):
  - rust/tests/behaviour/rules.bzl
  - java/test/behaviour/rules.bzl
  - python/tests/behaviour/behave_rule.bzl
  - csharp/Test/rules.bzl (csharp_behaviour_test)
  - http-ts/tests/behaviour/rules.bzl Integration tests are tagged per-target (java, python, http-ts) or via the csharp_integration_test macro (csharp); the cluster failover targets keep their existing external/no-cache/no-sandbox tags.

### DIFF
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -126,7 +126,7 @@ build:
         bazel run --config=ci @typedb_dependencies//distribution/artifact:create-netrc
 
         tool/test/start-core-server.sh &&
-          bazel test --config=ci //rust/tests/behaviour/... --test_output=streamed --jobs=1 &&
+          bazel test --config=ci //rust/tests/behaviour/... --test_output=streamed &&
           export TEST_SUCCESS=0 || export TEST_SUCCESS=1
         tool/test/stop-core-server.sh
         exit $TEST_SUCCESS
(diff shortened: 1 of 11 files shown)


flyingsilverfin:
Note: not sure why but CI doesn't run on this repo for me - will run after merging and then ill fix it if anything breaks

farost:
> not sure why but CI doesn't run on this repo for me

lol

farost (approved):
Quite easy to fix if the CI breaks after the merge

PR #891, opened 2026-06-18
Fix transaction on_close callback panic

flyingsilverfin:
## Usage and product changes

A race between on_close and transaction teardown could cause the driver to panic and crash the user's application. We fix this by assuming an error from the register/acknowledgement channel means 'transaction already closed' instead of panicking. 

## Implementation

Don't abort the process when on_close races transaction teardown. A transaction close racing the teardown of its gRPC stream could abort the host process. Transaction::close() registers an on_close notifier and blocks on an acknowledgement from the transaction's background listener; if that listener had already stopped (because the stream was torn down — e.g. the server killing the transaction when its timeout expires), the acknowledgement never arrived and on_close() panicked via `expect`. The panic cannot unwind across the FFI boundary, so the whole host process (e.g. a JVM using the Java driver) aborted with SIGABRT instead of close() returning normally.

on_close (sync + async) now treats a closed registration/acknowledgement channel as "the transaction is already closed" and returns Ok. The callback is dropped un-invoked, matching close()/close_with_error(), which only run callbacks recorded before teardown. The two sibling unwrap()/expect() calls on the same teardown path — the listener's acknowledgement (recorded_signal.send) and close()'s notifier (closed_sink.send) — are likewise made best-effort, so a peer that has legitimately gone away (e.g. a cancelled close()) no longer panics.

Added an integration regression test that registers on_close and closes again after the transaction has already closed; it aborted the process before this change and passes after it.

### DIFF
--- a/rust/src/connection/network/transmitter/transaction.rs
+++ b/rust/src/connection/network/transmitter/transaction.rs
@@ -168,7 +168,6 @@ impl TransactionTransmitter {
         box_promise(async move {
             let (sender, mut sink) = unbounded_async();
             self.on_close_register_sink.send((Box::new(callback), sender)).ok();
-            // A closed channel means the listener already stopped: the transaction is closed.
             let _ = sink.recv().await;
             Ok(())
         })
@@ -182,7 +181,6 @@ impl TransactionTransmitter {
         box_promise(move || {
             let (sender, mut sink) = unbounded_async();
             self.on_close_register_sink.send((Box::new(callback), sender)).ok();
-            // A closed channel means the listener already stopped: the transaction is closed.
             let _ = sink.blocking_recv();
             Ok(())
         })
--- a/rust/src/connection/network/transmitter/transaction.rs
+++ b/rust/src/connection/network/transmitter/transaction.rs
@@ -119,7 +119,7 @@ impl TransactionTransmitter {
             if self.background_runtime.is_open() && self.is_open.compare_exchange(true, false).is_ok() {
                 let (closed_sink, mut closed_source) = unbounded_async();
                 let close_notifier_callback = Box::new(move |_error| {
-                    closed_sink.send(()).unwrap();
+                    closed_sink.send(()).ok();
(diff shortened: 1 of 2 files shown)


rust/src/connection/network/transmitter/transaction.rs
```diff
@@ -168,7 +168,8 @@ impl TransactionTransmitter {
         box_promise(async move {
             let (sender, mut sink) = unbounded_async();
             self.on_close_register_sink.send((Box::new(callback), sender)).ok();
-            sink.recv().await.expect("Did not receive on_close registration success signal");
+            // A closed channel means the listener already stopped: the transaction is closed.
+            let _ = sink.recv().await;
```
dmitrii-ubskii, line 171: Let's not mix `let _ = ` and `.ok()`
```suggestion
            sink.recv().await.ok();
```
Also for future reference `let` is not necessary, `_ = result;` also works.

rust/src/connection/network/transmitter/transaction.rs
```diff
@@ -119,7 +119,7 @@ impl TransactionTransmitter {
             if self.background_runtime.is_open() && self.is_open.compare_exchange(true, false).is_ok() {
                 let (closed_sink, mut closed_source) = unbounded_async();
                 let close_notifier_callback = Box::new(move |_error| {
-                    closed_sink.send(()).unwrap();
+                    closed_sink.send(()).ok();
                 });
```
dmitrii-ubskii, line 123: This (and a related one for the `sync` version) don't seem to be doing anything currently. It's probably still worth keeping so that any unresolved / cancelled promises don't cause a crash in the future.

dmitrii-ubskii (approved):

rust/src/connection/network/transmitter/transaction.rs
flyingsilverfin: ya ok let's keep it then anyways

PR #892, opened 2026-06-18
Fix string marshalling over JNI

### LINKED ISSUE #699: Java driver doesn't process strings with emojis correctly
## Description
Java driver doesn't process strings with emojis (e.g. 😎) correctly. Instead, it fails on the Rust side's bytes parsing.

## Environment

1. TypeDB distribution: Core
2. TypeDB version: 3.0.0-alpha-6 and earlier
3. Environment: Mac

Use a bdd test (available in our BDD repo as `cannot create database with an emoji` in `connection/database`):
``` 
  Background:
    Given typedb starts
    Given connection opens with default authentication
    Given connection is open: true
    Given connection has 0 databases

  Scenario: cannot create database with an incorrect name
    Then connection create database: 😎; fails
```

The database name's parsing will fail with an error:
```
thread '<unnamed>' panicked at c/src/memory.rs:109:13:
called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 0, error_len: Some(1) }
```

If we print the received bytes on the Rust side, it will show `eda0bdedb88e`. In the meantime, we'd expect `f09f988e` as the UTF-8 representation for this emoji. 

Characters inside the Basic Multilingual Plane (like Chinese chars) are processed correctly. The issue seems to be exclusive to chars outside of BMP.

We'll need to modify strings processing in SWIG for Java (and probably other languages... at least Python works correctly, others will be tested later).

dmitrii-ubskii:
## Usage and product changes

We forcibly re-encode Java strings from MUTF-8 (modified UTF-8 that uses surrogate characters for code points starting at U+10000) to UTF-8 before passing them over JNI.

## Implementation

Resolves #699

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -73,7 +73,7 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/typedb/typedb-behaviour",
-    commit = "1fc96034c69924aebdd40648429508ead1cf2f3f",
+    commit = "cad3c7b5b0475b3fe9bba81c0c425df6dcdf5b6a",
 )
 
 # Rust rules need typedb_dependencies for patching
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -73,7 +73,7 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/typedb/typedb-behaviour",
-    commit = "cad3c7b5b0475b3fe9bba81c0c425df6dcdf5b6a",
+    commit = "0a0e2dee11cef1f8e15a97e1428fb9d0c79c4225",
 )
 
 # Rust rules need typedb_dependencies for patching
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -73,7 +73,7 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/typedb/typedb-behaviour",
-    commit = "86d85293478196eac21ef4bc46b507f9d6918230",
+    commit = "ba391df36199957a39c02bb47aeda4c5e9410959",
 )
 
 # Rust rules need typedb_dependencies for patching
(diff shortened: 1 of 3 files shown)


c/src/common/memory.rs
```diff
@@ -96,7 +96,10 @@ pub(crate) fn free<T>(raw: *mut T) {
 
 pub(crate) fn string_view(str: *const c_char) -> &'static str {
     assert!(!str.is_null());
-    unsafe { CStr::from_ptr(str).to_str().unwrap() }
+    unsafe {
+        let cstr = CStr::from_ptr(str);
+        cstr.to_str().unwrap_or_else(|e| panic!("Could not parse {cstr:?}: {e}"))
+    }
```
dmitrii-ubskii, line 102: Improved panic message so we can at least see what went wrong.

Possibly worth making this return a `Result` that would become a handleable exception on the client language side? The side effect would be that every function acceptinga string would become fallible.

Either way, out of scope for this PR.

MODULE.bazel
```diff
@@ -73,7 +73,7 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/typedb/typedb-behaviour",
-    commit = "cd585f09c61381a31dc19c2e129cbf40d36acc3d",
+    commit = "0a0e2dee11cef1f8e15a97e1428fb9d0c79c4225",
```
dmitrii-ubskii: https://github.com/typedb/typedb-behaviour/pull/423

c/swig/typedb_driver_java.swg
```diff
@@ -46,7 +46,11 @@
     protected transient boolean swigCMemOwn;
 
     protected $javaclassname(long cPtr, boolean cMemoryOwn) {
-        super((typedb_driverJNI.error_code(cPtr, null) + " " + typedb_driverJNI.error_message(cPtr, null)).strip());
+        super((
+            new String(typedb_driverJNI.error_code(cPtr, null), java.nio.charset.Charset.forName("UTF-8"))
+            + " "
+            + new String(typedb_driverJNI.error_message(cPtr, null), java.nio.charset.Charset.forName("UTF-8"))
+        ).strip());
```
dmitrii-ubskii, line 53: I'm not sure if there's a way to reuse the string `javain` typemap here.

farost (approved):

PR #898, opened 2026-07-03
Update the test script for the latest cluster setup

farost:
## Usage and product changes
Upgrade the test infrastructure to support the flag-driver initialization of TypeDB Cluster.

## Implementation
Update server and BDD references to the latest commits.
Include the new `--server.clustering.init=true` flag to allow new clusters initialization when needed for the first test node.

### DIFF
--- a/tool/test/cluster-server.sh
+++ b/tool/test/cluster-server.sh
@@ -139,11 +139,6 @@ server_start() {
     init_arg="--server.clustering.init=true"
   fi
 
-  # Restrict umask so the admin unix socket is created with mode 0600 (world-r/w
-  # sockets are rejected by the admin tool with [ADM9]). The subshell isolates
-  # the change from anything else in the caller's environment.
-  umask 0077
-
   # Use setsid on Linux to start in a new session (prevents process group cleanup).
   # On macOS, nohup + disown is sufficient.
   local setsid_cmd=""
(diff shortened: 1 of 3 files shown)


tool/test/cluster-server.sh
```diff
@@ -131,6 +134,11 @@ server_start() {
   # by the outside-sandbox cluster start and the unlink fails with EPERM, even
   # though typedb itself can still truncate-and-write via the existing path.
 
+  local init_arg=""
+  if [ "${INIT}" = "true" ]; then
+    init_arg="--server.clustering.init=true"
+  fi
```
dmitrii-ubskii: Can it be `init_arg="--server.clustering.init=$INIT"`? That way we also get the server to complain about any invalid values.

tool/test/start-cluster-servers.sh
```diff
@@ -56,7 +62,11 @@ done
 
 echo Starting a cluster consisting of $NODE_COUNT servers...
 for i in $(seq 1 $NODE_COUNT); do
-  "${CLUSTER_SERVER}" start $i
+  if [ "$i" -eq 1 ]; then
+    INIT=true "${CLUSTER_SERVER}" start $i
+  else
+    INIT=false "${CLUSTER_SERVER}" start $i
+  fi
```
dmitrii-ubskii, line 69: This seems backwards.

dmitrii-ubskii (approved):

tool/test/start-cluster-servers.sh
farost: If 1, then INIT is true. It's correct

tool/test/cluster-server.sh
farost: Good idea!

PR #905, opened 2026-07-21
Fix C integration tests reporting success on failed assertions

flyingsilverfin:
## Usage and product changes
Fix incorrect successful test reporting in C integration tests.

## Implementation
The query tests reused `success` as scratch storage for the define-query result, so once the define succeeded every later failure path returned true. TestQuery.TestData was green on CI while printing "Did not find inserted name 'John' in query result."

`success` is now assigned exactly twice per test: false at declaration and true on the final line.

Fixing the latch exposed a real failure in test_query_data: `name` was defined without  `@independent`, so the standalone attribute it inserts is not retained and the match found nothing. test_query_given already got this right.

### DIFF
--- a/c/src/given_rows.rs
+++ b/c/src/given_rows.rs
@@ -97,7 +97,10 @@ pub extern "C" fn given_rows_header_builder_finish(builder: *mut GivenRowsHeader
 
 /// Creates a new <code>givenRow</code> of the specified capacity,
 #[unsafe(no_mangle)]
-pub extern "C" fn given_rows_builder_new(header: *mut GivenRowsHeader, row_count_hint: usize) -> *mut GivenRowsBuilder {
+pub extern "C" fn given_rows_builder_new(
+    header: *const GivenRowsHeader,
+    row_count_hint: usize,
+) -> *mut GivenRowsBuilder {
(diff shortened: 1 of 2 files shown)


c/src/given_rows.rs
```diff
@@ -97,7 +97,10 @@ pub extern "C" fn given_rows_header_builder_finish(builder: *mut GivenRowsHeader
 
 /// Creates a new <code>givenRow</code> of the specified capacity,
 #[unsafe(no_mangle)]
-pub extern "C" fn given_rows_builder_new(header: *mut GivenRowsHeader, row_count_hint: usize) -> *mut GivenRowsBuilder {
+pub extern "C" fn given_rows_builder_new(
+    header: *const GivenRowsHeader,
```
flyingsilverfin, line 101: @dmitrii-ubskii  not sure about this can you verify? Claude did it as a drive-by i can just undo these mut -> const changes if you think not good. Im not up to date on what the mean/do

dmitrii-ubskii: It's pointing into an `Arc`, `*const` is more correct.

farost (approved):

PR #910, opened 2026-07-29
Fix remote cache setup

krishnangovindraj:
## Implementation
Moves python installation into sub-shell and does pushd/popd.
Removes from windows because it doesn't look like we ever do it in windows.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,15 +67,16 @@ commands:
           amazon-linux-extras install java-openjdk11 -y
           yum install wget make gcc gcc-c++ openssl-devel bzip2-devel libffi-devel zlib-devel file lsof which procps tar git -y
           
-          cd /tmp
-          wget https://www.python.org/ftp/python/3.9.6/Python-3.9.6.tgz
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -67,15 +67,16 @@ commands:
...
+          (pushd /tmp
+            wget https://www.python.org/ftp/python/3.9.6/Python-3.9.6.tgz
+            tar -xvf Python-3.9.6.tgz
+            cd Python-3.9.6
+            ./configure --enable-optimizations --enable-shared
+            make altinstall
+            echo "/usr/local/lib" >> /etc/ld.so.conf.d/python3.9.conf
+            ldconfig
+            ln -s /usr/local/bin/python3.9 /usr/bin/python3
+          popd)
```
dmitrii-ubskii: You don't need both `pushd`/`popd` and a subshell. `cd`s in the subshell don't affect the parent shell. I'd remove the parentheses.

dmitrii-ubskii (approved):

PR #913, opened 2026-08-06
Make the HTTP driver correctly always follow cluster 'misdirected request' redirects

farost:
## Usage and product changes
Make the http-ts driver retry a couple of times when redirecting request (usually, it is needed when a new primary is selected, and there can be a lag between the states of the target cluster). 

Additionally, simplify the address resolution logic and avoid appending excessive schemas to the target URLs, expecting correct input params from the users.

### Example: 
If you specify three addresses, and the driver first tries a non-primary while connecting, this replica will say "I'm not a primary, go to XXX". If this XXX does not https://, the redirect fails. And it does not:
* understand that the returned address without the schema is exactly one of the addresses from the connection string
* try the other addresses from the connection string (because there are no retries)

After this change:
It **does**:
* understand that the returned address without the schema is exactly one of the addresses from the connection string (matching between the redirect address and the configured addresses is canonical)
* try the other addresses from the connection string (retries are introduced)

## Implementation
Replicate the retry logic from the Rust driver (which is propagated to all other grpc-based drivers).

### DIFF
--- a/http-ts/src/index.ts
+++ b/http-ts/src/index.ts
@@ -44,7 +44,13 @@ const HTTP_MISDIRECTED = 421;
 //   SRV14 - server returned "Not yet initialised". Happens on a freshly restarted secondary
 //           that has not yet loaded the system DB, or briefly on a node mid-promotion. Either
 //           way another origin is likely healthy, so fall back rather than retry in place.
-const TRANSIENT_SIGNIN_ERROR_CODES = new Set(["HDR2", "SRV14"]);
+//   CSV8/CSV9 - node is not the primary; another origin may be, so fall back.
+const TRANSIENT_SIGNIN_ERROR_CODES = new Set(["HDR2", "SRV14", "CSV8", "CSV9"]);
+
+// A misdirected (421) result that survived redirect-following and origin fallback means
+// no reachable server is primary right now (typically mid-election); wait and retry.
+const PRIMARY_FAILOVER_RETRIES = 3;
+const PRIMARY_SELECTION_WAIT_MS = 2000;
 
 // Result of attempting a single request against a single origin.
 //   Response          - reached the server and got an HTTP response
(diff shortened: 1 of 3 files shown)


http-ts/src/index.ts
```diff
@@ -44,7 +44,13 @@ const HTTP_MISDIRECTED = 421;
 //   SRV14 - server returned "Not yet initialised". Happens on a freshly restarted secondary
 //           that has not yet loaded the system DB, or briefly on a node mid-promotion. Either
 //           way another origin is likely healthy, so fall back rather than retry in place.
-const TRANSIENT_SIGNIN_ERROR_CODES = new Set(["HDR2", "SRV14"]);
+//   CSV8/CSV9 - node is not the primary; another origin may be, so fall back.
```
farost, line 47: More error handling (similar to Rust)

http-ts/src/index.ts
```diff
@@ -340,20 +355,20 @@ export class TypeDBHttpDriver {
...
+        return driverError("HDR2", `Unexpected signin response from ${this.currentOrigin} (HTTP ${resp.status})`);
     }
 
-    private async jsonOrNull(resp: Response) {
-        const contentLengthRaw = resp.headers.get("Content-Length");
-        if (!contentLengthRaw) return null;
-        const contentLength = parseInt(contentLengthRaw || "");
-        if (isNaN(contentLength)) throw `Received invalid Content-Length header: ${contentLengthRaw}`;
-        return contentLength > 0 ? await resp.json() : null;
+    private async jsonOrNull(resp: Response): Promise<any> {
```
farost, line 366: Simpler = better

http-ts/src/params.ts
```diff
@@ -59,31 +59,21 @@ export function hostPortFromOrigin(origin: string): string {
...
-                || hostPortFromOrigin(ta.external) === primaryAddress) {
+            if (hostPortFromOrigin(ta.internal) === target || hostPortFromOrigin(ta.external) === target) {
                 return ta.external;
             }
         }
-        return deriveOriginWithScheme(params.translatedAddresses[0].external, primaryAddress);
-    }
-}
-
-function deriveOriginWithScheme(referenceUrl: string, hostPort: string): string {
```
farost, line 80: it was too eager, we're good to go on our own.

alexjpwalker (approved):

PR #921, opened 2026-08-13
Make driver shutdown deterministic

farost:
## Usage and product changes

Closing or dropping a driver now returns only once all of its background threads have stopped, fixing rare crashes at process exit (random `AddressSanitizer:DEADLYSIGNAL` failures in the C integration CI).

## Implementation

`BackgroundRuntime::drop` joins the runtime worker thread; the worker stops the `tokio` runtime itself via `shutdown_timeout`, keeping the wait bounded and off any async context.

The callback handler thread is stopped afterwards via an explicit shutdown message, skipping the self-join when a close callback owns the last driver reference. 

Introduce a new `test_lifecycle` suite covering shutdown scenarios (absorbing the on-close tests). Additionally, make integration tests across Rust/Java/Python delete their databases on teardown so suites sharing a server stay clean.


java/test/integration/DriverTest.java
```diff
@@ -52,6 +52,7 @@ public static void setUpClass() {
 
     @AfterClass
     public static void close() {
+        Util.deleteDatabaseIfExists(typedbDriver, DB_NAME);
```
farost: I got sick of having leftovers after integration tests.

farost:
Doesn't solve the issue, giving up for now!

PR #924, opened 2026-09-09
Limit database import and export memory footprint

farost:
## Usage and product changes

Database import and export can no longer use gigabytes of RAM on large databases. A 25 GB / 5.6M-instance import that used to drive the client to 22.7 GB (and get OOM-killed) now holds flat at ~44 MB, with identical data and no speed change. 

Before, database import used unbounded channels, which, with slower server-side processing (which does much more work compared to direct file reads), placed almost the whole file into RAM, awaiting its data reads. 

The export did not suffer from this issue, because its unbounded channel depends on the inputs from the slower server. However, in case of a unique mismatch between the client/server hardware, its network channel became bounded, too, which did not affect the performance of exports in regular situations.

Important: the size of the RAM depends on the size of entries we don't currently check, but the existing caps for the buffers are safe enough to consider potential memory blowouts not at all critical.

## Implementation

Use bounded channels for both import and export:
- `SyncReceiver` / sync `Sender` for import's inner channels for file reading and network submissions
-  `BoundedNetworkStream` instead of `UnboundedNetworkStream` for database export to stop reading the server's batches in case we're slow to process them.


rust/src/common/stream_sync.rs
```diff
@@ -43,3 +43,20 @@ impl<T> Iterator for NetworkStream<T> {
         self.receiver.blocking_recv()
     }
 }
+
+pub struct BoundedNetworkStream<T> {
```
farost, line 47: The same as the Unbounded version, but in case if we're slower than the server, we don't blow out our memory for large files (database export)

rust/src/connection/network/transmitter/export.rs
```diff
@@ -61,27 +72,32 @@ impl DatabaseExportTransmitter {
...
             match grpc_source.next().await {
-                Some(Ok(message)) => response_sender
-                    .send(DatabaseExportResponse::try_from_proto(message))
-                    .expect("Expected export channel to be open. It is a bug."),
+                // send().await blocks when the client is behind (bounded queue full), so we stop
+                // pulling from the server and HTTP/2 flow control paces it. An error means the
+                // consumer is gone, so we stop.
+                Some(Ok(message)) => {
+                    if response_sender.send(DatabaseExportResponse::try_from_proto(message)).await.is_err() {
+                        break;
```
farost, line 90: An error means the consumer is gone, so we stop. The client's disk could get overfilled and drop the stream, which would lead to crush here -- incorrect. The actual error is returned from the place it originates. So it's not our job to reason whether it's correct or not.

tool/test/cluster-server.sh
```diff
@@ -29,7 +29,7 @@
 #   CLUSTER_DIR         - Directory containing numbered node dirs (default: CWD)
 #   ENCRYPTION_ENABLED  - Enable encryption (default: true)
 #   DEPLOYMENT_ID       - Deployment ID for diagnostics (default: test)
-#   INIT                - When 'true', passes --server.clustering.init=true (default: false)
+#   INIT_CREATE_CLUSTER - When 'true', the node creates a new cluster if absent (default: false)
```
farost, line 32: The flag was renamed in cluster

tool/test/start-cluster-servers.sh
```diff
@@ -86,6 +86,10 @@ if [ "$NODE_COUNT" -gt 1 ]; then
       fi
       if [ "$attempt" -eq "$REGISTER_MAX_RETRIES" ]; then
         echo "Failed to register replica ${i} after ${REGISTER_MAX_RETRIES} attempts"
+        for n in $(seq 1 $NODE_COUNT); do
```
farost, line 89: Took me some time to realise what's the issue, so added additional logs for failed attempts

rust/src/connection/network/transmitter/export.rs
flyingsilverfin: yeah looks good to me!

flyingsilverfin (approved):

PR #925, opened 2026-09-14
Fix document Kind decoding to match protocol wire values

flyingsilverfin:
## Usage and product changes

Fix mapping of `fetch` leaf kinds, which incorrectly mapped Attributes to Roles

## Implementation

The Kind enum in `answer.proto` maps `Entity=0, Relation=1, Attribute=3, Role=4`... It's not entirely clear why! However, to avoid breaking compatibility of existing drivers, we update the drivers to match.

The driver's TryFromProto<i32> for Kind assumed a contiguous 0-3 range, so it decoded Attribute (wire 3) as Role, and Role (wire 4) failed with UnexpectedKind. This also affected constraint Kind decoding in query analysis, which reuses the same decoder.

Note: the server side correctly emits the right enum values.


rust/src/connection/network/proto/concept.rs
```diff
@@ -332,11 +332,12 @@ impl TryFromProto<ValueProto> for Value {
...
+        // Wire values follow ConceptDocument.Node.Leaf.Kind in answer.proto, which reserves 2.
         match kind {
             0 => Ok(Self::Entity),
             1 => Ok(Self::Relation),
-            2 => Ok(Self::Attribute),
-            3 => Ok(Self::Role),
+            3 => Ok(Self::Attribute),
+            4 => Ok(Self::Role),
             _ => Err(ConnectionError::UnexpectedKind { kind }.into()),
         }
```
krishnangovindraj: I've seen recommendations for something like this:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=527f4bc7b4897e8e84dadced957bed23

It's long, but it's harder to get wrong. We could do the same for QueryType and any other enums.

flyingsilverfin: Good idea, lemme do that

flyingsilverfin: Turns out here that claude found an even better solution!

flyingsilverfin: doesnt work in the QueryType though

rust/src/connection/network/proto/concept.rs
```diff
@@ -419,3 +421,22 @@ impl IntoProto<value_proto::DatetimeTz> for DateTime<TimeZone> {
...
+        assert_eq!(Kind::try_from_proto(3).unwrap(), Kind::Attribute);
+        assert_eq!(Kind::try_from_proto(4).unwrap(), Kind::Role);
+    }
+
+    #[test]
+    fn kind_rejects_reserved_and_unknown_values() {
+        assert!(Kind::try_from_proto(2).is_err());
+        assert!(Kind::try_from_proto(5).is_err());
+    }
+}
```
krishnangovindraj: These are a bit strange to have here. It made sense in the previous implementaiton

krishnangovindraj (approved):

rust/src/connection/network/proto/concept.rs
krishnangovindraj: This seems to work fine for me for QueryType
```
impl TryFromProto<i32> for QueryType {
    fn try_from_proto(query_type: i32) -> Result<Self> {
        use typedb_protocol::query::Type as QueryTypeProto;
        let proto_type = QueryTypeProto::try_from(query_type);
        match proto_type {
            Ok(QueryTypeProto::Read) => Ok(Self::ReadQuery),
            Ok(QueryTypeProto::Write) => Ok(Self::WriteQuery),
            Ok(QueryTypeProto::Schema) => Ok(Self::SchemaQuery),
            Err(_) => Err(ConnectionError::UnexpectedQueryType { query_type }.into()),
        }
    }
}
```

PR #926, opened 2026-09-15
Add delete concepts analyze

flyingsilverfin:
## Change & Motivation

This PR adds the new `DeleteConcepts` support across every driver: Rust, Python, Java, C#, and the standalone HTTP TypeScript driver. This matches the updated protocol which gained a `DeleteConcepts` constraint in the query-analysis structure ([typedb-protocol #258](https://github.com/typedb/typedb-protocol/commit/310fef4e4ed2e735ed0fb2dc3248b4a35e604806)), though TypeDB server already emits it correctly.

## Implementation

- Bumps the typedb_protocol pin past the commit that adds the delete_concepts wire field. 
- Adds `Constraint::DeleteConcepts { variables }` (or the equivalent per language) end to end: Rust core decode, the shared C FFI layer, and each language's independently hand-maintained wrapper hierarchy — Python, Java, and C# each wrap the same C ABI separately, so the addition isn't automatic in any of them. http-ts talks to the server's HTTP API directly and needed the same addition in its own type definitions. 
- Add  C SWIG %newobject for the new getter
- Regenerate API docs for the new public types.
- Fixes pre-existing unrelated bug uncovered while verifying this: http-ts's analyze BDD steps imported two type names (QueryConstraintAny, QueryVertex) that don't exist anywhere in http-ts/src and couldn't have compiled


csharp/Test/Behaviour/Util/FunctorEncoder.cs
```diff
@@ -248,6 +248,11 @@ public string Encode(IConstraint constraint)
                     var iid = constraint.AsIid();
                     return MakeFunctor("Iid", Encode(iid.Variable), iid.Iid);
                 }
+                else if (constraint.IsDeleteConcepts)
+                {
+                    var deleteConcepts = constraint.AsDeleteConcepts();
+                    return MakeFunctor("DeleteConcepts", EncodeList(deleteConcepts.Variables.Select(Encode)));
+                }
```
flyingsilverfin, line 255: Idk if this is right?

http-ts/tests/behaviour/steps/analyze.ts
```diff
@@ -26,8 +26,8 @@ import {
     AnalyzedFetch,
     AnalyzedFunction,
     AnalyzedPipeline,
-    QueryConstraintAny,
-    QueryVertex,
+    ConstraintAny,
+    ConstraintVertexAny,
```
flyingsilverfin, line 30: Pls verify

c/swig/typedb_driver_csharp.swg
krishnangovindraj, line 1: I'd expect some changes in `typedb_driver.i` as well. Maybe just:
`%newobject constraint_delete_concepts_get_variables;`

csharp/Test/Behaviour/Util/FunctorEncoder.cs
krishnangovindraj: Select seems to be the equivalent of map. 
If it passes this is fine since this is test code.

http-ts/tests/behaviour/steps/analyze.ts
krishnangovindraj: Yes, those look like renames that didn't get propagated to the tests.

krishnangovindraj (approved):
Just one %newobject I'd expect in `typedb_driver.i`

